Issue #2, opened 2021-12-25
liblinker_fix_3ds.a and libpthread_3ds.a


AzureMarker:
Where did these files come from? Can they be generated/compiled instead of committed?

Meziu:
https://github.com/Meziu/pthread-3ds

https://github.com/Meziu/rust-linker-fix-3ds

AzureMarker:
I'm trying to find some way to include them in the build explicitly instead of passing around the compiled outputs. The issue is that lots of things need to use their symbols (ex. std for pthread) and Cargo doesn't link crates to each other unless they specify a dependency. Hence the static lib approach. But maybe there's another way?

Meziu:
Originally I wanted to include the definitions into ctru-rs, but quickly noticed the compiler didn't accept C definitions from inside the crate, so I had to link them as C libraries instead. I don't think there is much to do (other than make a build script or something, though I don't know how that would work for a crate imported from cargo, which is the ultimate goal of this project).

AzureMarker:
I was able to get it narrowed down to the following missing symbols after adding the crates as dependencies and implementing the workaround for https://github.com/rust-lang/rust/issues/47384 (adding an empty function to the crates and calling it in main):
```
/opt/devkitpro/devkitARM/bin/../lib/gcc/arm-none-eabi/11.1.0/../../../../arm-none-eabi/bin/ld: /opt/devkitpro/devkitARM/arm-none-eabi/lib/armv6k/fpu/libc.a(lib_a-sysfstat.o): in function `fstat':
(.text.fstat+0x24): undefined reference to `_fstat_r'
/opt/devkitpro/devkitARM/bin/../lib/gcc/arm-none-eabi/11.1.0/../../../../arm-none-eabi/bin/ld: /opt/devkitpro/devkitARM/arm-none-eabi/lib/armv6k/fpu/libc.a(lib_a-syslseek.o): in function `lseek':
(.text.lseek+0x30): undefined reference to `_lseek_r'
/opt/devkitpro/devkitARM/bin/../lib/gcc/arm-none-eabi/11.1.0/../../../../arm-none-eabi/bin/ld: /opt/devkitpro/devkitARM/arm-none-eabi/lib/armv6k/fpu/libc.a(lib_a-sysopen.o): in function `open':
(.text.open+0x2c): undefined reference to `_open_r'
[...]

PR #5, opened 2021-12-28
Use linker-fix-3ds via dependency

AzureMarker:
This avoids pulling in a binary blob (liblinker_fix_3ds.a). Related to #2.

Need to merge https://github.com/Meziu/rust-linker-fix-3ds/pull/3 first though, so it can be used as a dependency.

I'm trying to do this same change with the pthread library, but I'm running into some missing symbols when I switch to the dependency instead of binary static lib.

Also includes some automatic rustfmt changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,11 +10,8 @@ edition = "2018"
 crate-type = ["lib", "staticlib"]
 name = "ctru"
 
-[dependencies.libc]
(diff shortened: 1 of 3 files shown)


Meziu:
Wow, didn't expect this was possible just by changing the crate type :laughing:. Guess I have still much to learn...

PR #6, opened 2022-01-01
Implement Drop for Console so printing doesn't crash after dropping

AzureMarker:
This will fix the exception during panic unwinding where it tries to print after the console has been deallocated.


Meziu (approved):
Ok, this will have the console act as a backtrace printer after the `main` scope. A `thread::sleep` or a similar fix should be implemented in the `std` exit code to show the user the panic info.

PR #12, opened 2022-01-08
Add wide screen mode support to Gfx

AzureMarker:
I was just poking around libctru and was curious what this was. Now we have ctru-rs support and an example!


Meziu:
While not a necessity for the moment, support for more features is always welcome. 😄

PR #14, opened 2022-01-10
Testing and fixing the file system support

### LINKED ISSUE #11: Test file system functionality and fix if necessary
The SD alone is fine, but maybe heading off into `romfs` territory would be good. I haven't had the time to do this myself, yet. We have to test filesystems before reworking `cargo-3ds`, because that would require integration for building with `romfs` inclusion.

_Originally posted by @Meziu in https://github.com/Meziu/ctru-rs/issues/9#issuecomment-1006875085_

AzureMarker:
Closes #11 

~~Before this PR gets out of draft I also want to test the ctru-rs file system support.~~ (Edit: I can do this separately.) Maybe we won't need the SD card code if std already does it?

### DIFF
--- a/ctru-rs/src/lib.rs
+++ b/ctru-rs/src/lib.rs
@@ -37,20 +37,23 @@ pub mod services;
 pub mod srv;
 pub mod thread;
 
-#[cfg(all(feature = "romfs", romfs_exists))]
-pub mod romfs;
-#[cfg(not(all(feature = "romfs", romfs_exists)))]
-pub mod romfs {
-    //! The RomFS folder has not been detected and/or the `romfs` feature has not been enabled.
-    //!
-    //! Configure the path in Cargo.toml (the default path is "romfs"). Paths are relative to the
-    //! `CARGO_MANIFEST_DIR` environment variable, which is the directory containing the manifest of
-    //! your package.
-    //!
-    //! ```toml
-    //! [package.metadata.cargo-3ds]
-    //! romfs_dir = "romfs"
-    //! ```
+cfg_if::cfg_if! {
+    if #[cfg(all(feature = "romfs", romfs_exists))] {
+        pub mod romfs;
+    } else {
+        pub mod romfs {
+            //! The RomFS folder has not been detected and/or the `romfs` feature has not been enabled.
+            //!
+            //! Configure the path in Cargo.toml (the default path is "romfs"). Paths are relative to the
+            //! `CARGO_MANIFEST_DIR` environment variable, which is the directory containing the manifest of
+            //! your package.
+            //!
+            //! ```toml
+            //! [package.metadata.cargo-3ds]
+            //! romfs_dir = "romfs"
+            //! ```
+        }
+    }
 }
 
 pub use crate::error::{Error, Result};
--- a/ctru-rs/Cargo.toml
+++ b/ctru-rs/Cargo.toml
@@ -11,6 +11,7 @@ crate-type = ["rlib"]
 name = "ctru"
 
 [dependencies]
+cfg-if = "1.0"
 ctru-sys = { path = "../ctru-sys", version = "0.4" }
 const-zero = "0.1.0"
 linker-fix-3ds = { git = "https://github.com/Meziu/rust-linker-fix-3ds.git" }


AzureMarker:
Note: there's some issue in release mode causing a segfault. Currently investigating:
image

Meziu:
I rebased, you should pull the latest ctru-rs to and rebase this commit upon it.

Meziu:
I reset it to the state before rebasing. To be honest this only reinforces to me the idea that the Drop implementation isn't the way to go. We should either go the "print method" route, or find another way to fix this, like initliazing the defaultConsole in `ctru::init`

AzureMarker:
Dumped the LLVM IR and MIR to try and debug. Maybe it's a miscompilation?

Rust: https://github.com/AzureMarker/rust-horizon/blob/58a74d1ffae2f55c6bb9fdae74d940244cc4e012/library/std/src/sys/unix/fs.rs#L589-L600
LLVM IR: https://gist.github.com/AzureMarker/65edc6d64e52a09f7eaffd3728a86b9c
MIR: https://gist.github.com/AzureMarker/78f9267fab2ce1f32177af64cfa194f0

Meziu:
Before thinking at miscompilations (not because improbable, but for optimism), I'd suggest too look at the validity of the costants involved, and even if there is any call to `realpath`, because it is probably wrong.

AzureMarker:
You know, that is actually a really good call out. I hope that's the issue. There's a call to `lstat` here which might under the hood call realpath. I pulled up the docs for realpath and noticed that the output parameter could be null (requiring the implementation to allocate the buffer):
https://man7.org/linux/man-pages/man3/realpath.3.html

The linker fix implementation blindly memcpy's to that output path, which is very similar to what's happening in this case. There's some data being written to a null pointer. I'll try implementing a better realpath to see if that fixes it.

AzureMarker:
It still seems to happen, even with this realpath impl:
```rust
#[no_mangle]
unsafe extern "C" fn realpath(
    path: *const libc::c_char,
    mut resolved_path: *mut libc::c_char,
) -> *mut libc::c_char {
    let path_len = libc::strlen(path);

    if resolved_path.is_null() {
        resolved_path = libc::malloc(path_len + 1) as _;
    }

    libc::memcpy(resolved_path as _, path as _, libc::strlen(path) + 1);

    resolved_path
}

```

I also didn't see it trigger the breakpoint I set for realpath.

AzureMarker:
I added the `|m| m.file_type()` closure to the MIR link. It looks tame enough:
```
fn sys::unix::fs::<impl at /home/mark/media/CLionProjects/rust-horizon/build/x86_64-unknown-linux-gnu/stage1/lib/rustlib/src/rust/library/std/src/sys/unix/fs.rs:538:1: 675:2>::file_type::{closure#0}(_1: [closure@/home/mark/media/CLionProjects/rust-horizon/build/x86_64-unknown-linux-gnu/stage1/lib/rustlib/src/rust/library/std/src/sys/unix/fs.rs:598:42: 598:59], _2: FileAttr) -> sys::unix::fs::FileType {
    debug m => _2;                       // in scope 0 at /home/mark/media/CLionProjects/rust-horizon/build/x86_64-unknown-linux-gnu/stage1/lib/rustlib/src/rust/library/std/src/sys/unix/fs.rs:598:43: 598:44
    let mut _0: sys::unix::fs::FileType; // return place in scope 0 at /home/mark/media/CLionProjects/rust-horizon/build/x86_64-unknown-linux-gnu/stage1/lib/rustlib/src/rust/library/std/src/sys/unix/fs.rs:598:46: 598:46
    let mut _3: &sys::unix::fs::FileAttr; // in scope 0 at /home/mark/media/CLionProjects/rust-horizon/build/x86_64-unknown-linux-gnu/stage1/lib/rustlib/src/rust/library/std/src/sys/unix/fs.rs:598:46: 598:59

    bb0: {
        StorageLive(_3);                 // scope 0 at /home/mark/media/CLionProjects/rust-horizon/build/x86_64-unknown-linux-gnu/stage1/lib/rustlib/src/rust/library/std/src/sys/unix/fs.rs:598:46: 598:59
        _3 = &_2;                        // scope 0 at /home/mark/media/CLionProjects/rust-horizon/build/x86_64-unknown-linux-gnu/stage1/lib/rustlib/src/rust/library/std/src/sys/unix/fs.rs:598:46: 598:59
        _0 = FileAttr::file_type(move _3) -> bb1; // scope 0 at /home/mark/media/CLionProjects/rust-horizon/build/x86_64-unknown-linux-gnu/stage1/lib/rustlib/src/rust/library/std/src/sys/unix/fs.rs:598:46: 598:59
                                         // mir::Constant
                                         // + span: /home/mark/media/CLionProjects/rust-horizon/build/x86_64-unknown-linux-gnu/stage1/lib/rustlib/src/rust/library/std/src/sys/unix/fs.rs:598:48: 598:57
                                         // + literal: Const { ty: for<'r> fn(&'r sys::unix::fs::FileAttr) -> sys::unix::fs::FileType {sys::unix::fs::FileAttr::file_type}, val: Value(Scalar(<ZST>)) }
    }

    bb1: {
        StorageDead(_3);                 // scope 0 at /home/mark/media/CLionProjects/rust-horizon/build/x86_64-unknown-linux-gnu/stage1/lib/rustlib/src/rust/library/std/src/sys/unix/fs.rs:598:58: 598:59
        return;                          // scope 0 at /home/mark/media/CLionProjects/rust-horizon/build/x86_64-unknown-linux-gnu/stage1/lib/rustlib/src/rust/library/std/src/sys/unix/fs.rs:598:59: 598:59
    }
}
```

Looking through the LLVM IR is much more difficult.

AzureMarker:
I'm also curious if it's due to an issue in our libc definition of these types.

AzureMarker:
If you want to take a look, here's my ELF, 3DSX, and a core dump I got from gdb via `generate-core-file`:
[file_explorer-core-dump.zip](https://github.com/Meziu/ctru-rs/files/7852154/file_explorer-core-dump.zip)

AzureMarker:
Added ARM disassembly:
https://gist.github.com/AzureMarker/fcfe5381ac5541f15d67eeb699a6b595

Meziu:
Wow, it auto closed when merging the one in `rust-linker-fix`. Please reopen it

AzureMarker:
I don't have permission to re-open this. You probably do though.

Meziu:
I pressed the button 15 times and thought only you could. Guess the github iOS app is just trash, but I knew that already since it doesn't let you view comments on commits.

Meziu:
Since everything else is closed, and this has some `libc`/`std` implications, this is now the priority :confetti_ball:. Heading back into `gdb` is an experience I would've gladly never have to do again.

AzureMarker:
I'm going to try to switch it to the libctru types locally and see if that fixes the crash in release mode.

AzureMarker:
Weird... After the merge I don't get the crash in release mode.

Meziu:
Hmm, maybe you had some dependency not updated? Idk, I'll try it tomorrow morning and see how it goes.

AzureMarker:
Maybe the libc update fixed it? I tried to checkout the commit before the merge in this PR, but it still works.

AzureMarker:
Good news, romfs works via the normal std interface (if you mount it to "romfs", you can just open the path `romfs:/`).

AzureMarker:
For some reason Luma3ds was giving me garbage screenshots (seems related to the Homebrew app), and Citra doesn't support wide screen mode, plus Citra's screenshot tool wasn't saving anything. Had to do a manual desktop screenshot.
image

Meziu:
I tried running your example and I stumbled upon the release crash. I thought I may have not updated my workspace, so I did a `cargo update` to be sure, but nothing changed. I thought you had done something wrong, but I decided to test more, so I rebuilt my whole `stage2` compiler. It works now. I bet the changes in our `libc` influenced the `std` in some way, and now it works. Cool. Since this works already, tell me when you want to merge it.

Meziu:
> For some reason Luma3ds was giving me garbage screenshots (seems related to the Homebrew app)

I tried around for a bit and it seems like a problem with `Console` applications. It may be related to how the `Console` renders to the screen.

Meziu (commented):
> Seems to crash the 3DS if there is no actual RomFS data in the 3dsx
file, so there's room for improvement.

We can get around it by having a `#[cfg(romfs)]` check on the `ctru::romfs` module. We can activate the flag via `cargo-3ds` if we find the `romfs` directory. This stops compilation of programs using the module without a directory, but also means the error message is slightly misleading (since it just tells the user the module doesn't exist). Using features is a no go since cargo-3ds won't be able to activate it from the upper level of an user project.

AzureMarker:
> We can activate the flag via `cargo-3ds` if we find the `romfs` directory.

I think we could even do this in a ctru-rs build script, which would be better.

Meziu:
> > We can activate the flag via `cargo-3ds` if we find the `romfs` directory.
> 
> 
> 
> I think we could even do this in a ctru-rs build script, which would be better.

True, you could implement it if you have time for it.

Meziu:
@AzureMarker though, I was thinking, maybe we should add it to the Cargo manifest instead (like `icon` and `description`). I think that would make the user more aware of the use. It surely feels better than just getting it from the env.

AzureMarker:
How about a feature flag plus the build script method? That way they have to manually enable it, but they still can't break things if there isn't a romfs path? We could also make the romfs path configurable in Cargo.toml, though that would require support in cargo-3ds.

Meziu:
Cool. That looks like the best idea.

AzureMarker:
It would also be great if we could detect if romfs works at runtime, so we wouldn't have to do any of these conditional compilation things. I'll see if that's possible...

AzureMarker:
I looked at if we could detect RomFS support at runtime (if the 3dsx is configured with it), and it's possible. The hard part is we'd have to do what libctru does, which is open up the 3dsx file at runtime, read the header, and determine if the RomFS offset is valid (ex. in range of the file). libctru doesn't do this check unfortunately, which causes the segfault.

I implemented the feature flag + cfg selector we discussed, along with making the path configurable. This required a change in cargo-3ds to support it:
https://github.com/Meziu/cargo-3ds/pull/10

AzureMarker:
I think this PR is now merge-able / ready for review. We can make more PRs to flesh out the `romfs` module as needed.

ctru-rs/src/lib.rs
```diff
@@ -38,6 +38,22 @@ pub mod services;
 pub mod srv;
 pub mod thread;
 
+#[cfg(all(feature = "romfs", romfs_exists))]
+pub mod romfs;
+#[cfg(not(all(feature = "romfs", romfs_exists)))]
+pub mod romfs {
```
Meziu, line 44: We should use `cfg_if`, it's nicer to read.

Meziu (changes requested):
You should remove the `sdmc` module. It's useless to keep (I am pretty sure that it was needed in 2015 but `libctru` has added on-startup support for `sdmc` ever since

AzureMarker:
I looked at removing the FS service (`services/fs.rs`). It came (somewhat) from this PR:
https://github.com/rust3ds/ctru-rs/pull/10

It looks like it might have some valid uses (there's a lot of archive IDs in `ArchiveID`), so I'll leave it for now. It's kind of a mess though since it copies and tweaks a bunch of std types like `File`...

Meziu:
> It looks like it might have some valid uses (there's a lot of archive IDs in `ArchiveID`), so I'll leave it for now. It's kind of a mess though since it copies and tweaks a bunch of std types like `File`, `OpenOptions`, etc...

I looked at it too. It's very interesting (stuff like ExtData, SaveData and even the NaND access). I never used this functionality, even when using `libctru` in C, yet it looks like a very useful addition. I wonder if we could implement it using `std::file` and not re-implementations.

PR #17, opened 2022-01-11
Add back Console::drop

### LINKED ISSUE #15: Console printing and lifetimes
>I reset it to the state before rebasing. To be honest this only reinforces to me the idea that the Drop implementation isn't the way to go. We should either go the "print method" route, or find another way to fix this, like initliazing the defaultConsole in ctru::init (edit: we can't do this second option either because it must be the user to initialize Gfx, making the whole Console implemenetations unusable until then. With this knowledge, I believe we should also require to pass the Gfx object to Console to be sure the Console init won't crash)

Written by me in #14.

@AzureMarker let's talk here about this.

Priority is for no crashes to be caused from `safe` code the user may write. This would be best to solve with support for `println!` and other "printing" functions, since we can't limit access to `std` code.

AzureMarker:
This is needed for soundness. Otherwise libctru will have a dangling pointer and will write to it when printing.

Closes #15 (I think, maybe we also need some panic changes before closing?)

### DIFF
--- a/ctru-rs/Cargo.toml
+++ b/ctru-rs/Cargo.toml
@@ -12,6 +12,7 @@ name = "ctru"
 
 [dependencies]
 ctru-sys = { path = "../ctru-sys", version = "0.4" }
+const-zero = "0.1.0"
 linker-fix-3ds = { git = "https://github.com/Meziu/rust-linker-fix-3ds.git" }
 pthread-3ds = { git = "https://github.com/Meziu/pthread-3ds.git" }
 libc = { git = "https://github.com/Meziu/libc.git" }
--- a/ctru-rs/src/console.rs
+++ b/ctru-rs/src/console.rs
@@ -53,22 +53,23 @@ impl Default for Console {
 
 impl Drop for Console {
     fn drop(&mut self) {
+        static mut EMPTY_CONSOLE: PrintConsole = unsafe { const_zero::const_zero!(PrintConsole) };
+
         unsafe {
             // Safety: We are about to deallocate the PrintConsole data pointed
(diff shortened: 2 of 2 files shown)


Meziu:
Testing your branch against a simple "make console, drop it and then print" binary, made it crash on `drawChar`:
image

Does it work to you?

Meziu:
Just to be on the same page, my code is:

```
ctru::init();
    let gfx = Gfx::default();
    let hid = Hid::init().expect("Couldn't obtain HID controller");
    let apt = Apt::init().expect("Couldn't obtain APT controller");
    let _console = Console::default();
    drop(_console);

    println!("Works");
```

AzureMarker:
I guess I had the misconception that the initial selected console value is the default console. It's actually a zeroed-out console:
https://github.com/devkitPro/libctru/blob/5f13628dac75206f0c97d29a7427ce8284d910f1/libctru/source/console.c#L65-L67

(I was wondering if it would use uninitialized memory or zeroed memory, and found this: https://stackoverflow.com/questions/4080673/are-the-members-of-a-global-structure-initialized-to-zero-by-default-in-c)

So I'll change it to set the console to some zeroed out console on drop, if it was the selected console.

AzureMarker:
Unfortunately right now `std::mem::zeroed` is not `const`:
https://github.com/rust-lang/rust/issues/62061

So I used the recommended workaround:
https://crates.io/crates/const-zero

AzureMarker:
It's no longer simple, but at least it's safe.

Meziu:
Nice, not printing once the Console is dropped is both expected behaviour and easy to use.

PR #19, opened 2022-01-14
Fix dropping Gfx before Console

AzureMarker:
This prevents dropping the Gfx object before the console. We want to do this because printing after dropping gfx doesn't do anything (gfx isn't rendering the text anymore).

Note that even though Rust doesn't understand the relationship between Console and println, even if you don't use the console variable Rust will prevent Gfx dropping because Console's Drop code _might_ use the Gfx object (it doesn't, but we get the error we want anyway).

Also made importing in the examples more consistent.

### DIFF
--- a/ctru-rs/examples/buttons.rs
+++ b/ctru-rs/examples/buttons.rs
@@ -1,5 +1,5 @@
 use ctru::console::Console;
-use ctru::gfx::Gfx;
+use ctru::gfx::{Gfx, Screen};
 use ctru::services::apt::Apt;
 use ctru::services::hid::{Hid, KeyPad};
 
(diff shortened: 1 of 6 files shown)


Meziu:
So, the only difference between this and holding a pointer to `Gfx` is "not using the memory to hold the pointer", but the compiler effectively behaves like `Console` held an immutable reference (I noticed you also can't have `&mut` references towards `Gfx` because of the mutability safety)

Meziu:
I believe there is an issue with this PR, I reverted it to be sure

Meziu:
Still, I don't see why we would want to do this. If the user decides to drop `Gfx`, not havi graphics is exactly the outcome they would expect. This also doesn't cause any memory issues or errors related to `libctru`.I don't understand why we should lock a safe and intended behaviour from occurring.

AzureMarker:
Oh, I thought you were expecting this behavior and was surprised when it didn't work like this. We could revert it in that case.

However, why would anyone want to do such a thing? For one, the console feature requires graphics to function, so why drop graphics and not console as well. Second, there's better ways to clear the screen (ex. the console's `clear` function, or just dropping the console).

Meziu:
No no, I believe there is an use to keeping a reference, other than for the lifetime.

PR #22, opened 2022-01-15
Added mutability checks for Gfx

Meziu:
Related to #19. @AzureMarker 

The `consoleInit` process actually depends (and assumes) changes in the `gfx` environment, mainly related to framebuffers. By using Rust's mutable reference checks, we can prevent the environment from changing while a `Console` object is alive. This change has however a problem, in that (as it is written right now) the 2 screens (which would normally be worked with separately in this contexts) are actually both stuck together, since the `ctru-rs` `Gfx` implementation puts their control under one struct.

I believe we should separate the `Screen`s, thus to not block the `Gfx` mutability as a whole, but just singular screen controllers. Maybe internal mutability and moving some of the `Gfx` methods to the singular screens (like `set_framebuffer_format`) is the way.


ctru-rs/src/gfx.rs
```diff
@@ -114,14 +114,12 @@ impl Gfx {
...
     /// if double buffering is enabled
-    pub fn get_raw_framebuffer(&self, screen: Screen, side: Side) -> (*mut u8, u16, u16) {
-        unsafe {
-            let mut width: u16 = 0;
-            let mut height: u16 = 0;
-            let buf: *mut u8 =
-                ctru_sys::gfxGetFramebuffer(screen.into(), side.into(), &mut width, &mut height);
-            (buf, width, height)
-        }
+    pub unsafe fn get_raw_framebuffer(&self, screen: Screen, side: Side) -> (*mut u8, u16, u16) {
```
AzureMarker, line 117: Clippy warns that this should have a `# Safety` section to note why it's unsafe.

ctru-rs/src/gfx.rs
```diff
@@ -48,14 +48,14 @@ impl Gfx {
...
-    pub fn set_3d_enabled(&self, enabled: bool) {
+    pub fn set_3d_enabled(&mut self, enabled: bool) {
         unsafe { ctru_sys::gfxSet3D(enabled) }
     }
 
     /// Enable or disable the wide screen mode (top screen).
     ///
     /// This only works when 3D is disabled.
-    pub fn set_wide_mode(&self, enabled: bool) {
+    pub fn set_wide_mode(&mut self, enabled: bool) {
```
AzureMarker, line 58: This change means the wide mode example needs to be updated.

AzureMarker:
> This change has however a problem, in that (as it is written right now) the 2 screens (which would normally be worked with separately in this contexts) are actually both stuck together, since the `ctru-rs` `Gfx` implementation puts their control under one struct.
> 
> I believe we should separate the `Screen`s, thus to not block the `Gfx` mutability as a whole, but just singular screen controllers. Maybe internal mutability and moving some of the `Gfx` methods to the singular screens (like `set_framebuffer_format`) is the way.

I'm not sure I follow this. I think you're pointing out that the `Gfx` struct (and libctru's gfx API by association) affect both screens? I guess I haven't looked too much at libctru's gfx API to see if you can set settings per-screen (based on the selected screen?), but if so then the second paragraph makes a little more sense.

There also seem to be some functions that can run while the console is around, and they work kind of as expected (ex. wide mode toggle, though the width doesn't get updated in console so it wraps at the old width). Is this something we need to support, or is it not how the libctru API is expected to be used?

ctru-rs/src/gfx.rs
AzureMarker: The wide mode example might need to recreate the console when switching modes, if my understanding of the goal state is correct.

If you try the simple fix of making `gfx` mutable in the wide mode example to fix the compile error, it works. But we also don't want it to work since we're calling a mutable method on gfx after passing an immutable reference to console. The reason this works is due to [Non-Lexical Lifetimes](https://rust-lang.github.io/rfcs/2094-nll.html) (console's use of the gfx lifetime ends right after its creation). You can see this in action if you use the console at the end of the program, such as by adding `drop(_console)` (making the mutable gfx call in between two console usages). We should decide if this is something we want to prevent, since it is probably common for the user to never interact with the console after creation.

AzureMarker: Maybe we could have the console consume the Gfx object (pass ownership), and then expose a `fn gfx(&self) -> &Gfx` method on `Console`.

This might make some other use cases harder though, for example initializing two screens. I think I need to better understand how libctru intends its gfx API to work in some of these scenarios. Then we could find a way to enforce that usage pattern in Rust.

Meziu:
Ok, I wasn't clear. By my "assumes changes", I mean a Console makes some changes to the `gfx` environment when it is initialized. 

This is mainly about: the screen's framebuffer format (the framebuffers are screen specific, so putting those methods under them isn't an issue), the screen's double buffering mode (it must be off for a Console to work properly), and the wide mode check (as you said, it doesn't work properly because the Console cannot detect a change after init).

These are all screen specific features, that must not change in state while the Console exists. This is why I want to use the mutability checks in the first place. Though, as I stated, right now they are effected globally by a single Gfx instance, and that's why I believe we should separate them.

ctru-rs/src/gfx.rs
Meziu: I should run clippy more. I will add it when changing the other stuff, though it's pretty obvious that a getter for raw pointers to a framebuffer (which as stated in libctru, it may not even be the framebuffer currently used if you are in double buffering mode), should be unsafe.

ctru-rs/src/gfx.rs
Meziu: No, the consuming isn't an option. The user wants to keep ownership at all times of the Gfx service. Isn't there a macro, or a #[] statement to prevent this behaviour?

ctru-rs/src/gfx.rs
AzureMarker: Well actually it's safe to do that. What's unsafe is dereferencing the pointer. This program is perfectly safe:
```rust
fn foo(foo: *mut usize) {
    println!("Hello, I'm at {:?}", foo);
}

fn main() {
    foo(0 as *mut usize);
}
```
[Playground link](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=d5c688c8afb983fe1e9b8dbe794e9a9b)

Meziu: An "unsafe" function doesn't have to run unsafe code. I marked it unsafe because it returns a pointer to raw graphic memory, which isn't something safe to do under any circumstance.

AzureMarker:
> This is mainly about: the screen's framebuffer format (the framebuffers are screen specific, so putting those methods under them isn't an issue), the screen's double buffering mode (it must be off for a Console to work properly)
>
> These are all screen specific features

I was typing up something here to argue that these are not `Screen` specific features but `Gfx` features, when I realized I was confusing `Screen` and `Console` types. We only have a `Screen` enum at the moment, but if we made that into a full struct like `Console` then having those gfx methods on it would make more sense. My concern was tying these screen APIs behind owning a `Console` (which in hindsight isn't what you were talking about, sorry).

Maybe we could do something like this: Add `top_screen` and `bottom_screen` methods to `Gfx` which return `Screen` references (we could have immutable and mutable versions, or use something like `Arc` to avoid lifetimes). Then prevent the user from creating their own `Screen` objects. This way we will only ever have two `Screen` objects (at least per `Gfx` object, which we should have only one of anyways). I looked at how libctru's API models this, and I see they use an enum for `gfxScreen_t` just like our current `Screen`, and that gets passed in as a param to gfx APIs. So we'd just have to make sure that doesn't clash with our API model.

ctru-rs/src/gfx.rs
AzureMarker: Macros just generate code, so they don't fundamentally affect how Rust works (i.e. they can't turn off NLL, but they might be able to emulate turning it off by inserting a `drop(console)` call). I don't think we want to introduce any new macros or proc-macro attributes into user code at this point though.

I tried out the gfx mutable fix in the wide mode example again (make `gfx` var mutable) and it fails with this error, which is what we want to happen:
```
error[E0502]: cannot borrow `gfx` as mutable because it is also borrowed as immutable
  --> ctru-rs/examples/gfx_wide_mode.rs:26:13
   |
14 |     let _console = Console::init(&gfx, Screen::Top);
   |                                  ---- immutable borrow occurs here
...
26 |             gfx.set_wide_mode(!gfx.get_wide_mode());
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ mutable borrow occurs here
...
33 | }
   | - immutable borrow might be used here, when `_console` is dropped and runs the `Drop` code for type `ctru::console::Console`
```

I guess I forgot to update my comment before, because I definitely tried this out. So it is working as we expect, but only because we have the `Drop` impl on `Console`.

To fix the wide screen example I tried something like this, but it doesn't compile...
```diff
diff --git a/ctru-rs/examples/gfx_wide_mode.rs b/ctru-rs/examples/gfx_wide_mode.rs
index f4ba365..64670c9 100644
--- a/ctru-rs/examples/gfx_wide_mode.rs
+++ b/ctru-rs/examples/gfx_wide_mode.rs
@@ -10,10 +10,10 @@ fn main() {
     ctru::init();
     let apt = Apt::init().unwrap();
     let hid = Hid::init().unwrap();
-    let gfx = Gfx::default();
-    let _console = Console::init(&gfx, Screen::Top);
+    let mut gfx = Gfx::default();
+    let mut console = Console::init(&gfx, Screen::Top);
 
-    println!("Press A to enable/disable wide screen mode.");
+    println!("Press A to enable/disable wide screen mode. Extra words here so the text wraps.");
 
     while apt.main_loop() {
         hid.scan_input();
@@ -23,7 +23,14 @@ fn main() {
         }
 
         if hid.keys_down().contains(KeyPad::KEY_A) {
+            drop(console);
+
             gfx.set_wide_mode(!gfx.get_wide_mode());
+
+            console = Console::init(&gfx, Screen::Top);
+            println!(
+                "Press A to enable/disable wide screen mode. Extra words here so the text wraps."
+            );
         }
 
         gfx.flush_buffers();
```

Meziu:
That was what I thought of too. Giving methods to the Screen struct will be enough I think.

ctru-rs/src/gfx.rs
AzureMarker: I'm pretty sure this is a bug in rustc, so I opened an issue while we try to work around it:
https://github.com/rust-lang/rust/issues/92984

AzureMarker:
We'd also want to ensure `Screen` only runs gfx functions after a `Gfx` is created. So we'd need it to hold a `PhantomData<&'gfx Gfx>` like `Console` does. So it probably can't be just an enum anymore.

ctru-rs/src/gfx.rs
Meziu: Oh wow, I don't know if it's a problem in rustc, but for now I'd say we should just work on having an acceptable API (with the Screens and all). After that, we should work on testing it out.

Meziu:
Why do that, can't we just make 2 screen instances as private members of the `Gfx` struct and pass them as `Arc`s or something else to the structs (like `Console`) that need them? I didn't yet write any code about this idea, so I'm sorry if I'm talking without thinking much about the issue.

AzureMarker:
With that approach, how would a user set a graphics option on a screen, like double buffering?

AzureMarker:
One idea:
```rust
pub struct Gfx {
    pub top_screen: Screen,
    pub bottom_screen: Screen,
}

// Make it impossible for users to create a screen of their own
pub struct Screen(ScreenKind);
pub enum ScreenKind { Top, Bottom }

// Gets a &mut Screen and calls a method on it.
gfx.top_screen.set_wide_mode(true);

// Initializes a console using the top screen.
let top_console = Console::init(&gfx.top_screen);
let bottom_console = Console::init(&gfx.bottom_screen);
```

It might be nicer to use public fields on `Gfx` for screens because then we can take advantage of partial borrows (aka splitting borrows):
https://doc.rust-lang.org/nomicon/borrow-splitting.html

Meziu:
I don't even know if you can have two consoles for one screen to be honest. 

At this point I just wish their API didn't fall apart for a slightly different order of operations. But ranting is useless, so off we go.

ctru-rs/src/gfx.rs
AzureMarker: It is safe to return a raw pointer to somewhere. See the stuff I linked earlier

AzureMarker: As long as there's no way to trigger undefined behavior it's safe (just need to check that the libctru call is safe under the hood).

Meziu:
@AzureMarker I've made these changes. Sadly the borrow splitting idea won't work because taking a mutable reference to a member of a struct is still like taking a mutable reference to the whole struct, and thus calls to `Gfx` methods won't work because of held references.

This changes aren't clean, nor pretty to be honest. Tell me what you think. We'll still have some stuff to do before pushing.

AzureMarker:
Here's a few changes to simplify things and fix the wide_screen example (The other examples also need fixing, but they're straightforward):
```diff
diff --git a/ctru-rs/examples/gfx_wide_mode.rs b/ctru-rs/examples/gfx_wide_mode.rs
index f4ba365..8510dc5 100644
--- a/ctru-rs/examples/gfx_wide_mode.rs
+++ b/ctru-rs/examples/gfx_wide_mode.rs
@@ -1,7 +1,6 @@
 extern crate ctru;
 
 use ctru::console::Console;
-use ctru::gfx::Screen;
 use ctru::services::hid::KeyPad;
 use ctru::services::{Apt, Hid};
 use ctru::Gfx;
@@ -11,9 +10,9 @@ fn main() {
     let apt = Apt::init().unwrap();
     let hid = Hid::init().unwrap();
     let gfx = Gfx::default();
-    let _console = Console::init(&gfx, Screen::Top);
+    let mut console = Console::init(gfx.top_screen.borrow());
 
-    println!("Press A to enable/disable wide screen mode.");
+    println!("Press A to enable/disable wide screen mode. Extra words here so the text wraps.");
 
     while apt.main_loop() {
         hid.scan_input();
@@ -23,7 +22,17 @@ fn main() {
         }
 
         if hid.keys_down().contains(KeyPad::KEY_A) {
-            gfx.set_wide_mode(!gfx.get_wide_mode());
+            drop(console);
+
+            let mut top_screen = gfx.top_screen.borrow_mut();
+            let wide_mode = top_screen.get_wide_mode().unwrap();
+            top_screen.set_wide_mode(!wide_mode).unwrap();
+            drop(top_screen);
+
+            console = Console::init(gfx.top_screen.borrow());
+            println!(
+                "Press A to enable/disable wide screen mode. Extra words here so the text wraps."
+            );
         }
 
         gfx.flush_buffers();
diff --git a/ctru-rs/src/gfx.rs b/ctru-rs/src/gfx.rs
index 13be457..a18a9ac 100644
--- a/ctru-rs/src/gfx.rs
+++ b/ctru-rs/src/gfx.rs
@@ -1,6 +1,6 @@
 //! LCD screens manipulation helper
 
-use std::cell::{BorrowError, BorrowMutError, Ref, RefCell, RefMut};
+use std::cell::RefCell;
 use std::default::Default;
 use std::ops::Drop;
 
@@ -11,8 +11,10 @@ use crate::services::gspgpu::{self, FramebufferFormat};
 ///
 /// The service exits when this struct is dropped.
 pub struct Gfx {
-    top_screen: RefCell<Screen>,
-    bottom_screen: RefCell<Screen>,
+    pub top_screen: RefCell<Screen>,
+    pub bottom_screen: RefCell<Screen>,
+    // We don't want to allow users to create this struct manually
+    _private: ()
 }
 
 /// Available screens on the 3DS
@@ -51,29 +53,10 @@ impl Gfx {
         Gfx {
             top_screen: RefCell::new(Screen::Top),
             bottom_screen: RefCell::new(Screen::Bottom),
+            _private: ()
         }
     }
 
-    /// Try to get an immutable reference to the Top screen
-    pub fn get_top_screen(&self) -> Result<Ref<'_, Screen>, BorrowError> {
-        self.top_screen.try_borrow()
-    }
-
-    /// Try to get a mutable reference to the Top screen
-    pub fn get_top_screen_mut(&self) -> Result<RefMut<'_, Screen>, BorrowMutError> {
-        self.top_screen.try_borrow_mut()
-    }
-
-    /// Try to get an immutable reference to the Bottom screen
-    pub fn get_bottom_screen(&self) -> Result<Ref<'_, Screen>, BorrowError> {
-        self.bottom_screen.try_borrow()
-    }
-
-    /// Try to get a mutable reference to the Bottom screen
-    pub fn get_bottom_screen_mut(&self) -> Result<RefMut<'_, Screen>, BorrowMutError> {
-        self.bottom_screen.try_borrow_mut()
-    }
-
     /// Flushes the current framebuffers
     pub fn flush_buffers(&self) {
         unsafe { ctru_sys::gfxFlushBuffers() };
@@ -203,6 +186,7 @@ impl Default for Gfx {
         Gfx {
             top_screen: RefCell::new(Screen::Top),
             bottom_screen: RefCell::new(Screen::Bottom),
+            _private: ()
         }
     }
 }
```

AzureMarker:
Another idea I had is making separate types for the top and bottom screens. Both could implement one trait for common behavior, but the top screen would have extra methods like `set_3d_enabled`. This would avoid the current runtime errors and unwrapping that handles if these methods are called on the bottom screen.

Meziu:
Made the split. It works nicely now, and the structs that want a generic reference to a Screen can use `impl` or `dyn` to get it, which makes more sense when writing integrations. This looks good to me, tell me your thoughts too @AzureMarker

ctru-rs/src/gfx.rs
```diff
@@ -97,40 +119,90 @@ impl Gfx {
...
-    /// if double buffering is enabled
-    pub fn get_raw_framebuffer(&self, screen: Screen, side: Side) -> (*mut u8, u16, u16) {
+impl Screen for TopScreen {
+    fn into_raw(&self) -> ctru_sys::gfxScreen_t {
+        ctru_sys::GFX_TOP
+    }
+
+    fn set_double_buffering(&mut self, enabled: bool) {
+        unsafe { ctru_sys::gfxSetDoubleBuffering(ctru_sys::GFX_TOP, enabled) }
+    }
```
ian-h-chamberlain, line 153: I wonder if some of these could have default trait impls since they just make a call using the result of `self.into_raw()`. Would just

AzureMarker: This is a good idea. We can use `into_raw` to get the `gfxScreen_t` value.

ctru-rs/src/gfx.rs
```diff
@@ -9,17 +10,40 @@ use crate::services::gspgpu::{self, FramebufferFormat};
 /// provides helper functions and utilities for software rendering.
 ///
 /// The service exits when this struct is dropped.
-pub struct Gfx(());
+pub struct Gfx {
+    top_screen: RefCell<TopScreen>,
+    bottom_screen: RefCell<BottomScreen>,
```
AzureMarker, line 15: I recommend making these fields public like in the diff I shared. There's not much reason to hide them (since we can use something like `_private: ()` to keep the struct init private), and it saves a lot of code duplication (you're just re-declaring the borrow methods from Rc).

ctru-rs/src/gfx.rs
```diff
@@ -9,17 +10,40 @@ use crate::services::gspgpu::{self, FramebufferFormat};
...
+    /// calling screen and `Side`), the width of the framebuffer in pixels, and the height of
+    /// the framebuffer in pixels
+    ///
+    /// Note that the pointer returned by this function can change after each call to this function
+    /// if double buffering is enabled
+    fn get_raw_framebuffer(&self, side: Side) -> (*mut u8, u16, u16);
 }
 
+pub struct TopScreen;
+pub struct BottomScreen;
```
AzureMarker, line 45: These should have a private field so users can't create them:
```suggestion
pub struct TopScreen(());
pub struct BottomScreen(());
```

AzureMarker: Alternatively:
```suggestion
pub struct TopScreen {
    _private: ()
}
pub struct BottomScreen {
    _private: ()
}
```

ctru-rs/src/gfx.rs
```diff
@@ -9,17 +10,40 @@ use crate::services::gspgpu::{self, FramebufferFormat};
...
-#[derive(Copy, Clone, Debug)]
-pub enum Screen {
-    /// The top screen
-    Top,
-    /// The bottom screen
-    Bottom,
+/// Trait implemented by TopScreen and BottomScreen for common methods
+pub trait Screen {
+    /// Returns the libctru value for the Screen kind
+    fn into_raw(&self) -> ctru_sys::gfxScreen_t;
```
AzureMarker, line 21: This should be `as_raw` since it's "free" and doesn't take ownership:
```suggestion
    fn as_raw(&self) -> ctru_sys::gfxScreen_t;
```

More info about naming guidelines:
https://rust-lang.github.io/api-guidelines/naming.html

Meziu:
Fixed the problems with the latest commit, including struct access. I had thought `RefCell` wasn't a safe struct to leave public, but by a close inspection of the docs I noticed there isn't any unwanted behaviour, so now the screen fields are public.

ian-h-chamberlain (approved):

ctru-rs/src/gfx.rs
```diff
@@ -31,6 +64,15 @@ pub enum Side {
...
 }
 
+/// A handle to libctru's gfx module. This module is a wrapper around the GSPGPU service that
+/// provides helper functions and utilities for software rendering.
+///
+/// The service exits when this struct is dropped.
+pub struct Gfx {
+    pub top_screen: RefCell<TopScreen>,
+    pub bottom_screen: RefCell<BottomScreen>,
+}
```
AzureMarker, line 74: This needs a private field, otherwise users can just directly create the struct (maybe causing bad things on drop since gfx wasn't properly initialized):
```suggestion
pub struct Gfx {
    pub top_screen: RefCell<TopScreen>,
    pub bottom_screen: RefCell<BottomScreen>,
    _private: ()
}
```

ctru-rs/src/console.rs
```diff
@@ -1,27 +1,28 @@
...
+    _screen: PhantomData<&'screen ()>,
 }
 
-impl<'gfx> Console<'gfx> {
+impl<'screen> Console<'screen> {
     /// Initialize a console on the chosen screen, overwriting whatever was on the screen
     /// previously (including other consoles). The new console is automatically selected for
     /// printing.
-    pub fn init(_gfx: &'gfx Gfx, screen: Screen) -> Self {
+    pub fn init(screen: Ref<'screen, impl Screen>) -> Self {
```
AzureMarker, line 18: This should be just be a reference, since it doesn't _need_ to be a reference-counted handle:
```suggestion
    pub fn init(screen: &'screen impl Screen) -> Self {
```
Also could use `dyn` here instead of `impl`, but it doesn't really matter.

Meziu: I believe you haven't looked at my code, since it's the second time you talked about an `Rc` implementation. This isn't reference counting, this is internal mutability. Reference counting would bring us to have multiple owners for the `Screen`, which isn't what we want. This is a `RefCell`, which is a smart pointer for a mutating object. It's used only for internal mutability inside `Gfx`, so that we may use the screens/gfx separately.

Also, without using a `Ref` object, the rules for `RefCell`'s borrowing aren't correctly applied. We may even have to store the reference, now that I think about it.

Meziu:
Yes, using the lifetime alone led to being able to mutably borrow the `Screen` even if the `Console` was using it. NOW it should be all fine. Hope this is the last commit...

ctru-rs/src/console.rs
AzureMarker: Sorry, the types are named pretty similar and sometimes I mix them up. But `RefCell` does actually count references (just not in the way "reference-counted" implies):
https://doc.rust-lang.org/stable/src/core/cell.rs.html#687

The code as it was when I reviewed it though would have functioned the same if you used `Ref` or `&`. At that time, the `Ref` wasn't used, so the only functional thing happening was associating the `'screen` lifetime with the returned `Console<'screen>` instance. Which would work the same with `Ref` or `&`.

Either way the code has now been updated to hold the ref. But if we just use `Ref` then there's really no difference compared to using immutable references, with or without `PhantomData`. There's no internal mutability happening here, so removing the `RefCell` doesn't break anything - Yes, I tested this locally. Probably you meant to use `RefMut`? That would actually utilize the `RefCell` (letting you take mutable references to the fields of `Gfx` but still allowing `&self` calls to it).

AzureMarker:
> Since the Console holds the Screen anyways, you could also add the `set_wide_screen` directly to the Console object

The point of passing the screen to the Console via a `RefCell` type (and it should be `RefMut`), besides to get the actual `gfxScreen_t`, is so the screen/gfx can't be modified while the Console exists. If we added this, then you could change the screen width while a console is alive. The console wouldn't get resized (just delegated to the left of the screen), which is probably the least harmful thing that can happen.

I did some experiments though for if we did decide to do this:

If we did do this, it should be by exposing a method on `Console` to get an `&mut dyn Screen`:
```rust
pub fn screen(&mut self) -> &mut dyn Screen {
    self.screen.deref_mut()
}
```

Alternatively, we could include the screen type as a generic on `Console` and get something pretty cool:
```rust
pub struct Console<'screen, S: Screen> {
    context: Box<PrintConsole>,
    screen: RefMut<'screen, S>,
}

impl<'screen, S: Screen> Console<'screen, S> {
    // ...

    pub fn screen(&mut self) -> &mut S {
        self.screen.deref_mut()
    }
}

let mut console = Console::init(gfx.top_screen.borrow_mut());

// If we used the bottom screen this wouldn't compile
console.screen().set_wide_mode(true);
```

ctru-rs/src/console.rs
```diff
@@ -1,27 +1,27 @@
...
 use ctru_sys::{consoleClear, consoleInit, consoleSelect, consoleSetWindow, PrintConsole};
 
-use crate::gfx::{Gfx, Screen};
+use crate::gfx::Screen;
 
-pub struct Console<'gfx> {
+pub struct Console<'screen> {
     context: Box<PrintConsole>,
-    _gfx: PhantomData<&'gfx ()>,
+    _screen: Ref<'screen, dyn Screen>,
```
AzureMarker, line 10: Since we're now actually holding on to this data, we should remove the underscore:
```suggestion
    screen: Ref<'screen, dyn Screen>,
```

AzureMarker:
> NOW it should be all fine. Hope this is the last commit...

Sorry :(, but I think it's really close now, just a few more tweaks, plus making the examples compile.

Meziu:
> console.screen().set_wide_mode(true);

I was thinking at something like a `console.set_wide_mode(true)`. It looks redundant, but it could include setup code to re-init the Console object to have more columns.

Meziu:
@AzureMarker fixed and added examples. I put `hello-world` and `network-sockets` here because they needed changes for the `mut-gfx` PR anyways, but any changes or refactoring of the code should be done in another issue/PR. For now they just work, including `gfx-wide-mode`.

AzureMarker:
> > console.screen().set_wide_mode(true);
> 
> I was thinking at something like a `console.set_wide_mode(true)`. It looks redundant, but it could include setup code to re-init the Console object to have more columns.

Yeah, I haven't looked into this to see if it's possible to re-init the console or manually adjust it, but that should work.

ctru-rs/examples/gfx-wide-mode.rs
```diff
@@ -11,7 +10,7 @@ fn main() {
     let apt = Apt::init().unwrap();
     let hid = Hid::init().unwrap();
     let gfx = Gfx::default();
-    let _console = Console::init(&gfx, Screen::Top);
+    let mut _console = Console::init(gfx.top_screen.borrow_mut());
```
AzureMarker, line 13: nit: This should be `console` since it's referenced later.

ctru-rs/examples/hello-world.rs
```diff
@@ -1,56 +1,44 @@
+extern crate ctru;
```
AzureMarker, line 1: nit: This isn't needed (though I do see the other examples have it from before, when the crate used Rust 2015, so I'm just commenting on the new places it's added to)

ctru-rs/examples/hello-world.rs
```diff
@@ -1,56 +1,44 @@
...
+use ctru::gfx::Gfx;
 use ctru::services::apt::Apt;
 use ctru::services::hid::{Hid, KeyPad};
 
-fn main() {
-    // Initialize ctrulib service handles.
-    // Service handles are internally reference-counted. When all instances of a
-    // service handle go out of scope, the service will be closed.
-    ctru::init();
+extern crate ferris_says;
```
AzureMarker, line 7: nit: This isn't needed

ctru-rs/examples/network-sockets.rs
```diff
@@ -0,0 +1,66 @@
+extern crate ctru;
```
AzureMarker, line 1: nit: This isn't needed

ctru-rs/src/gfx.rs
```diff
@@ -31,6 +64,16 @@ pub enum Side {
...
 }
 
+/// A handle to libctru's gfx module. This module is a wrapper around the GSPGPU service that
+/// provides helper functions and utilities for software rendering.
+///
+/// The service exits when this struct is dropped.
+pub struct Gfx {
+    pub top_screen: RefCell<TopScreen>,
+    pub bottom_screen: RefCell<BottomScreen>,
+    _private: (),
```
AzureMarker, line 74: I ran Clippy against the code and it flagged this. TIL you can use `#[non_exhaustive]` on more than just enums:
https://rust-lang.github.io/rust-clippy/master/index.html#manual_non_exhaustive
```
warning: this seems like a manual implementation of the non-exhaustive pattern
  --> ctru-rs/src/gfx.rs:71:1
   |
71 |   pub struct Gfx {
   |   ^-------------
   |   |
   |  _help: add the attribute: `#[non_exhaustive] pub struct Gfx`
   | |
72 | |     pub top_screen: RefCell<TopScreen>,
73 | |     pub bottom_screen: RefCell<BottomScreen>,
74 | |     _private: (),
75 | | }
   | |_^
   |
   = note: `#[warn(clippy::manual_non_exhaustive)]` on by default
help: remove this field
  --> ctru-rs/src/gfx.rs:74:5
   |
74 |     _private: (),
   |     ^^^^^^^^^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_non_exhaustive
```

ctru-rs/src/gfx.rs
```diff
@@ -1,23 +1,56 @@
...
+        }
+    }
+}
+
+pub struct TopScreen {
+    _private: (),
+}
+pub struct BottomScreen {
+    _private: (),
 }
```
AzureMarker, line 54: Similar to `Gfx`, this could use `#[non_exhaustive]`, though Clippy didn't flag it.

I was curious if Clippy didn't flag it because it doesn't work, but it does work according to the RFC: https://rust-lang.github.io/rfcs/2008-non-exhaustive.html#unit-structs

AzureMarker (commented):
Very close!!! I'm looking forward to this merging. Just a few small nits. The example nits could be done separately like you mentioned. I can open a PR for these to ease the burden on you :).

ctru-rs/src/console.rs
AzureMarker: Oh, I didn't think far enough ahead, this causes an unused field warning... Well, we could leave it for now since we might use it in the future (ex. for a wrapped `set_wide_mode`). Or revert it to `_screen` if you want.

Meziu:
Solved all the suggestions

AzureMarker (approved):

ctru-rs/src/gfx.rs
AzureMarker: @Meziu @ian-h-chamberlain FYI I just learned the above rustc bug is now fixed on stable!

Issue #18, opened 2022-01-11
Port 3ds examples from devkitPro/3ds-examples


AzureMarker:
There's a bunch of stuff here that could be ported to test the crate:
https://github.com/devkitPro/3ds-examples

ian-h-chamberlain:
I've started on a rough port of the `time/rtc` example there and should have a PR before long. Figured I would post here to avoid any duplication of work.

It might be useful to use this as a tracking issue for the examples with a [task list](https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists#about-issue-task-lists) in the body of the issue, but would require creating a new issue for each example, so perhaps that's overkill.

AzureMarker:
Thanks for the heads up @ian-h-chamberlain! Let us know if you have any issues getting your environment set up.

Techie-Pi:
> It might be useful to use this as a tracking issue for the examples with a [task list](https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists#about-issue-task-lists) in the body of the issue, but would require creating a new issue for each example, so perhaps that's overkill.

From my experience, it isn't needed to create an issue for each example, you can just create a normal list and it will be shown as  a task list

hYdos:
might be a good idea to add a 3d rendering example as it seems it may need to differ a bit from C

Meziu:
> might be a good idea to add a 3d rendering example as it seems it may need to differ a bit from C

If your intention is to have graphics rendering, I suggest sticking to the raw `citro3d-sys` bindings, basically exactly as it would work in C.

Once `citro3d-rs` is finalised, we will make examples for rendering there.

PR #23, opened 2022-01-18
Add time_rtc example

ian-h-chamberlain:
Ref #18 

Finally, I was able to get an example running for `time/rtc`, which as far as I can tell is functionally the same as the devkitpro version! It took me a while to figure out I should be using `ctru_sys::*` instead of `libc::*` – with the latter I was getting garbage data from `gmtime`. Any idea why that is? I suspect it has something to do with the linkage or initialization of these two reentrancy structures, but that's just a suspicion.

```rs
extern "C" {
    pub static mut _impure_ptr: *mut _reent;
}
extern "C" {
    pub static _global_impure_ptr: *mut _reent;
}
```

I also switched the dep to be `libc = "0.2"` since there is already a patch in the workspace root and it takes precedence AFAIK. If we want a more specific version, or e.g. different `features`, I can make that change as well.

Edit: oops, I meant to attach screenshots.

## `ctru_sys` functions:
correct

## `libc` functions (plus a debug print which shows the result from `time()` seems to be correct):
incorrect

### DIFF
--- a/ctru-rs/Cargo.toml
+++ b/ctru-rs/Cargo.toml
@@ -18,3 +18,6 @@ pthread-3ds = { git = "https://github.com/Meziu/pthread-3ds.git" }
 libc = "0.2"
 bitflags = "1.0.0"
 widestring = "0.2.2"
+
+[dev-dependencies]
+ferris-says = "0.2.1"
--- a/ctru-rs/Cargo.toml
+++ b/ctru-rs/Cargo.toml
@@ -15,6 +15,6 @@ ctru-sys = { path = "../ctru-sys", version = "0.4" }
 const-zero = "0.1.0"
 linker-fix-3ds = { git = "https://github.com/Meziu/rust-linker-fix-3ds.git" }
 pthread-3ds = { git = "https://github.com/Meziu/pthread-3ds.git" }
-libc = { git = "https://github.com/Meziu/libc.git" }
+libc = { version = "0.2" }
 bitflags = "1.0.0"
 widestring = "0.2.2"
--- a/ctru-rs/Cargo.toml
+++ b/ctru-rs/Cargo.toml
@@ -15,6 +15,6 @@ ctru-sys = { path = "../ctru-sys", version = "0.4" }
 const-zero = "0.1.0"
 linker-fix-3ds = { git = "https://github.com/Meziu/rust-linker-fix-3ds.git" }
 pthread-3ds = { git = "https://github.com/Meziu/pthread-3ds.git" }
-libc = { version = "0.2" }
+libc = "0.2"
 bitflags = "1.0.0"
 widestring = "0.2.2"
--- a/ctru-rs/examples/time_rtc.rs
+++ b/ctru-rs/examples/time_rtc.rs
@@ -86,7 +86,7 @@ fn main() {
         // Get the current time. ctru_sys bindings should be used rather than
         // plain libc ones, for reasons I don't understand yet...
         let unix_time: ctru_sys::time_t = unsafe { ctru_sys::time(std::ptr::null_mut()) };
-        let time = unsafe { ptr::read(ctru_sys::gmtime(&unix_time as *const _)) };
+        let time = unsafe { *ctru_sys::gmtime(&unix_time as *const _) };
 
         let hours = time.tm_hour;
         let minutes = time.tm_min;
--- a/ctru-rs/examples/time_rtc.rs
+++ b/ctru-rs/examples/time_rtc.rs
@@ -1,0 +1,118 @@
+use std::ptr;
+
+use ctru::console::Console;
+use ctru::gfx::{Gfx, Screen};
+use ctru::services::apt::Apt;
+use ctru::services::hid::{Hid, KeyPad};
+
+const MONTHS: [&str; 12] = [
+    "January",
+    "February",
+    "March",
+    "April",
+    "May",
+    "June",
+    "July",
+    "August",
+    "September",
+    "October",
+    "November",
+    "December",
+];
+
+const WEEKDAYS: [&str; 7] = [
+    "Sunday",
+    "Monday",
+    "Tuesday",
+    "Wednesday",
+    "Thursday",
+    "Friday",
+    "Saturday",
+];
+
+const MONTH_TO_WEEKDAY_TABLE: [i32; 12] = [
+    0 % 7,   //january    31
+    31 % 7,  //february   28+1(leap year)
+    59 % 7,  //march      31
+    90 % 7,  //april      30
+    120 % 7, //may        31
+    151 % 7, //june       30
+    181 % 7, //july       31
+    212 % 7, //august     31
+    243 % 7, //september  30
+    273 % 7, //october    31
+    304 % 7, //november   30
+    334 % 7, //december   3
+];
+
+fn is_leap_year(year: i32) -> bool {
+    year % 4 == 0 && !(year % 100 == 0 && year % 400 != 0)
+}
+
+/// <http://en.wikipedia.org/wiki/Calculating_the_day_of_the_week>
+fn get_day_of_week(mut day: i32, month: usize, mut year: i32) -> usize {
+    day += 2 * (3 - ((year / 100) % 4));
+    year %= 100;
+    day += year + (year / 4);
+    day += MONTH_TO_WEEKDAY_TABLE[month]
+        - if is_leap_year(year) && month <= 1 {
+            1
+        } else {
+            0
+        };
+    (day % 7).try_into().expect("day is not valid usize")
+}
+
+fn main() {
+    ctru::init();
+
+    let gfx = Gfx::default();
+    let hid = Hid::init().expect("Couldn't obtain HID controller");
+    let apt = Apt::init().expect("Couldn't obtain APT controller");
+
+    let _console = Console::init(&gfx, Screen::Top);
+
+    print!("\x1b[30;16HPress Start to exit.");
+
+    // Main loop
+    while apt.main_loop() {
+        // Scan all the inputs. This should be done once for each frame
+        hid.scan_input();
+
+        if hid.keys_down().contains(KeyPad::KEY_START) {
+            break;
+        }
+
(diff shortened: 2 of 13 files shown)


Meziu:
I'll put this briefly on hold until #22 is closed, since that will require small changes in the `Console::init`.

ian-h-chamberlain:
Got it, thanks. Re: `libctru` vs `libc`, I think I finally figured out why it doesn't work. There is a chain of initialization which creates global state that is used by `libctru`, which we can probably hook into if we want to:

* [`initSystem` called](https://github.com/devkitPro/devkitarm-crtls/blob/master/3dsx_crt0.s#L48)
* [`__libctru_init` called](https://github.com/devkitPro/libctru/blob/4e25fb1d6c2ea124a9011c4b65f76f2968a9fb97/libctru/source/system/stack_adjust.s#L8)
* [`__system_initSyscalls` called](https://github.com/devkitPro/libctru/blob/4e25fb1d6c2ea124a9011c4b65f76f2968a9fb97/libctru/source/system/ctru_init.c#L25)
* [thread-local reentrancy struct + syscalls initialized](https://github.com/devkitPro/libctru/blob/4e25fb1d6c2ea124a9011c4b65f76f2968a9fb97/libctru/source/system/syscalls.c#L42)

I guess that calling the `libc` bindings uses the `newlib` default reentrancy pointer or something instead of the libctru-initialized one. I wonder if this would potentially cause issues with `std` compatibility in the future, since it seems like that setup is supposed to make `newlib` use the dynamic reent (i.e. the `libctru` versions of things)...

Meziu:
I don't see how `libc` could be directly linking with those addresses rather than `libctru`'s, since libc only holds external statics and function definition. If the error is in `gmtime`, the one linked is the one in the devkitARM toolchain.

ian-h-chamberlain:
> If the error is in `gmtime`, the one linked is the one in the devkitARM toolchain.

Yeah, I thought so too, but the devkitpro-included example in C worked fine, and the Rust example didn't work until I switched from using `libc` to `ctru_sys` bindings, which doesn't make much sense to me (shouldn't they both link to the same extern symbol?).

The implementation of `gmtime` itself must work, since otherwise I wouldn't be able to use `ctru_sys::gmtime` to get correct behavior. I'll see if I can find any more clues, but the difference in behavior between `libc::gmtime` and `ctru_sys::gmtime` seems to me to be like it could cause problems for `std` itself, if something used by `std` had a similar issue.

AzureMarker:
That chain of calls is pretty interesting! I found out that there's some other initialization that seems to happen automatically:
https://github.com/devkitPro/libctru/search?q=__appInit

ian-h-chamberlain:
Okay, I think I found the _actual_ cause of the problem. All that reentrancy stuff seems like a red herring, and this was a simple ABI mismatch. With this diff in `libc`:
```patch
diff --git a/src/unix/newlib/mod.rs b/src/unix/newlib/mod.rs
index b1300f169..f1d738cb0 100644
--- a/src/unix/newlib/mod.rs
+++ b/src/unix/newlib/mod.rs
@@ -28,9 +28,16 @@ pub type socklen_t = u32;
 pub type speed_t = u32;
 pub type suseconds_t = i32;
 pub type tcflag_t = ::c_uint;
-pub type time_t = i32;
 pub type useconds_t = u32;
 
+cfg_if! {
+    if #[cfg(target_os = "horizon")] {
+        pub type time_t = ::c_longlong;
+    } else {
+        pub type time_t = i32;
+    }
+}
+
 s! {
     // The order of the `ai_addr` field in this struct is crucial
     // for converting between the Rust and C types.
```

The code works as expected with `libc::` types instead of `ctru_sys` types. So I guess I can open a PR against the `libc` fork to cover this, but I'll leave the example as it stands to encourage any new users to stick with the `ctru_sys` bindings instead.

I suppose it's to be expected that bindgen would catch something like this vs handwritten bindings. To that end, I wonder if there's an easy way to automate type checking between the two and update `libc` where there is a mismatch.

AzureMarker:
Good find! Definitely make a PR against our libc fork

ian-h-chamberlain:
> I found out that there's some other initialization that seems to happen automatically:
https://github.com/devkitPro/libctru/search?q=__appInit

That is also interesting, as it seems to imply there is not truly a need for the Rust code:
```rust
let hid = Hid::init().expect("Couldn't obtain HID controller");
let apt = Apt::init().expect("Couldn't obtain APT controller");
```
That said, the API handle those provide makes sense to me, so I'm not sure this should change much in the way of "idiomatic" code. Rust is all about explicit and those appInit calls seem a bit ... magic, in my mind. But good to know about in general!

Meziu:
Yes, many `libctru` services are initialiazed (in part or completely, I believe the SD card too works like this) by the init code before execution. The problem is obviously leaving the unsafe/undroppable implementation free to use by the user, which I believe isn't what Ferris would want us to do.

Meziu:
@ian-h-chamberlain Looking back your code I noticed something a bit obvious: you are not using `std` implementations to get the current time! I know this is the code implemented by the original example, but I believe these examples should show how to have that same functionality using our Rust toolchain. I don't know how far compatibility could go, but using a crate such as `chrono` would be a fun spin on the example's implementation, while still retaining the same output format.

ian-h-chamberlain:
> you are not using std implementations to get the current time!

Yes, I think when I tried before that `chrono` uses some libc APIs that aren't provided by the devkitARM version of `newlib`. I can try again though as that was while I was in the midst of trying to get my enviroment set up. I might find other `libc` changes to be made meanwhile...

ian-h-chamberlain:
Opened https://github.com/Meziu/libc/pull/2 for libc changes

ctru-rs/Cargo.toml
```diff
@@ -15,6 +15,6 @@ ctru-sys = { path = "../ctru-sys", version = "0.4" }
 const-zero = "0.1.0"
 linker-fix-3ds = { git = "https://github.com/Meziu/rust-linker-fix-3ds.git" }
 pthread-3ds = { git = "https://github.com/Meziu/pthread-3ds.git" }
-libc = { git = "https://github.com/Meziu/libc.git" }
+libc = { version = "0.2" }
```
AzureMarker, line 18: If this is just going to specify the version, this should be:
```suggestion
libc = "0.2"
```

ctru-rs/examples/time_rtc.rs
```diff
@@ -0,0 +1,118 @@
...
+        // Scan all the inputs. This should be done once for each frame
+        hid.scan_input();
+
+        if hid.keys_down().contains(KeyPad::KEY_START) {
+            break;
+        }
+
+        // Get the current time. ctru_sys bindings should be used rather than
+        // plain libc ones, for reasons I don't understand yet...
+        let unix_time: ctru_sys::time_t = unsafe { ctru_sys::time(std::ptr::null_mut()) };
```
AzureMarker, line 88: It would be great if this used std, though I see there's some discussion about how that might require some more work...

ctru-rs/examples/time_rtc.rs
```diff
@@ -0,0 +1,118 @@
...
+        let time = unsafe { ptr::read(ctru_sys::gmtime(&unix_time as *const _)) };
```
AzureMarker, line 89: If you aren't able to use std, then maybe replace this `ptr::read` with a simple dereference?
```suggestion
        let time = unsafe { *ctru_sys::gmtime(&unix_time as *const _) };
```

ctru-rs/examples/time_rtc.rs
Meziu: It wouldn't really require more work. It requires `libc`'s PR to be merged...

ian-h-chamberlain:
Ok, I tried both `chrono` and `time` to see if we could leverage `std::time` for this example.

`time` fails at runtime, as it seems to lack an implementation for the `clock_gettime` syscall:
time_runtime


`chrono` (with `default-features = false, features = ["clock"]`) fails to build, as it seems libc is missing a field:
```
error[E0609]: no field `tm_gmtoff` on type `tm`
   --> /Users/ianchamberlain/.cargo/registry/src/github.com-1ecc6299db9ec823/chrono-0.4.19/src/sys/unix.rs:101:26
    |
101 |         let gmtoff = out.tm_gmtoff;
    |                          ^^^^^^^^^ unknown field
    |
    = note: available fields are: `tm_sec`, `tm_min`, `tm_hour`, `tm_mday`, `tm_mon` ... and 4 others
```

I could try to add `gmtoff` to the `libc` fork, but its existence in `newlib` appears to be behind a compile-time flag, and I can't seem to find whether or not devkitARM defines it (it seems like the answer is no from what I have checked).
```c
struct tm
{
  int	tm_sec;
  int	tm_min;
  int	tm_hour;
  int	tm_mday;
  int	tm_mon;
  int	tm_year;
  int	tm_wday;
  int	tm_yday;
  int	tm_isdst;
#ifdef __TM_GMTOFF
  long	__TM_GMTOFF;
#endif
#ifdef __TM_ZONE
  const char *__TM_ZONE;
#endif
};
```

Even after trying to add it to `libc`, it ends up failing for the same reason as `time`. So, it seems like we would need to either write our own implementation of `clock_gettime` to use these mainstream crates, or provide an API in `ctru-rs` to get the system time in a format compatible with one of those crates. 

Let me know what you think makes sense – it would be nice to be able to use some popular crates for stuff like this, so maybe figuring out how to implement `clock_gettime` would be the best bet (perhaps as a follow-up to this PR).

Meziu:
Wait, `clock_gettime` isn't implemented? Weird, can't we implement it in linker-fix-3ds? I am pretty sure `libctru` supports that. Still, not really related to this PR anyways. This will be merged this wag first, and then we'll see.

Meziu:
@ian-h-chamberlain make this compatible with the latest `ctru` so I can merge it.

ian-h-chamberlain:
@Meziu done, let me know if there's anything else needed. I can look into the `clock_gettime` implementation in the meantime, but I couldn't find any obvious references to it when I looked.

AzureMarker (approved):

PR #24, opened 2022-01-18
Use a module for bindings instead of `include!`

ian-h-chamberlain:
Modules are better understood by rust-analyzer compared to macro expansion, and we can still use the same bindgen script and lint attrs.

Before, "go to definition" would jump to the `include!()` call, but now it actually resolves to the `extern` declaration for functions, or the constant value for constants. I think this provides a better developer experience, since it more easily lets you find related functions, see constant values etc.

AFAIK, there is no functional difference between this and `include!()`, and I was able to build all the existing examples.


AzureMarker:
Thanks for the PR!

PR #27, opened 2022-01-27
Update time_rtc example to use `time` crate

ian-h-chamberlain:
Depends on https://github.com/Meziu/rust-linker-fix-3ds/pull/6

Couple of notes:
* This works as-is because the `time` crate does not enable its `local-offset` feature by default and therefore does not use `libc::tm::tm_gmtoff`, which is missing from the newlib definitions: https://github.com/rust-lang/libc/blob/master/src/unix/newlib/mod.rs#L105
* `chrono` seems like a slightly nicer crate to use, but it _does_ support some timezone-awareness by default, and fails to build due to the aforementioned `tm_gmtoff`. Probably a PR upstream for `chrono` to support `target_env = "newlib"` more gracefully could resolve that, but it seemed a little beyond me...

For `time` itself:
* There is a `formatting` feature which allows a format like `"[hour]:[minute]:[second]"`, but what I have now seems slightly more rust-idiomatic to me at least? I can enable that feature and use it if that seems preferable/
* I ripped out all the weekday, month code etc. since `time` provides that stuff itself.


AzureMarker:
We could add `libc::tm::tm_gmtoff` to our libc fork, if we wanted.

AzureMarker (approved):

Meziu:
> We could add `libc::tm::tm_gmtoff` to our libc fork, if we wanted.

No we can't. `Newlib` (at least devkitARM's 3DS fork) doesn't support it in it's functions. Sad, yeah, but it's normal when working with _literally any target other than linux_.

PR #28, opened 2022-01-27
Alternative panic hook exit

### LINKED ISSUE #26: Test threading
All of the "environmental" `std::thread` features (like Mutex, LocalKeys and such) work in single threaded programs, but we never tested actually spawning threads. In the remote eventuality they work, we should implement them via `pthread-3ds` and possibly remove the `thread` module here in `ctru`. 

@AzureMarker

Meziu:
Closes #26.

Requiring an input isn't flexible, at all. If the user has no `Console` to print to, they may not even know it panicked. Also, for every panicking thread created the input would be required to continue, otherwise stopping the execution of the whole program. It looks sketchy, but sleeping actually solves most of the issues. For one, the thread will always exit on panic, and won't wait indefinetly for an input. Also, since `svcThreadSleep` yields the thread to the OS, threads panicking won't take any CPU time at all, but other threads will run instead.

@AzureMarker


ian-h-chamberlain:
> Also, for every panicking thread created the input would be required to continue, otherwise stopping the execution of the whole program.

It seems to me like this might still be desirable to have for the main thread panicking, or it might be hard to read the location of the panic. Could we change behavior of the hook depending on which thread `init()` was called from, i.e. something like

```rust
    let main_thread = thread::current().id(); // does this work in the current thread impl?

    let default_hook = std::panic::take_hook();
    let new_hook = Box::new(move |info: &PanicInfo| {
        println!("\x1b[1;31m\n--------------------------------------------------");
        default_hook(info);

        if thread::current().id() == main_thread {
            loop {
                hid.scan_input();
                if hid.keys_down().contains(services::hid::KeyPad::KEY_SELECT) {
                    break;
                }
            }
        } else {
            println!("\nThe thread will exit in 5 seconds");
            thread::sleep(std::time::Duration::from_secs(5));
        }
    });
```

Actually, in this case I'm not sure you'd need the sleep at all, non-main threads could simply exit immediately but as long as the console was still alive (i.e. if `join()`ed) I think the message would stay alive. I believe this is close to the behavior on tier 1 platforms where a console is presumed to exist, from my testing.

Meziu:
> In one local version I had, I created a console in the handler itself in case of a panic before gfx initialization. I think I also had to add some code to flush / swap buffers in order to get a proper screenshot of the panic... I'll see if I still have that code stashed somewhere...

You probably can't do that  anymore since the `Console` now requires a mutable reference to `Gfx`. Since the panic! is called before exiting the `r#try`, the user-created `Gfx` would be still alive, so you won't be able to get more of those. Still, I like the "main-only" idea.

ctru-rs/src/lib.rs
```diff
@@ -11,14 +11,27 @@ pub fn init() {
...
+            let hid = services::hid::Hid::init().unwrap();
+
+            loop {
+                hid.scan_input();
+                if hid.keys_down().contains(services::hid::KeyPad::KEY_SELECT) {
+                    break;
+                }
+            }
+        }
+        println!("\x1b[1;37m\n"); // Get back to white
```
ian-h-chamberlain, line 34: Would it be a little more agnostic to use a "reset" ansi code (either `1;39` or `0`) instead of explicitly white? I haven't tested the behavior of that vs using colors...

Also there's an extra `\n` here, not sure if that's intentional

```suggestion
        println!("\x1b[0m"); // Get back to white
```

ian-h-chamberlain (approved):

ctru-rs/src/lib.rs
ian-h-chamberlain: There's a decent crate [`ansi_term`](https://docs.rs/ansi_term/latest/ansi_term/) that does all the work of remembering what codes mean and when to reset styles, etc. which is decent. I was also experimenting with [`ansi-escapes`](https://docs.rs/ansi-escapes/latest/ansi_escapes/) for cursor positioning. IIRC both of them built and worked as expected on 3DS, but it might not be desirable to add them as deps of `ctru-rs` itself

Meziu: Generally speaking it will be the user to print those, so it's at their discretion. This are the only `print!`s in the crate, we don't need to add dependencies for now.

AzureMarker:
One idea is to check if there is a current console, then if there is you can print the panic. Alternatively, don't check and just always print.

ctru-rs/src/lib.rs
```diff
@@ -11,20 +11,27 @@ pub fn init() {
...
+            let hid = services::hid::Hid::init().unwrap();
+
+            loop {
+                hid.scan_input();
+                if hid.keys_down().contains(services::hid::KeyPad::KEY_SELECT) {
+                    break;
+                }
             }
         }
+        println!("\x1b[0m"); // Get back to white
```
AzureMarker, line 34: White might not be the color they were using

Meziu: I would expect the color to be specified each time, though it doesn't solve the issue. Can we make "one-time" ANSI codes? I don't know.

P.S.
If a thread crashes, you have bigger problems than using the wrong color.

Meziu: Since ANSI is parsed as normal text, and that text (in our case) ends up in `stdout`, we can't do much about missing colors. Usually speaking, it is indeed expected from the user to use codes for each colored text, but the `Console`s don't reset the color each time, so I guess we must make a choice:

We leave it this way and leave the burden on the user *or* we don't use red, though a green error text would be jarring.

We must also consider the user is already expected to have their console screwed up one way or another, since printing with `newline`s usually make the whole screen's text wrap. Tell me what you think

Meziu:
> One idea is to check if there is a current console, then if there is you can print the panic. Alternatively, don't check and just always print.

I think that would be easy, we could just check if the pointer for the selected pointer.

ctru-rs/src/lib.rs
ian-h-chamberlain: As an experiment, I ran this code on my host system:
```rust
fn main() {
    let red = ansi_term::Color::Red;
    eprintln!("{}hello world", red.prefix());

    panic!("main thread panic");
}
```

The panic output is indeed colored red: 
Screen Shot 2022-01-27 at 3 36 06 PM

There is no special handling to "uncolor" text when printing panics to stderr on a typical OS, so one could argue that maybe there shouldn't be any code to color it either? `"\x1b[0m"` technically resets it back to unstyled text, which might _also_ not be what the user had before panicking. 

Since the console is global, any thread that prints to it could modify the colors using ANSI codes, so I'm not sure it makes sense to try and work around that...

Meziu: > Since the console is global, any thread that prints to it could modify the colors using ANSI codes, so I'm not sure it makes sense to try and work around that...

Basically this, we can either choose to *not* add to the chaos caused by the codes, or adding to it expecting some handling to exist anyways since the chaos itself exists anyways. Both approaches are equally valid. 

@AzureMarker @ian-h-chamberlain tell me what you prefer, a vote works best with an uneven number of voters.

ian-h-chamberlain: I think my vote is to remove color / special formatting altogether from the default hook and let users add it if they want. Some applications may prefer to panic to the bottom screen, or use wide mode, etc., and I think color formatting falls into the same category. In any of those use cases someone could always do this or equivalent in their app setup to achieve the same thing, right?

```rust
std::panic::set_hook(Box::new(move |info| {
    let red = ansi_term::Color::Red;
    eprint!("{}", red.prefix());
    old_hook(info);
    eprint!("{}", red.suffix());
}));
```

Meziu:
With `console_exists` it's possible to check the init status of the currently selected Console. If no Console was created, or existing implementations were dropped, the function will return false. This way the "Press select" message won't be show and the app will just exit. I didn't put the check over the whole hook because there may be a chance of retrieving `stdout` data (maybe to a File or from a debugger).

ctru-rs/src/lib.rs
Meziu: I'm also for this. No red coloring and no weird stuff. If it crashes, something is print, that's it.

Would like to hear from @AzureMarker for suggestions.

AzureMarker: Yeah removing the coloring from the ctru hook sounds good to me.

ctru-rs/examples/thread-basic.rs
```diff
@@ -13,10 +13,11 @@ fn main() {
     let hid = Hid::init().unwrap();
     let gfx = Gfx::default();
     let _console = Console::init(gfx.top_screen.borrow_mut());
+    drop(_console);
```
AzureMarker, line 16: Is this just for testing? The example code prints right after this.

ctru-rs/src/console.rs
```diff
@@ -5,6 +5,20 @@ use ctru_sys::{consoleClear, consoleInit, consoleSelect, consoleSetWindow, Print
 
 use crate::gfx::Screen;
 
+static mut EMPTY_CONSOLE: PrintConsole = unsafe { const_zero::const_zero!(PrintConsole) };
+
+pub fn console_exists() -> bool {
```
AzureMarker, line 10: This should be a static function on `Console`, i.e. `Console::exists`

ctru-rs/src/lib.rs
```diff
@@ -11,20 +11,27 @@ pub fn init() {
...
+        // Only for panics in the main thread
+        if main_thread == thread::current().id() && console::console_exists() {
+            println!("\nPress SELECT to exit the software");
+            let hid = services::hid::Hid::init().unwrap();
+
+            loop {
+                hid.scan_input();
+                if hid.keys_down().contains(services::hid::KeyPad::KEY_SELECT) {
+                    break;
+                }
```
AzureMarker, line 31: We could insert a yield here if we wanted to allow other threads to continue. I think there's a function from libctru or libc that would work?

Meziu: Threads with higher priority will continue automatically. A thread with lower priority probably depends on `main`, so at that point we may just cause a deadlock. I think it's fine this way.

Meziu:
Removed ANSI codes and formatting. Also made the function a static member of `Console` and added a comment

ian-h-chamberlain (approved):

AzureMarker (approved):

PR #29, opened 2022-01-28
Port 3ds-Example: graphics bitmap

ian-h-chamberlain:
Ref #18

I modified the framebuffer API a bit since I don't think it makes sense to allow arbitrary `Side` for the bottom screen.

Also, the frame buffer returning a tuple `(*mut u8, u16, u16)` seems a little odd to me and maybe a struct would be preferable? Let me know what you think.

Finally – do we want to roughly match the directory structure of 3ds-examples? There are _a lot_ of graphics examples and it might make sense to group them in a directory, as long as they still have unique names I think they are still easily runnable with `cargo 3ds link --example $name`. If so I can move this to a subdirectory.

Edit: screenshot! The colors on my device look a bit off, so I might try to play around with it to get a cleaner looking output... Also, for some reason, Luma screenshot isn't capturing the text on the top screen, but it is displayed there. Will investigate that.

ferris


AzureMarker:
> Also, the frame buffer returning a tuple `(*mut u8, u16, u16)` seems a little odd to me and maybe a struct would be preferable? Let me know what you think.

A struct would be nice, yeah.

> Finally – do we want to roughly match the directory structure of 3ds-examples?

Cargo doesn't automatically infer examples unless they're in the `examples` directory (not a sub-directory):
https://users.rust-lang.org/t/run-examples-in-subfolders-using-cargo/18154/2

If there are examples with more than just one file though, they can be in sub folders with a `main.rs`.

ctru-rs/src/gfx.rs
```diff
@@ -138,6 +123,44 @@ impl TopScreen {
...
     }
+
+    /// Returns a tuple containing a pointer to the top screen's framebuffer
+    /// (as determined by the `side`), the width of the framebuffer in pixels,
+    /// and the height of the framebuffer in pixels.
+    ///
+    /// Note that the pointer returned by this function can change after each call to this function
+    /// if double buffering is enabled
+    pub fn get_raw_framebuffer(&mut self, side: Side) -> (*mut u8, u16, u16) {
+        let mut width: u16 = 0;
```
AzureMarker, line 134: nit: Just a personal preference, but you don't need any of the local variable types in these functions. They get inferred from the return type.

AzureMarker:
> Luma screenshot isn't capturing the text on the top screen, but it is displayed there. Will investigate that.

I also found weird things with Luma and Console printing. It was just a black screen for me, like your top screen.

Meziu:
> Finally – do we want to roughly match the directory structure of 3ds-examples?

We can't make directories for examples. We can only group them by name (instead of a folder, put the folder name as a prefix to the file). It's not a big issue anyways.

ian-h-chamberlain:
> Cargo doesn't automatically infer examples unless they're in the examples directory (not a sub-directory):
https://users.rust-lang.org/t/run-examples-in-subfolders-using-cargo/18154/2
> 
> If there are examples with more than just one file though, they can be in sub folders with a main.rs.

> We can't make directories for examples. We can only group them by name (instead of a folder, put the folder name as a prefix to the file). It's not a big issue anyways.

Ah, I forgot these details... I suppose if it gets out of hand we could make `[[example]]` entries in `Cargo.toml`, but for now I think this convention works.

---- 

New changes I'm pushing: different conversion process for the image, which looks more correct. I guess the instructions from `3ds-examples` were outdated.

#### Before
ferris_yellow

#### After
ferris

* Added a struct `RawFrameBuffer`

I'll open a new issue for the Luma screenshot thing, since I think it might be something we can workaround on the `ctru-rs` side (I vaguely recall a code change I had which let me screenshot the console)

ctru-rs/src/gfx.rs
```diff
@@ -138,6 +136,34 @@ impl TopScreen {
...
+    /// Note that the pointer of the framebuffer returned by this function can
+    /// change after each call to this function if double buffering is enabled.
+    pub fn get_raw_framebuffer(&mut self) -> RawFrameBuffer {
+        RawFrameBuffer::for_screen_side(self.as_raw(), Side::Left.into())
+    }
+}
+
+impl RawFrameBuffer {
+    fn for_screen_side(screen: ctru_sys::gfxScreen_t, side: ctru_sys::gfx3dSide_t) -> Self {
+        let mut buf = RawFrameBuffer::default();
```
AzureMarker, line 161: Instead of creating a default and then modifying it, you can just construct the struct at the end of the function. This also avoids giving the struct a `Default` impl, which is callable by users (kind of misleading).

Meziu: Yes, remove the `Default` and just use the struct literal instead.

Meziu (changes requested):

ctru-rs/src/gfx.rs
```diff
@@ -28,29 +28,27 @@ pub trait Screen {
...
 #[non_exhaustive]
 pub struct BottomScreen;
 
+#[derive(Debug)]
+/// Representation of a framebuffer for one [`Side`] of the top screen, or the
+/// entire bottom screen. The inner pointer is only valid for one frame if double
+/// buffering is enabled. Data written to `ptr` will be rendered to the screen.
+pub struct RawFrameBuffer {
+    /// Pointer to graphics data to be rendered.
+    pub ptr: *mut u8,
```
Meziu, line 45: You should give this pointer the `Screen`'s lifetime, otherwise it would be valid after a `Gfx` drop.

Meziu (changes requested):

ctru-rs/src/gfx.rs
ian-h-chamberlain: Something like this? 
```rust
pub struct RawFrameBuffer<'screen> {
    /// Pointer to graphics data to be rendered.
    pub ptr: *mut u8,
    /// The width of the framebuffer in pixels.
    pub width: u16,
    /// The height of the framebuffer in pixels.
    pub height: u16,

    _screen: PhantomData<&'screen mut dyn Screen>,
}
```

Then the impl looks more like 
```rust
impl<'screen> RawFrameBuffer<'screen> {
    fn for_screen_side(screen: &'screen mut dyn Screen, side: ctru_sys::gfx3dSide_t) -> Self {
    }
}
```

This seems to prevent dropping the `Screen` before dropping the `FrameBuffer`, which I suppose is the goal?

ctru-rs/src/gfx.rs
```diff
@@ -28,29 +28,27 @@ pub trait Screen {
...
-    }
 }
 
 #[non_exhaustive]
 pub struct TopScreen;
+
 #[non_exhaustive]
 pub struct BottomScreen;
 
+#[derive(Debug)]
```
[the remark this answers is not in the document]
ian-h-chamberlain: I can still add it if you think it's useful in general, but adding a non-`pub` `PhantomData` member I think achieves the same thing.

ctru-rs/src/gfx.rs
Meziu: Yes, this should work.

Meziu (approved):

AzureMarker (approved):

PR #34, opened 2022-01-30
First (very rough) draft / proof-of-concept for citro3d example

ian-h-chamberlain:
I got the [simple tri example](https://github.com/devkitPro/3ds-examples/blob/master/graphics/gpu/simple_tri/source/main.c) using citro3d working! This is using [my fork](https://github.com/ian-h-chamberlain/citro3d-rs) of the `rust3ds` citro3d bindings, and it's _very_ messy, but I thought I'd show it off as a bit of progress with a screenshot:

simple_tri

My plan is to start adding more safe wrappers around citro3d, similar to how this crate is set up, with the goal that the example can be written mostly in safe code eventually.

I don't think this needs to be merged as-is, but it could be if you prefer I iterate on it in this repo instead of just locally. I'll probably start tracking issues related to citro3d in that fork and work on stuff there for the most part, then come back and update this with the new APIs.


Meziu:
I understand how cool this actually is, but please keep in mind that this is a *big* work to do. For now, we should finalize `ctru-rs`'s API and make a PR for `std`. I am eager to port `Citro3D`, but it comes after.

ian-h-chamberlain:
Totally fair point, and I'm happy to help with the effort to port `std` as well. I suppose I'll close this PR for now as it doesn't seem to provide much value other than showing "hey, we can call citro3d from Rust!"

PR #35, opened 2022-02-01
Add translations of 3DS result macros + debug impl

ian-h-chamberlain:
As @AzureMarker pointed out, we can use these to decode error return values from the libctru APIs. I also added a little more info to the `Debug` impl for `ctru_rs::Result` which should help debugging a bit in some cases.

See https://github.com/devkitPro/libctru/blob/master/libctru/include/3ds/result.h for the original definitions, which I assume are not generated by `bindgen` since they are macros and not simple `#define`s.

I intend to use these in https://github.com/Meziu/rust-linker-fix-3ds/pull/8 to translate from libctru codes to `errno` values, and make the `getrandom` behavior closer to spec.

### DIFF
--- a/ctru-rs/src/error.rs
+++ b/ctru-rs/src/error.rs
@@ -6,12 +6,13 @@ use ctru_sys::result::{R_DESCRIPTION, R_LEVEL, R_MODULE, R_SUMMARY};
 pub type Result<T> = ::std::result::Result<T, Error>;
 
 /// The error type returned by all libctru functions.
+#[non_exhaustive]
 pub enum Error {
-    Os(i32),
+    Os(ctru_sys::Result),
 }
 
-impl From<i32> for Error {
-    fn from(err: i32) -> Self {
+impl From<ctru_sys::Result> for Error {
+    fn from(err: ctru_sys::Result) -> Self {
         Error::Os(err)
     }
 }
@@ -19,15 +20,14 @@ impl From<i32> for Error {
 impl fmt::Debug for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match *self {
-            Error::Os(err) => {
-                write!(f, "libctru result code: {:08X}:", err)?;
-                f.debug_struct("")
-                    .field("description", &R_DESCRIPTION(err))
-                    .field("module", &R_MODULE(err))
-                    .field("summary", &R_SUMMARY(err))
-                    .field("level", &R_LEVEL(err))
-                    .finish()
-            }
+            Error::Os(err) => f
+                .debug_struct("Error")
+                .field("raw", &format_args!("{:#08X}", err))
+                .field("description", &R_DESCRIPTION(err))
+                .field("module", &R_MODULE(err))
+                .field("summary", &R_SUMMARY(err))
+                .field("level", &R_LEVEL(err))
+                .finish(),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


ctru-rs/src/error.rs
```diff
@@ -17,12 +19,22 @@ impl From<i32> for Error {
 impl fmt::Debug for Error {
```
AzureMarker, line 19: I can't put a comment on the right line, but we could replace all `i32` in this file with `ctru_sys::Result`. Not necessary though, but would be clearer.

ctru-rs/src/error.rs
```diff
@@ -17,12 +19,22 @@ impl From<i32> for Error {
...
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match *self {
-            Error::Os(err) => write!(f, "libctru result code: {:08X}", err),
+            Error::Os(err) => {
+                write!(f, "libctru result code: {:08X}:", err)?;
```
AzureMarker, line 23: This should either have a space at the end or be incorporated into the `debug_struct` call as a separate field (ex. `raw`).

ian-h-chamberlain: `debug_struct` adds a space by default since it normally is called with a name, e.g. `debug_struct("Foo")` becomes `Foo { .. }`, but I like the idea of a `raw` field... then maybe we just have something like
```rs
        match *self {
            Error::Os(err) => {
                f.debug_struct("Error")
                    .field("raw", &format!("{:#08X}", err))
                    .field("description", &R_DESCRIPTION(err))
                    .field("module", &R_MODULE(err))
                    .field("summary", &R_SUMMARY(err))
                    .field("level", &R_LEVEL(err))
                    .finish()
            }
        }
```

The downside of stuffing things into `field()` is they use their own default `Debug` impl, so there's not much customizing the format unless you print a string – and then, you'll get quotes around the value. But not a huge deal tbh

ian-h-chamberlain: I did find a bit of a hacky workaround, but it's not very customizable and not very nice... https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6ff1560fc0a8b9e4cf283856ed187c71

AzureMarker: By adding a space I meant after the colon before, because I think it would look like `libctru result code: -1:Error { ... }`

This also works, and is a little nicer:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ec0a9b23ebe2955318219845feac19d7
I discovered this after looking at this example: https://doc.rust-lang.org/stable/std/fmt/struct.Formatter.html#method.debug_struct

Meziu (approved):

ctru-rs/src/error.rs
```diff
@@ -1,28 +1,40 @@
...
     }
 }
 
 impl fmt::Debug for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match *self {
-            Error::Os(err) => write!(f, "libctru result code: {:08X}", err),
+            Error::Os(err) => f
+                .debug_struct("Error")
+                .field("raw", &format_args!("{:#08X}", err))
```
AzureMarker, line 25: nit: maybe prepend `0x` so it's clear this is a hex number?

ian-h-chamberlain: The `#` "alternate" marker does that automatically.

Issue #38, opened 2022-02-02
PS Service Integration in STD


Meziu:
Because `rust-linker-fix-3ds` uses the `PS` service to generate random data, the service needs to be initialised before using `getrandom`, or any type (like `HashMap`) that calls it under the impl. Now, while we can make docs about it for direct usage, this does create a problem when the function is called by the underlying implementation of an unknown crate. Should we init this service by default (like libctru already does with many services) in `ctru::init`?

P.S.
This is just a discussion thread because I don't know how to proceed about this. Tell me what you think. @AzureMarker @ian-h-chamberlain

ian-h-chamberlain:
I have a small local patch which creates a `Ps` struct that would be initialized similar to `Console`, in that you must bind it to an unused variable that lives for the duration of the program... The difference compared to `Console` is that `std` will panic if it's not initialized, instead of e.g. `println` doing nothing without a `Console`.

We could also add a feature `std` to this crate, which enables automatically `init`ing services like `Ps` that are expected to be used during the course of normal development with `std`, in which case we'd probably use `sync::Once` or something instead to initialize it, and I guess we'd just never call `psExit()`?

I'm open to different options, but I think the API being similar to `Console` sort of makes sense, plus that handle could be used for other things later on if we add functionality (like [`PS_GetLocalFriendCodeSeed()`, `PS_GetDeviceId()`](https://libctru.devkitpro.org/ps_8h.html)), which will never be called by `std`.

AzureMarker:
What's the error look like if something calls getrandom before PS is initialized? If it's informative then maybe it's not a big issue, and the user can figure it out.

Otherwise, I do like the std feature idea, but if we do that then we need to revisit how the rest of the crate handles std.

Using Once is another approach, and that might be fine if we can determine that not calling the exit function is OK.

The other PS functions are a good reason to have an explicit handle... This approach would be the best in terms of making things explicit and would allow surfacing the PS functions easily. However we need to make sure the user knows to use it (ex. When something needs randomness, like HashMap).

Meziu:
> What's the error look like if something calls getrandom before PS is initialized? If it's informative then maybe it's not a big issue, and the user can figure it out.

It's a generic "getrandom returned an unknown error".

Meziu:
[...]

PR #39, opened 2022-02-03
Add PS module for safe randomness initialization

ian-h-chamberlain:
Ref #38 

If we decide we want to do a `#[ctru::main]`, I think it would use this and probably also some other stuff that's current in `ctru::init`. Figured I'd put up the PR for this part at least.

### DIFF
--- a/ctru-rs/src/services/mod.rs
+++ b/ctru-rs/src/services/mod.rs
@@ -2,6 +2,7 @@ pub mod apt;
 pub mod fs;
 pub mod gspgpu;
 pub mod hid;
+pub mod ps;
 pub mod soc;
 pub mod sslc;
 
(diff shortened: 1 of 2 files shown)


AzureMarker (approved):

AzureMarker:
Maybe we should also add an example of using a hashmap in this PR? That would surface this service to the user a little better, and show them how to use the service.

ian-h-chamberlain:
I was thinking it might make sense as an integration test, but I can add an example since that seems more discoverable

Meziu:
Yeah, just a simple example would be best. We should open a Wiki guide for the basics once the packages goes online (like `ctru::init`, crate integrations and general know-how's).

Meziu (changes requested):
Add a simple example for HashMaps integration.

PR #41, opened 2022-02-03
Use custom test framework to `cargo test` !

### LINKED ISSUE #25: Unit Testing
I just discovered this exists: https://doc.rust-lang.org/unstable-book/language-features/custom-test-frameworks.html.
Would need integration with [cargo-3ds](https://github.com/Meziu/ctru-rs).

This issue has no priority, but it would be cool if we ever ran out of issues to work on (unlikely...).

ian-h-chamberlain:
Closes #25 , I think...

This basically plugs in the default test framework with hardcoded options
You can test it manually like this until cargo-3ds has support (.elf filename with hash may be different):

```sh
smdhtool --create ctru-rs-test "test ctru-s" "no author" /opt/devkitpro/libctru/default_icon.png target/armv6k-nintendo-3ds/debug/deps/ctru.smdh

cargo 3ds test  --no-run --lib --package ctru-rs && \
    3dsxtool --smdh=target/armv6k-nintendo-3ds/debug/deps/ctru.smdh \
        target/armv6k-nintendo-3ds/debug/deps/ctru-2d17fcbfba13d8a9.elf \
        target/armv6k-nintendo-3ds/debug/deps/ctru-test.3dsx && \
    3dslink  target/armv6k-nintendo-3ds/debug/deps/ctru-test.3dsx
```

For now I included the `Ps` changes since they seemed easy to plug into this, but I can take them out / wait for that to merge:

- [x] #39 

Things that seem to work correctly:

* Threading
* Output capture

Things that don't / won't work, as far as I know:

* Subprocessing
* Color output (we may be able to work around this)

Things I haven't tested

* Shuffling tests
* Benchmark tests
* Filtering / ignoring tests
* Timeouts
* Streamed output while running (probably would require some changes to flush the gfx while the tests are running)
* Logfile

# Screenshots
Failed Test
image

All Passed
image

### DIFF
--- a/ctru-rs/src/services/ps.rs
+++ b/ctru-rs/src/services/ps.rs
@@ -94,37 +94,48 @@ mod tests {
     fn construct_hash_map() {
         let _ps = Ps::init().unwrap();
 
-        let mut m: HashMap<i32, String> = HashMap::from_iter([
+        let mut input = vec![
             (1_i32, String::from("123")),
             (2, String::from("2")),
             (6, String::from("six")),
-        ]);
+        ];
 
-        println!("{:?}", m);
+        let map: HashMap<i32, String> = HashMap::from_iter(input.clone());
 
-        m.remove(&2);
-        m.insert(5, "ok".into());
+        let mut actual: Vec<_> = map.into_iter().collect();
+        input.sort();
+        actual.sort();
 
-        println!("{:#?}", m);
+        assert_eq!(input, actual);
     }
 
     #[test]
-    #[should_panic]
     fn construct_hash_map_no_rand() {
         // Without initializing PS, we can't use `libc::getrandom` and constructing
         // a HashMap panics at runtime.
+        //
+        // If any test case successfully creates a HashMap before this test,
+        // the thread-local RandomState in std will be initialized. We spawn
+        // a new thread to actually create the hash map, since even in multi-threaded
+        // test environment there's a chance this test wouldn't panic because
+        // some other test case ran before it.
+        //
+        // One downside of this approach is that the panic handler for the panicking
+        // thread prints to the console, which is not captured by the default test
+        // harness and prints even when the test passes.
+        crate::thread::Builder::new()
+            .stack_size(0x20_0000)
+            .spawn(|| {
+                let map: HashMap<i32, String> = HashMap::from_iter([
+                    (1_i32, String::from("123")),
+                    (2, String::from("2")),
+                    (6, String::from("six")),
+                ]);
 
-        let mut m: HashMap<i32, String> = HashMap::from_iter([
-            (1_i32, String::from("123")),
-            (2, String::from("2")),
-            (6, String::from("six")),
-        ]);
-
-        println!("{:?}", m);
-
-        m.remove(&2);
-        m.insert(5, "ok".into());
-
-        println!("{:#?}", m);
+                dbg!(map);
+            })
+            .unwrap()
+            .join()
+            .expect_err("should have panicked");
     }
 }
--- a/ctru-rs/src/services/ps.rs
+++ b/ctru-rs/src/services/ps.rs
@@ -27,3 +27,48 @@ impl Drop for Ps {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::collections::HashMap;
+
+    use super::*;
+
+    #[test]
+    fn construct_hash_map() {
+        let _ps = Ps::init().unwrap();
+
+        let mut m: HashMap<i32, String> = HashMap::from_iter([
+            (1_i32, String::from("123")),
+            (2, String::from("2")),
+            (6, String::from("six")),
+        ]);
+
+        println!("{:?}", m);
+
+        m.remove(&2);
+        m.insert(5, "ok".into());
+
+        println!("{:#?}", m);
+    }
+
+    #[test]
+    #[should_panic]
+    fn construct_hash_map_no_rand() {
+        // Without initializing PS, we can't use `libc::getrandom` and constructing
+        // a HashMap panics at runtime.
+
+        let mut m: HashMap<i32, String> = HashMap::from_iter([
+            (1_i32, String::from("123")),
+            (2, String::from("2")),
+            (6, String::from("six")),
+        ]);
+
+        println!("{:?}", m);
+
+        m.remove(&2);
+        m.insert(5, "ok".into());
+
+        println!("{:#?}", m);
+    }
+}
--- a/ctru-rs/src/test_runner.rs
+++ b/ctru-rs/src/test_runner.rs
@@ -1,0 +1,46 @@
+//! Custom test runner for building/running unit tests on the 3DS.
+
+extern crate test;
+
+use test::TestFn;
+
+use crate::console::Console;
+use crate::gfx::Gfx;
+use crate::services::hid::{Hid, KeyPad};
+
+/// A custom runner to be used with `#[test_runner]`. This simple implementation
+/// runs all tests in series, "failing" on the first one to panic (really, the
+/// panic is just treated the same as any normal application panic).
+pub(crate) fn test_runner(test_cases: &[&test::TestDescAndFn]) {
+    crate::init();
+
+    let gfx = Gfx::default();
+    let hid = Hid::init().expect("Couldn't obtain HID controller");
+    let _console = Console::init(gfx.top_screen.borrow_mut());
+
+    // TODO: may want to use some more features of standard testing framework,
+    // like output capture, filtering, panic handling, etc.
+    // For now this is works without too much setup.
+    for test_info in test_cases {
+        if let TestFn::StaticTestFn(testfn) = test_info.testfn {
+            println!("Running test {}", test_info.desc.name);
+            testfn();
+        } else {
+            println!(
+                "unsupported test type for {}: {:?}",
+                test_info.desc.name, test_info.testfn
+            );
+        }
+    }
+
+    println!("All tests passed! Press START to exit.");
+
+    // TODO: do we need apt.main_loop() here?
+    loop {
+        hid.scan_input();
+
+        if hid.keys_down().contains(KeyPad::KEY_START) {
+            break;
+        }
+    }
+}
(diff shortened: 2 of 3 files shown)


ctru-rs/src/test_runner.rs
```diff
@@ -0,0 +1,135 @@
...
+            testfn: TestFn::StaticBenchFn(f),
+            desc: test.desc.clone(),
+        },
+        _ => panic!("non-static tests passed to test::test_main_static"),
+    }
+}
+
+/// The following functions are stubs needed to link the test library,
+/// but do nothing because we don't actually need them for it to work (hopefully).
+// TODO: move to linker-fix-3ds ?
```
Meziu, line 102: I wouldn't. If we ever found a way to implement those, then we should. But keeping them there just to be unusable isn't helpful.

Meziu:
> This basically plugs in the default test framework with hardcoded options
> You can test it manually like this until cargo-3ds has support (.elf filename with hash may be different)

We should do that first. There is surely a way to get the target path of the test.

ian-h-chamberlain:
> > This basically plugs in the default test framework with hardcoded options
> > You can test it manually like this until cargo-3ds has support (.elf filename with hash may be different)
> 
> We should do that first. There is surely a way to get the target path of the test executable.

Ok, I think I've found the relevant details needed to make it work in `cargo-3ds`. I will make that my next priority after the `Ps` stuff is wrapped up.

ctru-rs/src/services/ps.rs
```diff
@@ -0,0 +1,74 @@
...
+            (2, String::from("2")),
+            (6, String::from("six")),
+        ]);
+
+        println!("{:?}", m);
+
+        m.remove(&2);
+        m.insert(5, "ok".into());
+
+        println!("{:#?}", m);
```
AzureMarker, line 52: For a test, this has a suspicious lack of asserts :wink:. Maybe add an assert at the end here just to make sure the hash map contents are as expected.

ctru-rs/src/services/ps.rs
```diff
@@ -0,0 +1,74 @@
...
```
AzureMarker, line 72: This section can be removed.

ctru-rs/src/test_runner.rs
AzureMarker: I agree, but I think the pthread function should be moved.

We may find down the road that some of the others need to be moved (they wouldn't normally link since they are only added in test binaries at the moment, so we could see a link error if one's needed). We can move them later as needed.

ctru-rs/src/test_runner.rs
```diff
@@ -0,0 +1,135 @@
+//! Custom test runner for building/running unit tests on the 3DS.
+
+extern crate test;
```
AzureMarker, line 3: Is this necessary?

ctru-rs/src/test_runner.rs
```diff
@@ -0,0 +1,135 @@
...
+    let opts = TestOpts {
+        list: false,
+        filters: Vec::new(),
+        filter_exact: false,
+        // Forking is not supported
+        force_run_in_process: true,
+        exclude_should_panic: false,
+        run_ignored: RunIgnored::No,
+        run_tests: true,
+        // Benchmarks are not supported
```
AzureMarker, line 41: What causes this?

ctru-rs/src/test_runner.rs
```diff
@@ -0,0 +1,135 @@
...
+        bench_benchmarks: false,
+        logfile: None,
+        nocapture: false,
+        // TODO: color doesn't work because of TERM/TERMINFO.
+        // With RomFS we might be able to fake this out nicely...
+        color: ColorConfig::AlwaysColor,
```
AzureMarker, line 47: How would this work? Also, shouldn't we be disabling color here instead of force-enabling it?

ctru-rs/src/test_runner.rs
```diff
@@ -0,0 +1,135 @@
...
+        format: OutputFormat::Pretty,
+        shuffle: false,
+        shuffle_seed: None,
+        // tweak values? This seems to work out of the box
+        test_threads: Some(3),
```
AzureMarker, line 52: I thought we disabled std thread spawning by breaking the pthread impl?
(I would like to keep std thread spawning, but also there's a good reason to avoid it)

ctru-rs/src/test_runner.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    gfx.flush_buffers();
+    gfx.swap_buffers();
+
+    loop {
+        hid.scan_input();
+
+        if hid.keys_down().contains(KeyPad::KEY_START) {
+            break;
+        }
```
AzureMarker, line 72: You should add a `gfx.wait_for_vblank()` here so it doesn't hog the CPU in a busy loop.

Meziu:
We absolutely need this, and have to start putting a lot of tests in `ctru`. As we saw with ThreadLocals, we can't be sure of changes because they have a massive impact on the whole ecosystem.

ctru-rs/src/services/ps.rs
ian-h-chamberlain: Sure, I was using it to test stdout capturing in tests, but I can remove it now.

ctru-rs/src/test_runner.rs
ian-h-chamberlain: Yes, for example: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=a37a055f3870e55bf55ba574a5ae6b86

At first I wasn't sure why, but I found this: https://doc.rust-lang.org/reference/names/preludes.html#extern-prelude 
The `test` crate is special since it is provided by rustc but not included in the extern prelude, so it must be declared with `extern crate` to be used, unlike typical Cargo crates.

ctru-rs/src/test_runner.rs
ian-h-chamberlain: Simply didn't test them, they might work? I figured benchmarking would need some changes to the options here and likely would use a different runner or something like that...

ctru-rs/src/test_runner.rs
ian-h-chamberlain: > Also, shouldn't we be disabling color here instead of force-enabling it?

Perhaps yes, or just setting it to "auto" which will I think leave it disabled.

> How would this work? 

The `test` implementation, on unix-like OSes, uses the `TERM` env var to learn what terminal type is being used, and `TERMINFO` to look up that terminal type in the [`terminfo` database](https://man7.org/linux/man-pages/man5/terminfo.5.html).

I think if we set the variables appropriately and add a terminfo file to the RomFS, we can get the test library to treat the console like a normal ANSI console, which would include hopefully include nice colored output. This is definitely a "future enhancement" kind of thing, though and shouldn't block this PR as far as I'm concerned.

ctru-rs/src/test_runner.rs
ian-h-chamberlain: Yeah, I wasn't really expecting this to work right away... maybe it's safer to disable concurrency in the test runner at least until #45 is resolved?

ian-h-chamberlain:
Sorry for the lack of updates here, I was away for the weekend and didn't have any time to work on it. Hoping to get to update this PR to address concerns this week and try to get this merged.

ctru-rs/src/test_runner.rs
ian-h-chamberlain: Okay, I figured it out. After `cargo update` this no longer seems to work like it did before, so I will swap to `None` instead. We can look at re-enabling after [#45](https://github.com/Meziu/ctru-rs/issues/45) is settled.

ian-h-chamberlain:
Pushed a set of changes based on feedback, I still need to do the `cargo-3ds` work, but it should at least build as-is with all the crates updated to the latest git versions.

ctru-rs/src/services/ps.rs
```diff
@@ -83,3 +83,59 @@ impl Drop for Ps {
...
+                    (1_i32, String::from("123")),
+                    (2, String::from("2")),
+                    (6, String::from("six")),
+                ]);
+
+                dbg!(map);
+            })
+            .unwrap()
+            .join()
+            .expect_err("should have panicked");
```
ian-h-chamberlain, line 139: This is a bit weird, hence the large comment, and relies on some implementation details of `std`, which I don't love. Here is the output I was referencing:


tests_Pass

Perhaps it makes more sense to remove this test case entirely, since we don't necessarily care if `HashMap::new` panics, but we do want the positive test case of being able to construct one when we have initialized PS. Let me know what you think or if you have a cleaner suggestion.

ctru-rs/src/test_runner.rs
AzureMarker: See https://github.com/Meziu/pthread-3ds/pull/8

ctru-rs/src/test_runner.rs
AzureMarker: I'd be OK with changing the comment to something like "Benchmarks are not currently supported"

ctru-rs/src/services/ps.rs
AzureMarker: Looks like this will be fixed once we move to std threads: https://github.com/rust-lang/rust/pull/78227

I think we should keep this test regardless.

ctru-rs/src/test_runner.rs
```diff
@@ -0,0 +1,126 @@
...
+    let gfx = Gfx::default();
+    let hid = Hid::init().unwrap();
+    let apt = Apt::init().unwrap();
+
+    let mut top_screen = gfx.top_screen.borrow_mut();
+    top_screen.set_wide_mode(true);
+    let _console = Console::init(top_screen);
+
+    // Start printing from the top left
+    print!("\x1b[1;1H");
```
AzureMarker, line 29: Question: what happens if we don't have this line?

ctru-rs/src/test_runner.rs
AzureMarker: Oh nvm, I saw that the comment was already updated.

AzureMarker (approved):
I haven't gotten a chance to actually test this, but it can merge before I get to it.

ctru-rs/src/test_runner.rs
ian-h-chamberlain: Hmm, good question. I think I added it when I first started testing, because there was some panic and it overwrote in the same position on the console... I tested again just now, and it seems to look okay without it, so I can probably just remove it now.

PR #42, opened 2022-02-05
Basic Tokio example

AzureMarker:
This PR adds a Tokio (futures) example. ~~It currently doesn't work :disappointed: (crashes or other odd behaviors).~~ IT WORKS!!

This needs a few changes in rustc and pthread-3ds:
- [x] https://github.com/Meziu/rust-horizon/pull/8
- [x] https://github.com/Meziu/pthread-3ds/pull/5
- [x] https://github.com/Meziu/pthread-3ds/pull/6

@Meziu let's move discussion about this example from #36 to here.

### DIFF
--- a/ctru-rs/Cargo.toml
+++ b/ctru-rs/Cargo.toml
@@ -26,6 +26,7 @@ toml = "0.5"
 [dev-dependencies]
 ferris-says = "0.2.1"
 time = "0.3.7"
+tokio = { version = "1.16", features = ["rt", "time", "sync", "macros"] }
 
 [features]
 default = ["romfs"]
--- a/ctru-rs/examples/futures-tokio-basic.rs
+++ b/ctru-rs/examples/futures-tokio-basic.rs
@@ -28,16 +28,21 @@ fn main() {
 
     let runtime_thread = ctru::thread::Builder::new()
         .affinity(1)
+        .stack_size(0x200000)
         .spawn(move || {
             runtime.block_on(async move {
+                println!("Start of future");
                 let mut wake_time = tokio::time::Instant::now() + Duration::from_secs(1);
+                let mut iteration = 0;
                 loop {
+                    println!("Start of loop");
                     let sleep_future = tokio::time::sleep_until(wake_time);
 
                     tokio::select! {
                         _ = &mut exit_receiver => break,
                         _ = sleep_future => {
-                            println!("Tick");
+                            println!("Tick {}", iteration);
+                            iteration += 1;
                             wake_time += Duration::from_secs(1);
                         }
                     }
--- a/ctru-rs/examples/futures-tokio-basic.rs
+++ b/ctru-rs/examples/futures-tokio-basic.rs
@@ -26,15 +26,15 @@ fn main() {
         .expect("Couldn't build runtime");
 
     let runtime_thread = ctru::thread::Builder::new()
+        // Run on the system core
         .affinity(1)
+        // Use a bigger stack size. Default is 0x1000 but we'd easily overflow that.
         .stack_size(0x200000)
         .spawn(move || {
             runtime.block_on(async move {
-                println!("Start of future");
                 let mut wake_time = tokio::time::Instant::now() + Duration::from_secs(1);
                 let mut iteration = 0;
                 loop {
-                    println!("Start of loop");
                     let sleep_future = tokio::time::sleep_until(wake_time);
 
                     tokio::select! {
--- a/ctru-rs/examples/futures-tokio-basic.rs
+++ b/ctru-rs/examples/futures-tokio-basic.rs
@@ -1,5 +1,6 @@
 use ctru::console::Console;
 use ctru::services::hid::KeyPad;
+use ctru::services::ps::Ps;
 use ctru::services::{Apt, Hid};
 use ctru::Gfx;
 use std::time::Duration;
@@ -9,10 +10,8 @@ fn main() {
     let gfx = Gfx::default();
     let hid = Hid::init().expect("Couldn't obtain HID controller");
     let apt = Apt::init().expect("Couldn't obtain APT controller");
+    let _ps = Ps::init().expect("Couldn't initialize PS service");
     let _console = Console::init(gfx.top_screen.borrow_mut());
-
-    // FIXME: replace this with `Ps` when #39 merges
-    assert!(unsafe { ctru_sys::psInit() } >= 0);
 
     // Give ourselves up to 30% of the system core's time
     apt.set_app_cpu_time_limit(30)
--- a/ctru-rs/examples/futures-tokio-basic.rs
+++ b/ctru-rs/examples/futures-tokio-basic.rs
@@ -38,12 +38,15 @@ fn main() {
                     let sleep_future = tokio::time::sleep_until(wake_time);
 
                     tokio::select! {
-                        _ = &mut exit_receiver => break,
+                        // Use the first available future instead of randomizing
+                        biased;
+
                         _ = sleep_future => {
                             println!("Tick {}", iteration);
                             iteration += 1;
                             wake_time += Duration::from_secs(1);
                         }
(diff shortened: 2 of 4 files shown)


Meziu:
Ehm, your changes (probably those to `LocalKey`) completely broke `panic!`. When panicking, the system will deadlock. :cry:
Fix your new changes first, then we can check for the `tasks`.

AzureMarker:
Really? I don't see that behavior. Maybe you need to run `cargo update`. Right now with all up to date stuff, I get an abort (ARM11 data abort exception) almost immediately with the debug build. The release build works except "Tick" isn't printed (pressing Start cleanly exits).

I'm not sure how a deadlock can occur, since the lock is only held for the function and only one lock is ever taken in one function (no lock A and then lock B).

AzureMarker:
> I'm circling around the problem. It looks to be in `tokio-1.16.1/src/macros/scoped_tls.rs:50`:
> 
> ```rust
>         let prev = self.inner.with(|c| {
>             let prev = c.get();
>             c.set(t as *const _ as *const ());
>             prev
>         });
> ```
> 
> This is a very specific function, and it uses `std::thread::local::LocalKey` impls. Before this function, the `core.tasks` vector is empty (you can check it here via `gdb` with `p f.core.tasks`), but right after it results completely full and with a broken `size` value. I bet the `LocalKey` implementation screws up the pointers and points to unrelated memory, but I don't know how.

I still see the task vector as empty after running this snippet (in a debug build). Still walking through the execution, but I did see it successfully print Tick once in a release mode during debugging (but it didn't work when not debugging??).

AzureMarker:
OK, so I found the line where the `tick` field in tokio's runtime core changes. It was 0 before we add the new thread key to this BTreeMap:
image

Now there's even an error when looking at the core, which is really strange. The BTreeMap got updated with the right value (`5 -> Some(0x119cc8)`).

AzureMarker:
So this is pretty weird... At some point during BTreeMap insert, it calls `len` on a node. That should be fine, it's just reading data, right? WRONG. This is where the tick field changes. I set a hardware breakpoint and it triggered during a call to `as_leaf_ptr`:
image

This is how it's called. Notice that the `self` variable holds a valid-looking pointer (`self as *const ()`), which is good. But if you look at where it is on the stack (`&self as *const ()`), that's very close to where the core is stored!
0x800dfcc (NodeRef variable on stack) vs 0x0800df80 (Tokio core location)
image

It looks like the stack is getting too big, and it overflows into the heap??? This is weird.

Meziu:
How could that be?? I don’t even know if that is something the stack *could* do. Also, how could the stack be anywhere near full capacity after just a couple of init functions?

AzureMarker:
So libctru just allocates space for the thread's TLS and stack in the heap:
https://github.com/devkitPro/libctru/blob/4d36c74be427d34ff8c7b06146893dccef43dcd5/libctru/source/thread.c#L54

So it makes sense why the stack is so close to the heap. My thread's data gets allocated at 0x800e300 (size = 0x2094 bytes):
image

Based on libctru's thread code, the thread allocation looks like this:
| Offset | Size | Usage |
|-|-|-|
| 0x800e300..0x800E77F | 0x480 bytes  | libctru `Thread_tag` data |
| 0x800E780..0x800F77F | 0x1000 bytes | Thread stack |
| 0x800F780..0x8010393 | 0xC14 bytes | Thread local storage |

So the top of the stack (where it would overflow) is 0x800E780, and the bottom is 0x800F780.
Interestingly, the libctru `Thread_tag` `stacktop` field is misnamed. It should be `stackbottom` since that's where the stack starts at. But stack naming is weird, so maybe they think of the stack as growing downward (which it kind of does - this is all very confusing).
image

Interestingly, the main thread is also allocated in the heap based on the `$sp` value, but I think that makes sense because we're running under Homebrew Menu.

By the time the new thread gets to the closure we passed in, we've already blown through about half of our stack ($sp = 0x800ed40, 0xA40 = 2624 bytes left):
image

And when we get to the `ScopedKey` call, we've just blown our stack ($sp = 0x800e768, a few bytes past 0x800e780):
image

The stack pointer in the previous frame was 0x800e7b0, so still valid.

But the program keeps going, so eventually it will reach Tokio's `tokio::runtime::basic_scheduler::Core` allocation, which spans from 0x800df80 to 0x800DFD0 (80 bytes). You can see here is when the stack starts overwriting parts of that data structure ($sp = 0x800dfc0, 16 bytes into the structure):
image

The default stack size in libctru (and ctru-rs) is 4096 or 0x1000 bytes. In comparison, the default stack size in Rust's std threads is 2MiB (main thread is OS-specific):
https://doc.rust-lang.org/stable/std/thread/index.html#stack-size

So, let's try bumping up the stack size to 2MiB (0x200000 bytes). We should also think about ways to trigger a stack overflow when one occurs, but this should work for now.

Aaaand... it doesn't crash? But it doesn't print "Tick" either... More debugging is necessary :smiley:.

AzureMarker:
> How could that be?? I don’t even know if that is something the stack _could_ do. Also, how could the stack be anywhere near full capacity after just a couple of init functions?

Didn't see your message until I posted my wall of text :laughing:. Hopefully that helps explain things.

Meziu:
> We should also think about ways to trigger a stack overflow when one occurs, but this should work for now.

Std Threads use a special struct under unix.
Look at [
](https://github.com/rust-lang/rust/blob/88fb06a1f331926bccb448acdb52966fd1ec8a92/library/std/src/sys/unix/thread.rs#L106)

Though, I don’t know how to use that on our threads (both because of possible lack of Newlib implementations, and because it is a private member of the standard lib, so it may require re-writing it).

AzureMarker:
> Though, I don’t know how to use that on our threads (both because of possible lack of Newlib implementations, and because it is a private member of the standard lib, so it may require re-writing it).

If all of the underlying syscalls/functionality is available, we can copy the implementation into ctru-rs like what was done for most of the thread module.

AzureMarker:
In the latest iteration, I think it's kind of working. I don't see the "Tick" messages, but if you wait a little bit before clicking Start to end the program, right before it ends there's a flash of Tick messages in the console. It's almost like the thread's stdout isn't flushed or something...

AzureMarker:
I've been debugging it, and I think the timer is just not waking up the runtime thread as expected. If I step through the program in gdb it all seems to work. But once I turn off breakpoints or run it without a debugger attached, it acts like the runtime thread never wakes up. Clicking Start to exit the program wakes it up, and it runs through all of the remaining ticks until it exits.

So there's something weird with the timer. We might need to check our condvar implementation, or something else. Needs more debugging.

There's also an interesting piece where `tokio::select` by default chooses a random branch if two of the futures are ready. This could account for some of the randomness seen on exit (how many ticks you get a flash of, or if you see any at all). Luckily there's a way to turn off that randomness, which I've just committed.

AzureMarker:
So... uh... 1644094578314000000 nanoseconds is just over 52 years 
image

I think I know why it's not waking up now.

AzureMarker:
Ah, I think the timeout that we get is supposed to be the UNIX timestamp that we should timeout at. But libctru expects the timeout to be a duration. Should be an easy fix.

AzureMarker:
So there actually IS a deadlock with the pthread keys implementation. During creation of a key, the map gets locked, but during the `RwLockWriteGuard` creation it checks if it's currently panicking (part of poison check), which then checks a thread-local variable, which needs to be created, so it creates a new pthread key, which needs to write to the KEYS map, which deadlocks.

I think a reentrant lock would handle this, or something that doesn't use poisoning like parking_lot.

AzureMarker:
Needs https://github.com/Meziu/pthread-3ds/pull/6, but then it's all fixed!

Meziu:
Wow, really wow. This is amazing. Thank you for all your work :pray:. Having tokio at our disposal means that multithreading just became a LOT easier. I and anyone else using our toolchain will be very grateful. :clap:

PR #47, opened 2022-02-12
Add support for reading files to file-explorer

ian-h-chamberlain:
Ref https://github.com/Meziu/rust-horizon/issues/11
edit: linked wrong issue because #11 on this repo is also related to filesystem, lol

Issues I've found so far:
* `PathBuf::parent` / `PathBuf::pop` seems to have strange behavior with the `romfs:/` prefix. See my comment on https://github.com/Meziu/rust-horizon/issues/11
* File size is definitely borked for romfs. On my system, all files and directories in the romfs have the same size, but it's different per invocation of the app... I'm guessing some kind of UB is happening here
* Due to above issue, `fs::read_to_string` doesn't work. Perhaps ideally, we'd have a test for both in `std` or in this library.

Tasks to probably address before merging:
- [x] https://github.com/Meziu/rust-horizon/pull/12
- [x] https://github.com/Meziu/libc/pull/6
- [ ] [Decide file path conventions](https://github.com/Meziu/rust-horizon/issues/13) and/or fix `parent()`, `components()`, `pop()` ? (`romfs:/` and friends) – probably not actually needed before merging
- [x] Switch to `fs::read_to_string`? Or at least add a test for it somewhere?

@Meziu @AzureMarker


Meziu:
Those Redox issues are quite old, has nothing of those made it into `std`? I wonder what the Redox users are doing to prevent that...

ctru-rs/examples/file-explorer.rs
```diff
@@ -74,8 +75,37 @@ impl<'a> FileExplorer<'a> {
     }
 
     fn print_menu(&mut self) {
-        println!("Viewing {}", self.path.display());
+        match std::fs::metadata(&self.path) {
```
AzureMarker, line 78: I see this error after reading `romfs:/test-file.txt`:
```
Failed to read romfs:: Illegal byte sequence (os error 138)
```

This is probably due to the weird path handling? Maybe we should handle this case in the code until the underlying path issue is fixed?

ctru-rs/examples/file-explorer.rs
```diff
@@ -85,34 +115,49 @@ impl<'a> FileExplorer<'a> {
...
             }
         }
+    }
 
-        println!("Start to exit, A to select an entry by number, B to go up a directory, X to set the path.");
+    fn print_file_contents(&mut self) {
+        match std::fs::read_to_string(&self.path) {
+            Ok(contents) => {
+                println!("File contents:\n{0:->80}", "");
+                println!("{contents}");
```
AzureMarker, line 133: Interestingly, for a really long file this takes a noticeably long amount of time (30 seconds on my `/retroarch/retroarch.cfg`), during which you see the lines scroll past pretty quickly. I wonder if the libctru console implementation is slow or something?

AzureMarker, line 133: Maybe we should paginate the output here too? I wonder if we should move the pagination message to the lower screen to avoid interrupting the top screen's output.

ctru-rs/examples/file-explorer.rs
ian-h-chamberlain: Yes, this is because of #13 as far as I know. I didn't want it to block this PR because I think it was an issue before but got a more exposed by this PR.

ctru-rs/examples/file-explorer.rs
ian-h-chamberlain: Oops, good point, I think when I had BufReader here it made more sense but using `read_to_string` is probably questionable when trying to paginate. Using the bottom screen to prompt for it probably takes a little work but sounds like a potentially nicer UX for this example.

ian-h-chamberlain: How big is that file, out of curiosity? I only tested with this small one but anything on the order of 30s seems worth investigating. In particular I would not expect romfs to be slow, but I haven't looked too close at the io implementations for it...

ctru-rs/examples/file-explorer.rs
AzureMarker: Yeah, not blocking, but maybe we can do something like in `run` with `self.path.components().count() > 1` to avoid popping the path in this case. Or maybe we just need to reset it manually. Up to you.

ctru-rs/examples/file-explorer.rs
AzureMarker: I think the prompt move should be as making a new temporary console and selecting it.

AzureMarker: It's 104 kB. I think it's just the default config, but here is the exact file (compressed so GitHub accepts it):
[retroarch_config.zip](https://github.com/Meziu/ctru-rs/files/8057564/retroarch_config.zip)

Meziu: Yep, it looks like that they _really_ needed a scroll effect, but since the console  cooperative thread doesn’t yield the scroll blocks the whole app, and takes quite a while, too.

AzureMarker: Oh, the scroll effect is semi-intentional? I guess fixing that would require libctru changes...

Issue #43, opened 2022-02-06
Provide a custom getrandom (crate) implementation


AzureMarker:
The [getrandom](https://github.com/rust-random/getrandom) crate fails to compile for unsupported systems (they don't blanket support `unix`-like systems), but it has a mechanism we can use to register a custom backing implementation:
https://docs.rs/getrandom/latest/getrandom/#unsupported-targets
https://docs.rs/getrandom/latest/getrandom/macro.register_custom_getrandom.html

We could implement this in ctru-rs and users would be able to use getrandom without issues.

Alternatively (and maybe in parallel), we could make a PR to getrandom to add support. We'd have to make sure to only depend on `libc::get_random` and not libctru.

Meziu:
We could make a module for third-party crate implementations, that looks to be a common thing we’ll have to do. Still, I bet that after getting an STD PR merged other crate distributors will be ok with the further support.

AzureMarker:
BTW one reason we'd want this is to inspect running programs using tokio-console:
https://github.com/tokio-rs/console/tree/main/console-subscriber

Meziu:
```
[Registering a custom getrandom implementation](https://docs.rs/getrandom/latest/getrandom/macro.register_custom_getrandom.html#registering-a-custom-getrandom-implementation)

Functions can only be registered in the root binary crate. Attempting to register a function in a non-root crate will result in a linker error. This is similar to [#[panic_handler]](https://doc.rust-lang.org/nomicon/panic-handler.html) or [#[global_allocator]](https://doc.rust-lang.org/edition-guide/rust-2018/platform-and-target-support/global-allocators.html), where helper crates define handlers/allocators but only the binary crate actually uses the functionality.
```

From [getrandom's docs](https://docs.rs/getrandom/latest/getrandom/macro.register_custom_getrandom.html). Makes sense honestly. It looks like they are keen on supporting new targets though. We should PR once the `std` is merged.

AzureMarker:
> We should PR once the `std` is merged.

We could technically make a PR beforehand since it's a no_std crate and just uses libc. But std is more of a priority.

AzureMarker:
@Meziu I'm working on a getrandom PR btw.

AzureMarker:
Opened https://github.com/rust-random/getrandom/pull/248

Meziu:
[...]

PR #49, opened 2022-02-23
Static init of the PS service

Meziu:
Related to #43.

With these changes, the `PS` service will run for the whole app runtime, and be destroyed at the program's end. This way all code using `libc::getrandom` (and thus, also `std::collections::HashMap`) will work by default when using `ctru::init` at the start of the program.

A similar problem to this would still arise when using `std::net` and other networking functionality without using the `Soc` service, but that one is way heavier than `PS` and it's only needed in very specific cases the user should be aware of.

Edit: Also, since these changes rend the idea of a `Ps` struct both useless and detrimental (like closing the service while some program process still needs it), the functions have been made statics of the `ctru::ps` module and the struct has been removed.

@AzureMarker

### DIFF
--- a/ctru-rs/examples/hashmaps.rs
+++ b/ctru-rs/examples/hashmaps.rs
@@ -2,22 +2,18 @@ use ctru::console::Console;
 use ctru::gfx::Gfx;
 use ctru::services::apt::Apt;
 use ctru::services::hid::{Hid, KeyPad};
-use ctru::services::ps::Ps;
 
 fn main() {
     // Initialize services
+    //
+    // HashMaps generate hashes thanks to the 3DS' criptografically secure generator.
+    // This generator is only active when activating the `PS` service.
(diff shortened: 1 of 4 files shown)


ctru-rs/examples/hashmaps.rs
```diff
@@ -2,23 +2,19 @@ use ctru::console::Console;
 use ctru::gfx::Gfx;
 use ctru::services::apt::Apt;
 use ctru::services::hid::{Hid, KeyPad};
-use ctru::services::ps::Ps;
 
 fn main() {
     // Initialize services
+    //
+    // HashMaps generate hashes thanks to the 3DS' criptografically secure generator.
```
AzureMarker, line 9: nit: While we're at it, "criptografically" -> "cryptographically"

AzureMarker (approved):

PR #50, opened 2022-02-26
Fixed duplicate services and constistency in naming

Meziu:
Weirdly, we missed one of the biggest memory safety issues in this whole library: there may be more than one handle to a service at once, yet services would break if more than an instance existed at a time.

For example:
```Rust
let gfx = Gfx::default();
let gfx2 = Gfx::default();

drop(gfx2);

// Now `gfx` is completely invalid and the program will crash when dropping it 
// since it'll try to double-free memory
```

Some services (like `Gfx`) completely crash the program, others simply stop working, but all are affected in an unexpected behaviour.

I wanted to keep the "own the service" behaviour of having a struct to hold them, because it fits nicely with Rust's `drop` mechanism, so the solution was to have an `AtomicBool` signaling the presence of the struct at runtime. This also comes with a new `ctru::Error` field to signal multiple instancing (we can add as many of those as we want since the `Os` field is separate). 

This also comes with a different `Gfx` default impl, since using the trait wouldn't work anymore with the possible error handling (changed `::default()` with `::init_default()` which now returns a `ctru::Result` object). At last, I've added tests in all service modules to test for these new changes. I believe we should start adding them from now on, since testing is an absolute breeze.

Overall it's a gigantic PR with many breaking changes, but it is absolutely needed something like this system to keep the whole library memory safe. Hope crashes won't be a thing anymore in the future.


AzureMarker:
Where is the double init problem documented? I haven't run across that before.

Meziu:
> Where is the double init problem documented? I haven't run across that before.

The issue isn’t with double initialisation, it’s with double ownership. As I showed in my snippet, if one of the two `Gfx` instances could get dropped before the other, but the drop code closes the service anyways, making the other struct invalid (functions would either do nothing or crash, depending on the service). We also can’t solve it with an `Rc` since that would undo the principle of mutability of the service’s state (Eg. keeping track of the current setup for the `Gfx` screens or `RomFS` mount).

Meziu:
> Weirdly, we missed one of the biggest memory safety issues in this whole library: there may only be more than one handle to a service at once, yet services would break if more than an instance existed at a time.
> 
> For example:
> 
> ```rust
> let gfx = Gfx::default();
> let gfx2 = Gfx::default();
> 
> drop(gfx2);
> 
> // Now `gfx` is completely invalid and the program will crash when dropping it 
> // since it'll try to double-free memory
> ```
> 
> Some services (like `Gfx`) completely crash the program, others simply stop working, but all are affected in an unexpected behaviour.
> 
> I wanted to keep the "own the service" behaviour of having a struct to hold them, because it fits nicely with Rust's `drop` mechanism, so the solution was to have an `AtomicBool` signaling the presence of the struct at runtime. This also comes with a new `ctru::Error` field to signal multiple instancing (we can add as many of those as we want since the `Os` field is separate).
> 
> This also comes with a different `Gfx` default impl, since using the trait wouldn't work anymore with the possible error handling (changed `::default()` with `::init_default()` which now returns a `ctru::Result` object). At last, I've added tests in all service modules to test for these new changes. I believe we should start adding them from now on, since testing is an absolute breeze.
> 
> Overall it's a gigantic PR with many breaking changes, but it is absolutely needed something like this system to keep the whole library memory safe. Hope crashes won't be a thing anymore in the future.

ctru-rs/src/lib.rs
```diff
@@ -38,12 +38,31 @@ pub fn init() {
...
+            // so an `Hid` stuct may still be alive and thus make the `panic_hook` panic.
+            // If that were to happen, the system would have to reboot to properly close the app.
+            // 
+            // Using `hidInit` is safe when another instance is open, and we can do safe operations afterwards.
+            // The only (probably) unsafe part of this system is the `hidExit`, since in a multithreaded
+            // environment some other threads may still be doing operations on the service
+            // before the cleanup, though the time window would be almost nonexistent, and it would only
+            // really be a problem in preemptive threads.
+            //
+            // TL;DR : This code is bad.
```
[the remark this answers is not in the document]
AzureMarker: Maybe we need to do something a little more fancy than an atomic boolean.

One idea is having a static `Mutex<usize>` which starts at zero and counts the number of open service handles. When it reaches zero again in drop, we can call the exit function. It needs to be a Mutex and not just an AtomicUsize because we need to hold the lock while we call the `hidExit` function, for example. Otherwise, we might decrement to zero, another thread calls `Hid::init` and increments to one (calling `hidInit`), and the drop function finally calls `hidExit`.

This would let us allow multiple handles to be created without issue, and only exiting the service once all are gone. Basically re-implementing reference counting, but it's a little hard to reuse `Arc` for this (feel free to try if you want, but it will probably be more verbose and complex).

We could code this pattern up into a `ServiceCleaner` or something (name tbd), like so (untested):
```rust
struct ServiceCleaner {
    counter: &'static Mutex<usize>
    exit: extern "C" fn()
}

impl ServiceCleaner {
    fn init(counter: &'static Mutex<usize>, exit: extern "C" fn()) -> Self {
        let mut value = counter.lock().unwrap() // todo: handle poisoning
        *value += 1;
        drop(value);
        Self { counter, exit }
    }
}

impl Drop for ServiceCleaner {
    fn drop(&mut self) {
        let mut value = self.counter.lock().unwrap(); // should probably handle poisoning - could just map_err to ignore it.
        *value -= 1;
        if *value == 0 {
            unsafe { self.exit() };
        }
    }
}

```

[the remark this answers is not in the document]
AzureMarker: One place to be careful about with the above is `Gfx`. We should probably still only let them have one instance of that struct.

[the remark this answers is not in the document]
Meziu: `Gfx` absolutely must not be re-initialised, since it takes parameters that change the internal state. Other services, like `Hid` or `SslC` have no problem being shared across threads, though we should be careful with what the services do.

ctru-rs/src/services/sslc.rs
```diff
@@ -1,35 +1,58 @@
...
-            } else {
-                Ok(())
-            }
+        let r = unsafe { ctru_sys::sslcGenerateRandomData(buf.as_ptr() as _, buf.len() as u32) };
+        if r < 0 {
+            Err(r.into())
+        } else {
+            Ok(())
         }
     }
```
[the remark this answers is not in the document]
Meziu: We talked about it when implementing `getrandom` in `linker-fix-3ds`. `SslC` looks to be just an overhead of `PS`, so we decided to use the “lower level” one.

ctru-rs/src/lib.rs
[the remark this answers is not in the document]
AzureMarker: Regarding the function naming, `init` is probably fine so long as the underlying ctru_sys funciton (ex. `aptInit`) is OK if it's called multiple times.

I checked all of the "*Init()" functions in libctru, and they all\* actually handle init/exit with reference counting correctly. So we could don't need to do anything special. Even the `gfxInit` function is partially guarded (the `gspInit` call only inits once, but the rest of the gfx settings get reset on every call).

\* romfsInit does try to mount it every time, so it might error out. And we still may still want to limit the Gfx initialization.

ctru-rs/src/error.rs
```diff
@@ -28,6 +29,7 @@ impl fmt::Debug for Error {
                 .field("summary", &R_SUMMARY(err))
                 .field("level", &R_LEVEL(err))
                 .finish(),
+            Error::ServiceAlreadyActive(service) => write!(f, "Service {service} already active"),
```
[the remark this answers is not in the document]
ian-h-chamberlain: We could use enum variants instead of the inner `&'static str` too if we wanted, since there is a finite well-known set of services. Might be annoying to keep in sync as we add new modules, though?

ian-h-chamberlain:
As an alternative design, I wonder if we could consider something like the singleton pattern laid out [in the embedded book](https://docs.rust-embedded.org/book/peripherals/singletons.html#how-do-we-do-this-in-rust).

I think it would look basically like
```rs
#[derive(Default)]
struct Services {
	gfx: Option<Gfx>,
	hid: Option<Hid>,
	//...
}

static mut SERVICES: Services = Services::default();
```

Then a user could either request a given service (`Gfx::try_init` or something maybe?), which could initialize and either create or replace the global `Option<Gfx>`. This would perhaps require APIs like `Gfx::force_init` or something if the user wanted to replace the existing service, but it seems like that might be necessary to support such a use case anyway.

There would probably still need to be some kind of lock on the singleton, for thread safety, but since it's `static` it should be safe to hand out references to the underlying service handles (`&'static Gfx` - is `&mut` okay here too?). We might want to avoid the `take_serial()` example they provide, since we wouldn't want to allow a user to call `take_gfx()` followed by `gfx_init` or something and end up with two "live" `Gfx`... but I think in principle this could work to ensure services are only initialized one at a time.

AzureMarker:
That singleton pattern is interesting (I haven't seen it before), but since libctru already reference counts most of the services, I think we should keep that pattern.

Meziu:
Fixed the nits and such. I believe the `ServiceCloser` idea would be better, but I won't be able to thoroughly test it for the time being.

Meziu:
Oh well, trying to test things out made me realize an issue with the `ServiceCloser` idea: `Mutex`es aren't `const`.
The idea of keeping the lock is good, but it's a problem to implement...

AzureMarker:
We can use the Mutex from the `spin` crate like we do with `RwLock` in pthread-3ds.

ctru-rs/src/gfx.rs
```diff
@@ -227,6 +227,9 @@ mod tests {
     #[test]
     fn gfx_duplicate() {
         // We don't need to build a `Gfx` because the test runner has one already
-        assert!(Gfx::init_default().is_err());
+        match Gfx::init() {
+            Err(Error::ServiceAlreadyActive("Gfx")) => return,
+            _ => panic!(),
+        }
```
AzureMarker, line 233: nit: this could be:
```suggestion
        assert!(matches!(Gfx::init(), Err(Error::ServiceAlreadyActive("Gfx"))));
```

ctru-rs/src/error.rs
```diff
@@ -29,7 +29,10 @@ impl fmt::Debug for Error {
                 .field("summary", &R_SUMMARY(err))
                 .field("level", &R_LEVEL(err))
                 .finish(),
-            Error::ServiceAlreadyActive(service) => write!(f, "Service {service} already active"),
+            Error::ServiceAlreadyActive(service) => f
+                .debug_tuple("ServiceAlreadyActive")
+                .field(&String::from(service))
```
AzureMarker, line 34: nit: this could just be the following?
```suggestion
                .field(service)
```

AzureMarker:
We may also be able to avoid any reference counting on our end for most services based on this:
https://github.com/Meziu/ctru-rs/pull/50#discussion_r817347456

Meziu:
> We can use the Mutex from the `spin` crate like we do with `RwLock` in pthread-3ds (yielding instead of busy-looping).

I went with `SyncLazy`. I believe it's better because it only initializes when needed, and not in all programs.

Meziu:
Ok, I changed most of everything. Now it uses `Mutex`, and all services have an abstract implementation of `ServiceHandler` (did it this way because the `init` functions had to be in the lock's lifetime too, and also because it was easy to expand). This is nowhere near final.

@AzureMarker there are many things you'll have to review, though one of the smallest is the presence of `Error::ServiceAlreadyActive`. With this implementation, it's used only by `Gfx`, so I removed the `&str` fields altogether (they were also really ugly).

ctru-rs/src/services/mod.rs
```diff
@@ -9,3 +9,48 @@ pub mod sslc;
 pub use self::apt::Apt;
 pub use self::hid::Hid;
 pub use self::sslc::SslC;
+
+use crate::Error;
+use std::sync::Mutex;
+pub(crate) struct ServiceHandler {
```
AzureMarker, line 15: nit: this should go in a separate file.

ctru-rs/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 #![feature(test)]
 #![feature(custom_test_frameworks)]
 #![test_runner(test_runner::run)]
+#![feature(once_cell)]
```
AzureMarker, line 6: nit: we can use the once_cell crate instead to avoid adding another feature here.

ctru-rs/src/lib.rs
```diff
@@ -28,6 +29,7 @@ pub fn init() {
 
     use std::panic::PanicInfo;
 
+    #[cfg(not(test))]
```
AzureMarker, line 32: nit: this panic hook setup could be in a separate function so we only need one `cfg(not(test))` and it doesn't get interwoven in `init`.

ctru-rs/src/romfs.rs
```diff
@@ -54,10 +54,14 @@ mod tests {
     #[test]
     fn romfs_duplicate() {
         let _romfs = RomFS::init().unwrap();
+        let lock = *ROMFS_ACTIVE.lock().unwrap();
+
+        assert_eq!(lock, 1);
+
+        drop(_romfs);
```
AzureMarker, line 61: Does this not cause a deadlock? Wouldn't the first lock not get dropped until the end of scope?

ctru-rs/src/services/apt.rs
```diff
@@ -1,25 +1,38 @@
...
+use std::lazy::SyncLazy;
+use std::sync::Mutex;
 
-use crate::error::Error;
+use crate::services::ServiceHandler;
 
 #[non_exhaustive]
-pub struct Apt(());
+pub struct Apt {
+    _service_handler: ServiceHandler,
```
AzureMarker, line 8: Most of these services don't need to use `ServiceHandler` because they are already reference counted by libctru:
https://github.com/devkitPro/libctru/blob/ebb53051886cfe966f1aa27ff3dd6689176de483/libctru/source/services/apt.c#L195
https://github.com/devkitPro/libctru/blob/ebb53051886cfe966f1aa27ff3dd6689176de483/libctru/source/services/apt.c#L345

ctru-rs/src/services/apt.rs
```diff
@@ -1,25 +1,38 @@
...
-        }
+
+                Ok(())
+            },
+            // `socExit` returns an error code. There is no documentantion of when errors could happen,
+            // but we wouldn't be able to handle them in the `Drop` implementation anyways.
+            // Surely nothing bad will happens :D
+            || unsafe {
+                // The socket buffer is freed automatically by `socExit`
+                ctru_sys::aptExit();
```
AzureMarker, line 31: The SOC comments got copied around too much and aren't relevant here.

ctru-rs/src/romfs.rs
```diff
@@ -11,39 +11,39 @@
...
-static ROMFS_ACTIVE: AtomicBool = AtomicBool::new(false);
+static ROMFS_ACTIVE: SyncLazy<Mutex<usize>> = SyncLazy::new(|| Mutex::new(0));
 
 impl RomFS {
     pub fn init() -> crate::Result<Self> {
-        match ROMFS_ACTIVE.compare_exchange(false, true, Ordering::SeqCst, Ordering::SeqCst) {
-            Ok(_) => {
+        let _service_handler = ServiceHandler::new(
+            &ROMFS_ACTIVE,
+            true,
```
AzureMarker, line 30: We should probably set this to false.

ctru-rs/src/services/mod.rs
AzureMarker, line 15: nit: Using `Handler` in a type name usually means there is a better name that can be used instead. `ServiceCloser` or `ServiceReference` are some ideas.

ctru-rs/src/services/soc.rs
```diff
@@ -1,18 +1,18 @@
...
 
-use crate::error::Error;
-use ctru_sys::{socExit, socInit};
+use crate::services::ServiceHandler;
 
 /// Soc service. Initializing this service will enable the use of network sockets and utilities
 /// such as those found in `std::net`. The service will be closed when this struct is is dropped.
 #[non_exhaustive]
 pub struct Soc {
-    soc_mem: *mut u32,
```
AzureMarker, line 8: We should still deallocate the memory. To store the memory address, you might need another Mutex.

ctru-rs/src/services/soc.rs
```diff
@@ -31,20 +31,28 @@ impl Soc {
...
-                } else {
-                    Ok(Self { soc_mem })
+                    return Err(r.into());
                 }
+
+                Ok(())
+            },
+            // `socExit` returns an error code. There is no documentantion of when errors could happen,
+            // but we wouldn't be able to handle them in the `Drop` implementation anyways.
+            // Surely nothing bad will happens :D
```
AzureMarker, line 48: We could panic in the drop, but it seems a little over the top.

ctru-rs/src/services/soc.rs
```diff
@@ -31,20 +31,28 @@ impl Soc {
...
+            || unsafe {
+                // The socket buffer is freed automatically by `socExit`
+                ctru_sys::socExit();
```
AzureMarker, line 51: Is this true? I couldn't find the place it's freed in libctru. Also, does that mean that the previous code was doing a double free?

ctru-rs/src/romfs.rs
Meziu: Why? The start code isn’t run again.

ctru-rs/src/services/apt.rs
Meziu: Ooops, yeah…

ctru-rs/src/services/apt.rs
Meziu: But there isn’t an automatic drop, the decrement is still only available in the `aptExit`.

ctru-rs/src/romfs.rs
Meziu: Yeah I named the variable wrongly. It doesn’t hold the lock (eve  though it’s called `lock`), it just holds the value by using a temporary lock.

ctru-rs/src/services/soc.rs
Meziu: I’d avoid panicking. Also, if the code works properly (and thus the service is actually open when calling the `Drop`) then I believe there aren’t any problems.

ctru-rs/src/services/soc.rs
Meziu: While searching around for this buffer, I saw this note in the docs:

> Note: The specified context buffer can no longer be accessed by the process which called this function, since the userland permissions for this block are set to no-access.

Asking in the homebrew forums for confirmation, Fenrir himself responded telling me a free is probably the case, possibly handled by the OS itself when the process is closed. A double free panic didn’t happen possibly because the functions (`free` has no way of error handling) simply ignore OS errors.

ctru-rs/src/services/mod.rs
Meziu: `ServiceCloser` is misleading, since it also starts the service.

ctru-rs/src/services/apt.rs
AzureMarker: We can call the exit functions in our drop like the code in master does. The service will only actually exit once its internal reference count is zero.

ctru-rs/src/romfs.rs
AzureMarker: That's fair. I guess if we add `&mut self` functions we should switch it to false, but it's fine for now.

ctru-rs/src/services/soc.rs
AzureMarker: nvm, turns out we lose access to this memory when we pass it in.

ctru-rs/src/lib.rs
```diff
@@ -27,9 +26,13 @@ pub fn init() {
         libc::atexit(services_deinit);
     }
 
+    #[cfg(test)]
+    panic_hook_setup();
```
AzureMarker, line 30: This should be `cfg(not(test))`

ctru-rs/src/services/mod.rs
```diff
@@ -5,52 +5,10 @@ pub mod hid;
 pub mod ps;
 pub mod soc;
 pub mod sslc;
+mod reference;
```
AzureMarker, line 8: I think you forgot to commit `reference.rs`

Meziu:
I reverted all services not needing the refcount (`Soc` doesn't have ref-count by itself, nor `RomFS`).

ctru-rs/src/lib.rs
```diff
@@ -26,6 +26,12 @@ pub fn init() {
         libc::atexit(services_deinit);
     }
 
+    #[cfg(not(test))]
+    _panic_hook_setup();
+}
+
+fn _panic_hook_setup() {
```
AzureMarker, line 33: nit: This shouldn't be prefixed with `_`. It's already a private function because it doesn't have `pub`.

ctru-rs/src/lib.rs
```diff
@@ -38,11 +44,13 @@ pub fn init() {
         // Only for panics in the main thread
         if main_thread == thread::current().id() && console::Console::exists() {
             println!("\nPress SELECT to exit the software");
-            let hid = services::hid::Hid::init().unwrap();
+
+            let hid = Hid::init().unwrap();
```
AzureMarker, line 48: A little out of scope for this PR, but maybe we should not unwrap here? Instead if there's an error, just return?

I don't have much opinion either way, but it stood out to me.

ctru-rs/src/services/reference.rs
```diff
@@ -0,0 +1,44 @@
...
+        counter: &'static Mutex<usize>,
+        allow_multiple: bool,
+        start: S,
+        close: E,
+    ) -> crate::Result<Self>
+    where
+        S: FnOnce() -> crate::Result<()>,
+        E: Fn() + 'static,
+    {
+        let mut value = counter.lock().unwrap(); // todo: handle poisoning
```
AzureMarker, line 19: Do you want to handle the poisoning TODOs in this PR?

ctru-rs/src/services/reference.rs
```diff
@@ -0,0 +1,44 @@
+use crate::Error;
+use std::sync::Mutex;
+pub(crate) struct ServiceReference {
+    counter: &'static Mutex<usize>,
+    close: Box<dyn Fn()>,
```
AzureMarker, line 5: I wonder if we should make this `dyn Fn() + Send + Sync`. I think this PR as-is would remove the `Send` and `Sync` auto-impls for the services that hold this struct.

ctru-rs/src/romfs.rs
```diff
@@ -10,27 +10,58 @@
...
-impl Drop for RomFS {
-    fn drop(&mut self) {
-        let mount_name = CStr::from_bytes_with_nul(b"romfs\0").unwrap();
-        unsafe { ctru_sys::romfsUnmount(mount_name.as_ptr()) };
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn romfs_duplicate() {
```
AzureMarker, line 55: nit: I think some of these test names are a little inaccurate. This one isn't testing duplicates, for instance.

AzureMarker (commented):
Just a few nits and questions. The PR looks good overall.

ctru-rs/src/lib.rs
Meziu: Yeah, you’re right. Maybe I could print an error if the user is catching `stdout`, but panicking is a bad idea.

ctru-rs/src/services/reference.rs
Meziu: I’d like to. What can we do about it, though?

AzureMarker: We can either ignore it, unwrap/panic, or handle it. In our case, what it means to be poisoned is that the lock may have been held during a panic, which dropped the lock. So, maybe the start function panics, for example. Or, start doesn't panic, but somehow there's a panic between start and the counter increment.

If we ignore it, that is probably fine, but we just need to make sure we can't panic between the start task succeeding and the counter getting incremented. This is probably fine. Unwrapping/panicking is also an option, and would be the safest, plus it's unlikely that the lock gets poisoned in the first place, so this is fine to do. Handling the poison error seems hard because that implies fixing the broken state. But it's hard to know what's broken. So not much to do here I think.

I may be missing a few things since this is all off the top of my head.

Meziu: I guess I'll change from an `unwrap` to an expect with a bit of info, but other than that it's hard. Still, panicking the thread while holding the lock is pretty much impossible, so I don't think there are issues.

Meziu:
Changed `Soc`'s `allow_multiple` to false, because the init is handled with a parameter and we can't return an `Ok` without telling the user the change won't be made.

PR #57, opened 2022-04-13
Write a console that does not flush gfx buffers

ian-h-chamberlain:
@Meziu @AzureMarker 

The default libctru console calls `gfxFlushBuffers` whenever it prints a `\n` or `\r` character, which is normally useful since its output will be visible immediately.

However, I have a use case where I'd like to print concurrently from multiple threads, and manage flushing of the graphics buffer myself at the end of the frame. I noticed that the `gfxFlushBuffers` seemed to cause a deadlock when run concurrently from multiple threads – perhaps a bug in the underlying implementation?

In any case, this PR hacks together a variation on `Console` that works basically the same as the default one, but does not flush the buffers. Even without fully understanding the deadlock issue I mentioned, I think this could be useful as most programs should probably only flush once per frame anyway: https://libctru.devkitpro.org/gfx_8h.html#aea1808bd74fe0c00f9794e455fc8499b
I'm hoping this will make the console usable as a debug tool on one screen while rendering / doing other stuff with the other one.

The implementation does feel a bit hacky, though, so if you have suggestions or think we shouldn't merge this, please let me know. I thought this seemed a bit easier than writing a full-blown console from scratch, anyway.

### DIFF
--- a/ctru-rs/src/console.rs
+++ b/ctru-rs/src/console.rs
@@ -5,6 +5,8 @@ use ctru_sys::{consoleClear, consoleInit, consoleSelect, consoleSetWindow, Print
 
 use crate::gfx::Screen;
 
+mod non_flushing;
+
 static mut EMPTY_CONSOLE: PrintConsole = unsafe { const_zero::const_zero!(PrintConsole) };
 
 pub struct Console<'screen> {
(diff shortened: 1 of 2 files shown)


Meziu:
I would avoid this approach. What we could do instead is actually re-write the console, as it looks like a buggy mess of a thing and doesn’t support functionality important to our environment (comes to mind the UTF-8 support missing). I wonder how hard it would be, though it’d also make switching Consoles and such much easier and safe via Rust.

AzureMarker:
In terms of rewriting Console, we should at least keep the libctru console available and have our own version in addition to it. Or, make it a separate library, since this crate is supposed to just be a nice wrapper around libctru.

Meziu:
> In terms of rewriting Console, we should at least keep the libctru console available and have our own version in addition to it. Or, make it a separate library, since this crate is supposed to just be a nice wrapper around libctru.

It depends on how good of a job we do 😜.
Still, I haven’t yet looked at how the console renders the text, is it CPU or GPU based?

Meziu:
Yeah `citro2d` has built-in GPU font-rendering, so I guess that would be the best choice. It is still weird that such a useful part of `libctru` isn't recommended, but aight.

ian-h-chamberlain:
Guess I'll close this since I have started work on a completely custom console instead, and this is kind of a hack anyway.

PR #59, opened 2022-05-03
Increase default stack size to 2MB with opt-out available

ian-h-chamberlain:
Fix / workaround for issue seen in #56 

Add a Cargo feature to allow users to set the stack size if desired, but
default to 2MB to handle more use cases.

@Meziu @AzureMarker what do you think of this approach? Open to alternate naming suggestions on the cargo feature as well, I don't love `big-stack` but wanted to post this as a starting point for discussion.

Another approach I considered was exporting a macro like 
```rs
ctru::stack_size!(64 * 1024);
```
But it felt a bit clunky and would require _every_ user to set this if their program had large enough stacks. I would imagine that custom stack size is a pretty advanced use case (assuming we pick a good default) so I think it should still be possible for users to opt-out of the default, but I'm not sure of another way besides a Cargo feature.

### DIFF
--- a/ctru-rs/Cargo.toml
+++ b/ctru-rs/Cargo.toml
@@ -32,8 +32,9 @@ tokio = { version = "1.16", features = ["rt", "time", "sync", "macros"] }
 cfg-if = "1.0.0"
 
 [features]
-default = ["romfs"]
+default = ["romfs", "big-stack"]
 romfs = []
+big-stack = []
 
 [package.metadata.cargo-3ds]
 romfs_dir = "examples/romfs"
--- a/ctru-rs/src/lib.rs
+++ b/ctru-rs/src/lib.rs
@@ -10,6 +10,10 @@ extern "C" fn services_deinit() {
     }
 }
 
+#[no_mangle]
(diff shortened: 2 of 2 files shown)


Meziu (approved):

AzureMarker (approved):
Thanks for fixing this!!

AzureMarker:
The feature name seems good enough to me. We could in the future have both the feature and a macro to set the stack size.

ian-h-chamberlain:
Another option I considered was something like `custom-stack-size` and invert the semantics – i.e. configure this size when `#[cfg(not(feature = "custom-stack-size"))]`. But the guidance from documentation kidn of seems to discourage this (https://doc.rust-lang.org/cargo/reference/features.html#feature-unification):


> A consequence of this is that features should be additive. That is, enabling a feature should not disable functionality, and it should usually be safe to enable any combination of features. A feature should not introduce a [SemVer-incompatible change](https://doc.rust-lang.org/cargo/reference/features.html#semver-compatibility).

So I guess this can be good enough for now.

PR #61, opened 2022-05-07
Add 3dslink redirection to Soc service

ian-h-chamberlain:
@Meziu @AzureMarker 

This allows you to send stdout/stderr back to the `3dslink` invocation, which will probably be really nice for graphical applications where you need both screens, etc. We might also consider putting it in the test runner (maybe as a feature flag or CLI arg?)

I'm not 100% sure about this API, maybe it would make sense to have separate `redirect_stdout()` and `redirect_stderr()` functions instead of just the one? I also wasn't sure it should be on `Soc`, but you have to initialize `Soc` before you can use this so it seemed like a somewhat reasonable place to have it.

Next up, I would probably like to add `cargo-3ds` support for this so you could do `cargo 3ds run --server` or something like that, instead of manually invoking it like I have done here.

----

Also, I fixed a broken thread-info compilation which was a small change.

### DIFF
--- a/ctru-rs/src/services/soc.rs
+++ b/ctru-rs/src/services/soc.rs
@@ -10,6 +10,7 @@ use crate::services::ServiceReference;
 #[non_exhaustive]
 pub struct Soc {
     _service_handler: ServiceReference,
+    sock_3dslink: libc::c_int,
 }
 
 static SOC_ACTIVE: Lazy<Mutex<usize>> = Lazy::new(|| Mutex::new(0));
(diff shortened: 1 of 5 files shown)


AzureMarker (commented):
Very cool!

ctru-rs/src/services/soc.rs
```diff
@@ -69,7 +105,7 @@ mod tests {
 
     #[test]
     fn soc_duplicate() {
-        let _soc = Soc::init().unwrap();
+        // let _soc = Soc::init().unwrap();
```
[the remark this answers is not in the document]
ian-h-chamberlain: Oops, yep, this is left over from my preemptive testing of using this feature for the test runner. Will revert it

ctru-rs/src/services/soc.rs
```diff
@@ -52,14 +53,49 @@ impl Soc {
...
+    /// Requires `3dslink` >= 0.6.1 and `new-hbmenu` >= 2.3.0.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if a connection cannot be established to the server, or
+    /// output was already previously redirected.
+    pub fn redirect_to_3dslink(&mut self, stdout: bool, stderr: bool) -> crate::Result<()> {
+        if self.sock_3dslink >= 0 {
+            // TODO AlreadyRedirected or something
+            return Err(crate::Error::ServiceAlreadyActive);
```
ian-h-chamberlain, line 78: oops, forgot to address this as well. Should this be a new enum variant for `crate::Error`, or maybe this function should not return `crate::Result` at all?

AzureMarker: Either re-redirect it (close the socket and reopen it?) Or return an error with a new variant. It'll probably depend on if libctru supports calling it twice.

ian-h-chamberlain: After some testing, it seems like either the client or the server implementation doesn't support more than one call. So I think I will just add a new enum variant to be safe.

ctru-sys/src/lib.rs
```diff
@@ -10,3 +10,10 @@ mod bindings;
 
 pub use bindings::*;
 pub use result::*;
+
+/// In lieu of a proper errno function exposed by libc
+/// (<https://github.com/rust-lang/libc/issues/1995>), this will retrieve the
```
AzureMarker, line 15: Maybe we should open a PR to libc for this as well.

AzureMarker (approved):

PR #63, opened 2022-07-10
Initial CFGU implementation

Techie-Pi:
Implementation of the [Configuration service](https://libctru.devkitpro.org/cfgu_8h.html). Only some region, model and language-related functions have been implemented.

<details>
<summary>Missing functions</summary>

- CFGU_GenHashConsoleUnique
- CFGU_GetRegionCanadaUSA
- CFGU_GetCountryCodeString
- CFGU_GetCountryCodeID
- CFGU_GetConfigInfoBlk2
- CFGU_GetConfigInfoBlk4
- CFGU_GetConfigInfoBlk8
- CFGU_SetConfigInfoBlk4
- CFGU_SetConfigInfoBlk8
- CFG_UpdateConfigSavegame
- CFGI_RestoreLocalFriendCodeSeed
- CFGI_RestoreSecureInfo
- CFGI_DeleteConfigSavefile
- CFGI_FormatConfig
- CFGI_ClearParentalControls
- CFGI_VerifySigLocalFriendCodeSeed
- CFGI_VerifySigSecureInfo
- CFGI_SecureInfoGetSerialNumber
- CFGI_GetLocalFriendCodeSeedData
- CFGI_GetLocalFriendCodeSeed
- CFGI_GetSecureInfoData
- CFGI_GetSecureInfoSignature

</details>


ctru-rs/src/services/cfgu.rs
```diff
@@ -0,0 +1,227 @@
+//! Configuration service
+//!
+//! This module contains basic methods to retrieve and change configuration from the console.
+
+#[derive(Debug)]
+#[repr(u8)]
+pub enum Region {
```
Meziu, line 7: You should reference the `ctru_sys` enums rather than hard-coding numbers. (Here and in the conversion methods).

ctru-rs/src/services/cfgu.rs
```diff
@@ -0,0 +1,227 @@
...
+            if r < 0 {
+                Err(r.into())
+            } else {
+                Ok(supported)
+            }
+        }
+    }
+
+    /// Check if the console is 2DS-like (2DS, New2DS, New2DSXL)
+    pub fn is_2ds_like(&self) -> crate::Result<bool> {
```
Meziu, line 137: `is_2ds` is a more clear name. `is_2ds_family` could be another option, but I’d like to hear @AzureMarker’s thoughts.

Meziu (changes requested):

ctru-rs/src/services/cfgu.rs
AzureMarker: `is_2ds_family` is a good option, otherwise `is_2ds`.

ctru-rs/src/services/cfgu.rs
Techie-Pi: (I think) this is resolved now

ctru-rs/src/services/cfgu.rs
Techie-Pi: In my opinion, ``is_2ds_family`` is the best option, because ``is_2ds`` could be somewhat ambiguous. I'll wait until a consensus is reached to change it tho

Meziu (approved):
LGTM, @AzureMarker what do you think?

Meziu:
> * Should I also derive `Clone` on the enums?
They have `Copy` by default, there is no reason to explicitly add `Clone` (that I know of)

> * Should I change the filename of the example to something like `configuration.rs`?
Maybe `system-configuration.rs` would be a better fit.

Techie-Pi:
> They have Copy by default, there is no reason to explicitly add Clone (that I know of)
Oh that's great

> Maybe system-configuration.rs would be a better fit.
Alright, I'm going to change it

AzureMarker:
> > * Should I also derive `Clone` on the enums?
> 
> They have `Copy` by default, there is no reason to explicitly add `Clone` (that I know of)

Enums don't have `Copy` by default, and `Copy` requires `Clone`, so yes please add a derive for both :).

ctru-rs/src/services/cfgu.rs
```diff
@@ -0,0 +1,227 @@
+//! Configuration service
+//!
+//! This module contains basic methods to retrieve and change configuration from the console.
+
+#[derive(Debug)]
+#[repr(u32)]
+pub enum Region {
+    Japan       = ctru_sys::CFG_REGION_JPN,
+    Usa         = ctru_sys::CFG_REGION_USA,
```
AzureMarker, line 9: Capitalize this to be `USA`

ctru-rs/src/services/cfgu.rs
```diff
@@ -0,0 +1,227 @@
...
+#[derive(Debug)]
+#[repr(u32)]
+pub enum Language {
+    Japan       = ctru_sys::CFG_LANGUAGE_JP,
+    English     = ctru_sys::CFG_LANGUAGE_EN,
+    French      = ctru_sys::CFG_LANGUAGE_FR,
+    German      = ctru_sys::CFG_LANGUAGE_DE,
+    Italian     = ctru_sys::CFG_LANGUAGE_IT,
+    Spanish     = ctru_sys::CFG_LANGUAGE_ES,
+    SimpChinese = ctru_sys::CFG_LANGUAGE_ZH,
```
AzureMarker, line 26: Rename to `SimplifiedChinese`

ctru-rs/src/services/cfgu.rs
```diff
@@ -0,0 +1,227 @@
...
+    Korean      = ctru_sys::CFG_LANGUAGE_KO,
+    Dutch       = ctru_sys::CFG_LANGUAGE_NL,
+    Portuguese  = ctru_sys::CFG_LANGUAGE_PT,
+    Russian     = ctru_sys::CFG_LANGUAGE_RU,
+    TradChinese = ctru_sys::CFG_LANGUAGE_TW,
```
AzureMarker, line 31: Rename to `TraditionalChinese`

ctru-rs/src/services/cfgu.rs
```diff
@@ -0,0 +1,227 @@
...
+        }
+    }
+
+    /// Gets system region from secure info
+    pub fn get_region(&self) -> crate::Result<Region> {
+        let mut region: u8 = 0;
+        let region_pointer: *mut u8 = &mut region;
+
+        unsafe {
+            let r = ctru_sys::CFGU_SecureInfoGetRegion(region_pointer);
```
AzureMarker, line 79: nit: This works too, and is a little less complicated:
```suggestion
        let mut region: u8 = 0;

        unsafe {
            let r = ctru_sys::CFGU_SecureInfoGetRegion(&mut region);
```

ctru-rs/src/services/cfgu.rs
```diff
@@ -0,0 +1,227 @@
...
+            }
+        }
+    }
+
+    /// Gets system region from secure info
+    pub fn get_region(&self) -> crate::Result<Region> {
+        let mut region: u8 = 0;
+        let region_pointer: *mut u8 = &mut region;
+
+        unsafe {
```
AzureMarker, line 78: You only need the unsafe block around the libctru function call. You can do something like this:
```rust
let r = unsafe { ctru_sys::CFGU_SecureInfoGetRegion(region_pointer) };
```

ctru-rs/src/services/cfgu.rs
```diff
@@ -0,0 +1,227 @@
...
+    pub fn is_2ds_family(&self) -> crate::Result<bool> {
+        let mut is_2ds_family: u8 = 0;
+        let is_2ds_family_pointer: *mut u8 = &mut is_2ds_family;
+
+        unsafe {
+            let r = ctru_sys::CFGU_GetModelNintendo2DS(is_2ds_family_pointer);
+            if r < 0 {
+                Err(r.into())
+            } else {
+                Ok(is_2ds_family < 1)
```
AzureMarker, line 146: The documentation is explicit that 0 equals yes, so change this to the following:
```suggestion
                Ok(is_2ds_family == 0)
```

ctru-rs/src/services/cfgu.rs
```diff
@@ -0,0 +1,227 @@
...
+            ctru_sys::CFG_LANGUAGE_EN  => Ok(Language::English),
+            ctru_sys::CFG_LANGUAGE_FR  => Ok(Language::French),
+            ctru_sys::CFG_LANGUAGE_DE  => Ok(Language::German),
+            ctru_sys::CFG_LANGUAGE_IT  => Ok(Language::Italian),
+            ctru_sys::CFG_LANGUAGE_ES  => Ok(Language::Spanish),
+            ctru_sys::CFG_LANGUAGE_ZH  => Ok(Language::SimpChinese),
+            ctru_sys::CFG_LANGUAGE_KO  => Ok(Language::Korean),
+            ctru_sys::CFG_LANGUAGE_NL  => Ok(Language::Dutch),
+            ctru_sys::CFG_LANGUAGE_PT  => Ok(Language::Portuguese),
+            ctru_sys::CFG_LANGUAGE_RU => Ok(Language::Russian),
```
AzureMarker, line 206: Run rustfmt on this file, it brings in some changes.

AzureMarker (changes requested):

ctru-rs/src/services/cfgu.rs
```diff
@@ -2,44 +2,44 @@
 //!
 //! This module contains basic methods to retrieve and change configuration from the console.
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
AzureMarker, line 5: Add `Copy` to the list of derives for all of these enums.

ctru-rs/src/services/cfgu.rs
```diff
@@ -0,0 +1,227 @@
...
+    Australia   = ctru_sys::CFG_REGION_AUS,
+    China       = ctru_sys::CFG_REGION_CHN,
+    Korea       = ctru_sys::CFG_REGION_KOR,
+    Taiwan      = ctru_sys::CFG_REGION_TWN,
+}
+
+#[derive(Debug)]
+#[repr(u32)]
+pub enum Language {
+    Japan       = ctru_sys::CFG_LANGUAGE_JP,
```
ian-h-chamberlain: nit: Japanese

ctru-rs/src/services/cfgu.rs
```diff
@@ -0,0 +1,227 @@
...
+    Dutch       = ctru_sys::CFG_LANGUAGE_NL,
+    Portuguese  = ctru_sys::CFG_LANGUAGE_PT,
+    Russian     = ctru_sys::CFG_LANGUAGE_RU,
+    TradChinese = ctru_sys::CFG_LANGUAGE_TW,
+}
+
+#[derive(Debug)]
+#[repr(u32)]
+pub enum SystemModel {
+    Model3DS        = ctru_sys::CFG_MODEL_3DS,
```
ian-h-chamberlain: nit: I'd omit the `Model` prefix since usage would look like `SystemModel::3DS`

ian-h-chamberlain (approved):
Thanks for the PR! LGTM for the most part

ctru-rs/src/services/cfgu.rs
ian-h-chamberlain: Oh right, numerals can't be the first char in an identifier. Oh well

ctru-rs/src/services/cfgu.rs
```diff
@@ -0,0 +1,212 @@
...
+    /// Initializes the CFGU service.
+    ///
+    /// # Errors
+    ///
+    /// This function will return Err if there was an error initializing the
+    /// CFGU service.
+    ///
+    /// ctrulib services are reference counted, so this function may be called
+    /// as many times as desired and the service will not exit until all
+    /// instances of Fs drop out of scope.
```
AzureMarker, line 61: Replace "Fs" with "Cfgu"

AzureMarker (commented):
LGTM, just one comment.

ctru-rs/src/services/cfgu.rs
ian-h-chamberlain: Also, if you are referencing a real item that's in scope, you can write ``[`Cfgu`]`` for generated docs to make a clickable link to the actual `Cfgu` struct, which isn't super relevant here since it's already under the impl, but it looks nice.

Techie-Pi: > Replace "Fs" with "Cfgu"

Oops, i used the Fs documentation as a reference an looks like I copy-pasted that and forgot to change it :p

Techie-Pi:
Also, this is probably the last change I'm going to be able to make in at least a month. I'm going to have surgery next week, and I won't be able to use the computer. (just fyi)

AzureMarker (approved):

SteveCookTU:
Was scrolling through and I noticed the use of enums for the parameters. In other services such as `fs` the `bitflags!` macro is used and even though these may not be bitflag related items would it be preferred for consistency over defining another macro?

AzureMarker:
> Was scrolling through and I noticed the use of enums for the parameters. In other services such as `fs` the `bitflags!` macro is used and even though these may not be bitflag related items would it be preferred for consistency over defining another macro?

If there's something that can reduce the duplication of converting between `u32` and the enum values, then I'd be open to a PR for that. But we should keep the in-memory representation the same (not make it into a bit field).

PR #65, opened 2022-07-17
Initial CAM implementation

SteveCookTU:
Hello all, wanted to start this PR as a basis of the initial CAM implementation. Before adding doc comments I wanted to get eyes on the simple wrapper functions. The only function not implemented is `CAMU_GetSuitableY2rStandardCoefficient`.

I want to use this PR sort of as a starting point of helping out the project and learning the preferred design and style for further contributions. 

Changes in other files are a result of `cargo fmt`,


ctru-rs/src/services/cam.rs
```diff
@@ -0,0 +1,1026 @@
+use bitflags::bitflags;
+use ctru_sys::Handle;
+
+pub struct Cam(());
+
+bitflags! {
+    #[derive(Default)]
+    pub struct CamPort: u32 {
+        const PORT_NONE = 0;
```
Meziu, line 9: You should use `ctru-sys` constants rather than hardcoded numbers.
This goes for all enums in the new module.

ctru-rs/examples/camera-image.rs
```diff
@@ -0,0 +1,201 @@
...
+        } else if !key_held.contains(KeyPad::KEY_R) {
+            held_r = false;
+        }
+
+        write_picture_to_frame_buffer_rgb_565(
+            gfx.top_screen.borrow_mut().get_raw_framebuffer(Side::Left),
+            &mut buf,
+            0,
+            0,
+            WIDTH as u16,
```
Meziu, line 99: Other than this specific line, there are many instances of `as i16` or `as u32`. Is there nothing to be done about it? (maybe the API is setup with the wrong parameters?)

ctru-rs/examples/camera-image.rs
```diff
@@ -0,0 +1,198 @@
...
+        if r < 0 {
+            panic!("Failed to close handle 2");
+        }
+    };
+
+    cam.activate(CamSelect::SELECT_NONE)
+        .expect("Failed to deactivate camera");
+}
+
+fn write_picture_to_frame_buffer_rgb_565(
```
Meziu, line 170: This function is doing weird pointer math to copy the image data to the framebuffer. Isn't there a simpler (or at least more abstract) way to handle this? It is really confusing for the user, which is only expected to see an example of how to use the modules.

example.rsf
```diff
@@ -0,0 +1,184 @@
+BasicInfo:
```
Meziu, line 1: See my review comment about this `.rsf` file.

.gitignore
```diff
@@ -4,3 +4,6 @@ Cargo.lock
 
 # IDE files
 .idea
+
+# 3ds files
+*.cia
```
Meziu, line 9: We haven't standardized anything like this. We may do something similar with `cargo-3ds` in the future, but it will most likely end up in the `target` directory for easy integration with Cargo. Also, look at the review comment.

Meziu (changes requested):

ctru-rs/examples/camera-image.rs
SteveCookTU: The issue isn't the API, more that numbers with different types are used in different kinds of math. `ctr-sys` takes the types they are casted to which I thought may want to stay consistent. If they are individually defined without reusing each other to determine values the need for casting could be removed.

ctru-rs/examples/camera-image.rs
SteveCookTU: I'll have to look into this. This math was taken from the `3ds-examples` repo from devkitpro.

example.rsf
SteveCookTU: I'm unable to see any review comment on the PR thread. Try reposting it or commenting it here.

ctru-rs/examples/camera-image.rs
SteveCookTU: Removed the pointer math. The values represented in the both output types are u16 hence was the slice was being casted to a `*const u16`. 

Knowing it will consistently be u16, thoughts on having the wrapper API do the casting and having a `&[u16]` or `&mut [u16]` being passed to the function?

example.rsf
Meziu: Basically: we don’t have any standard way to build `.cia` files yet, nor are we planning on having one. Designing support for `cargo-3ds` would (most likely) be the way to do it in the future. Moreover, our repository is a library and not a released application, so we won’t need a `.rsf` file here (counting it also has some problems with how its setup of which I won’t go into detail).

.gitignore
Meziu: (Look at the review comment about the `.rsf` file)

ctru-rs/examples/camera-image.rs
Meziu: In line 44 you set the Camera’s output format to `RGB_565`(it’s in the original example I bet, but I don’t know what’s the reasoning behind this since it’s a rare format for the framebuffer). Wouldn’t it be easier to use the standard `RGB8`? Maybe with a comfy `memcpy`.

SteveCookTU: According to libctru and the docs on 3dbrew it only accepts 2 output formats: `RGB565` and `YUV422`. I haven't found anything that mentions others. Both formats output pixels as u16.

example.rsf
SteveCookTU: That makes perfect sense. I had it there for building and testing the examples differently than a .3dsx file. I will remove them and possible open an issue on cargo-3ds for possible different output formats.

ctru-rs/examples/camera-image.rs
Meziu: First of all, this code (as-is) doesn't work to me. The top screen stays black.

Second thing: while I understand what the original code does, I suggest you should probably skip this whole "conversion" from `RGB565` to `RGB8`, and instead just use `top_screen.set_framebuffer_format` to set it to `RGB565`. This way the data is readable directly, but I believe it will still be needed to rotate the image of 90 degrees, as the screen's frambuffer is handled that way.

SteveCookTU: I'm not sure why it's not building for you, I have it running on my 3ds currently and just reinstalled everything to make sure I wasn't accidentally using an older version after changes.

I haven't done a deep dive through all of the screen stuff. I will see if just setting the top screen's format to `RGB565` works (but I believe it should).

The only thing I didn't do was init Ac which the example does with the intention of running it as a `3dsx` while I have been testing with a `CIA`. My knowledge on the 3ds services isn't enough to know if that would make a difference.

SteveCookTU: Rotating the image looks to be necessary. Only issue is that would also look messy transforming the buffer. Rotating a `u8` slice that should really be `u16` values might look more confusing than the conversion that's there now.

Meziu: Alright, just add a comment on top saying something like "This is unsafe code to copy and rotate the image buffer. The framebuffer should NOT be accessed by the CPU in normal circumstances, but it's quicker for this specific example".

ctru-rs/examples/camera-image.rs
```diff
@@ -0,0 +1,214 @@
...
+    cam.set_trimming(CamPort::PORT_CAM2, false)
+        .expect("Failed to disable trimming for Cam Port 2");
+
+    let mut buf = vec![0u8; BUF_SIZE];
+
+    gfx.flush_buffers();
+    gfx.swap_buffers();
+    gfx.wait_for_vblank();
+
+    let mut held_r = false;
```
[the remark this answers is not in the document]
Meziu: Oh wow, it looks to be hardware failure on my part. Sometimes it *does* stay still, but only if I smash the button full force on the exact center. It seemed weird, and the code in libctru was correct after reviewing it... I guess you should just use `keys_down` rather then `keys_held` and `held_r`. Even though it might fail sometimes, it's just more correct when looking at the documentation.

ctru-rs/examples/camera-image.rs
```diff
@@ -0,0 +1,214 @@
...
+
+    println!("\nPress R to take a new picture");
+    println!("Press Start to exit to Homebrew Launcher");
+
+    gfx.top_screen.borrow_mut().set_3d_enabled(false);
```
Meziu, line 72: What's the purpose of this? It's off by default.

SteveCookTU: I was testing stuff with 3D and forgot to remove it.

Decided against adding 3D because the example created a raw pointer to the slider value rather than using a lib function to retrieve it. Removing this line now.

ctru-rs/examples/camera-image.rs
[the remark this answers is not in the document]
SteveCookTU: I'm not sure the reasoning behind `keys_held` vs `keys_down` on their part. As you said keys_down is more accurate to the docs. Removed the held usage in favor of this.

ctru-rs/examples/camera-image.rs
SteveCookTU: Thoughts on the new version of the method? Instead of writing into the buffer directly it creates a converted version of the image and then does a copy of it into the buffer as seen in the `graphics-bitmap` example. The comment may not be necessary in this sense as it's a more clear example of how to do it without playing around with raw pointers.

SteveCookTU: Thinking about it, what's the size of the frame buffer on the `N3DS` or `O3DS`. If the image isn't the same size at that buffer then this could be causing undefined behavior by writing data outside of the buffer.

ctru-rs/src/services/cam.rs
```diff
@@ -0,0 +1,1027 @@
...
+        const SIZE_DS_LCD_X4      = ctru_sys::SIZE_DS_LCDx4;
+        const SIZE_CTR_TOP_LCD    = ctru_sys::SIZE_CTR_TOP_LCD;
+        const SIZE_CTR_BOTTOM_LCD = ctru_sys::SIZE_CTR_BOTTOM_LCD;
+    }
+}
+
+bitflags! {
+    #[derive(Default)]
+    pub struct CamFrameRate: u32 {
+        const FRAME_RATE_15       = ctru_sys::FRAME_RATE_15;
```
Meziu, line 68: Since the enum is named CamFrameRate, you can remove the `FRAME_RATE_` prefix from all variants.

ctru-rs/src/services/cam.rs
```diff
@@ -0,0 +1,1027 @@
...
+        const FLIP_HORIZONTAL = ctru_sys::FLIP_HORIZONTAL;
+        const FLIP_VERTICAL   = ctru_sys::FLIP_VERTICAL;
+        const FLIP_REVERSE    = ctru_sys::FLIP_REVERSE;
+    }
+}
+
+bitflags! {
+    #[derive(Default)]
+    pub struct CamSize: u32 {
+        const SIZE_VGA            = ctru_sys::SIZE_VGA;
```
Meziu, line 53: Since the enum is named CamSize, you can remove the `SIZE_` prefix from all variants.

ctru-rs/src/services/cam.rs
```diff
@@ -0,0 +1,1027 @@
...
+        const CONTEXT_A    = ctru_sys::CONTEXT_A;
+        const CONTEXT_B    = ctru_sys::CONTEXT_B;
+        const CONTEXT_BOTH = ctru_sys::CONTEXT_BOTH;
+    }
+}
+
+bitflags! {
+    #[derive(Default)]
+    pub struct CamFlip: u32 {
+        const FLIP_NONE       = ctru_sys::FLIP_NONE;
```
Meziu, line 43: Since the enum is named CamFlip, you can remove the `FLIP_` prefix from all variants.

ctru-rs/src/services/cam.rs
```diff
@@ -0,0 +1,1027 @@
...
+        const SELECT_OUT1_OUT2 = ctru_sys::SELECT_OUT1_OUT2;
+        const SELECT_IN1_OUT2  = ctru_sys::SELECT_IN1_OUT2;
+        const SELECT_ALL       = ctru_sys::SELECT_ALL;
+    }
+}
+
+bitflags! {
+    #[derive(Default)]
+    pub struct CamContext: u32 {
+        const CONTEXT_NONE = ctru_sys::CONTEXT_NONE;
```
Meziu, line 33: Since the enum is named CamContext, you can remove the `CONTEXT_` prefix from all variants.

ctru-rs/src/services/cam.rs
```diff
@@ -0,0 +1,1027 @@
+use bitflags::bitflags;
+use ctru_sys::Handle;
+
+pub struct Cam(());
+
+bitflags! {
+    #[derive(Default)]
+    pub struct CamPort: u32 {
+        const PORT_NONE = ctru_sys::PORT_NONE;
```
Meziu, line 9: Since the enum is named CamPort, you can remove the `PORT_` prefix from all variants.

ctru-rs/src/services/cam.rs
```diff
@@ -0,0 +1,1027 @@
...
+        const PORT_CAM1 = ctru_sys::PORT_CAM1;
+        const PORT_CAM2 = ctru_sys::PORT_CAM2;
+        const PORT_BOTH = ctru_sys::PORT_BOTH;
+    }
+}
+
+bitflags! {
+    #[derive(Default)]
+    pub struct CamSelect: u32 {
+        const SELECT_NONE      = ctru_sys::SELECT_NONE;
```
Meziu, line 19: Since the enum is named CamSelect, you can remove the `SELECT_` prefix from all variants.

ctru-rs/src/services/cam.rs
```diff
@@ -0,0 +1,1027 @@
...
+
+        const LENS_CORRECTION_DARK = ctru_sys::LENS_CORRECTION_DARK;
+        const LENS_CORRECTION_NORMAL = ctru_sys::LENS_CORRECTION_NORMAL;
+        const LENS_CORRECTION_BRIGHT = ctru_sys::LENS_CORRECTION_BRIGHT;
+    }
+}
+
+bitflags! {
+    #[derive(Default)]
+    pub struct CamOutputFormat: u32 {
```
Meziu, line 163: You should implement `TryFrom` here to handle `FramebufferFormat` to `CamOutputFormat` conversions.

Meziu (changes requested):

ctru-rs/examples/camera-image.rs
```diff
@@ -0,0 +1,205 @@
...
+        hid.scan_input();
+        keys_down = hid.keys_down();
+
+        if keys_down.contains(KeyPad::KEY_START) {
+            break;
+        }
+
+        if keys_down.bits() & KeyPad::KEY_R.bits() != 0 {
+            println!("Capturing new image");
+            gfx.flush_buffers();
```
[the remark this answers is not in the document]
SteveCookTU: Removed them. You're right, I'm not sure why I had them there to begin with.

ctru-rs/examples/camera-image.rs
```diff
@@ -0,0 +1,205 @@
...
+
+    while apt.main_loop() {
+        hid.scan_input();
+        keys_down = hid.keys_down();
+
+        if keys_down.contains(KeyPad::KEY_START) {
+            break;
+        }
+
+        if keys_down.bits() & KeyPad::KEY_R.bits() != 0 {
```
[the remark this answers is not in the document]
SteveCookTU: I think this is a question we should ask for all the examples. The issue with using `KeyPad::contains` or for any other bitflag is if I do any combination of buttons that simply contains the one specified, it will return true. 

For example, if I press A + START the application will exit because it contains the flag. For accuracy, I think masking the `keys_down` with the bits of the target button would be better.

ctru-rs/examples/camera-image.rs
```diff
@@ -0,0 +1,205 @@
...
+    let receive_event2 = cam
+        .set_receiving(
+            &mut buf[SCREEN_SIZE as usize..],
+            CamPort::PORT_CAM2,
+            SCREEN_SIZE,
+            buf_size as i16,
+        )
+        .expect("Failed to set receiving");
+
+    unsafe {
```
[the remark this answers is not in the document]
SteveCookTU: I can add a safe wrapper around the module. 

Thoughts on creating a `Handle` struct on the lib and implementing `Drop` for the handle could call the close event?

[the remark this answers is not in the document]
Meziu: If you want to implement this, open a separate PR. Otherwise I can look into this later.

ctru-rs/examples/camera-image.rs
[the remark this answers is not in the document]
Meziu: There is `.intersect` for that, which is a mask that returns a mask.

For example, to have A + START:
`!keys_down.intersect(KeyPad::KEY_A).intersect(KeyPad::KEY_START).is_empty()`

It is more verbose, but it's functionally the same and more readable. It's up to personal reference, but I don't find it problematic. Also, no examples use button combos right now, so I don't see why we should change the current source code in any case.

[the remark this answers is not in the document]
Meziu: You can just do unions for that.

Example:
`keys_down.contains(KeyPad::KEY_A | KeyPad::KEY_START)`

No examples use button combos though, so I still find no issue with just `.contains`.

ctru-rs/examples/camera-image.rs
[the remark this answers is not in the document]
SteveCookTU: I started working on this. Luckily a friend of mine has a very well documented version of this that I can port over.

ctru-rs/examples/camera-image.rs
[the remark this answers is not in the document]
SteveCookTU: Agreed but examples should probably infer intention as the picture in the example is meant to be take when R is pressed not any combination of keys that contains R.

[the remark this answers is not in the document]
SteveCookTU: Fair enough. I also sort of completely forgot they could just be compared since my brain thinking "bitflag". I'll change it back to contains for this example but I think the `buttons` example could be revisited then because it does use a bitmask to check for a single button and that's also what I was going off of.

ctru-rs/examples/camera-image.rs
```diff
@@ -0,0 +1,205 @@
...
+        if r < 0 {
+            panic!("Failed to close handle");
+        }
+        r = ctru_sys::svcCloseHandle(receive_event2);
+        if r < 0 {
+            panic!("Failed to close handle 2");
+        }
+    };
+
+    cam.activate(CamSelect::SELECT_NONE)
```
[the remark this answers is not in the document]
Meziu: Looking at the documentation, it's clear the different combinations are just bitwise OR versions. We can ditch the unions and just have 3 variants:
`Inward`
`Outward1`
`Outward2` (maybe if we understood which side is which we could call them `OutwardLeft` and `OutwardRight`).

These three variants are the only public ones available via our module, and thus the only one usable with `.activate()` (using bitwise ORs to include more than one).
After that, we can have `.deactivate()` which just sets it to `SELECT_NONE` under the hood.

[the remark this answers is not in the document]
SteveCookTU: Outward1 seems to the right camera (facing outward when taking a picture).

Also did testing and it seems the camera combinations can be activated individually and don't have to be passed through to the libctru function at the same time. The reason I thought otherwise was because when deactivating it is passing 0 which would "overwrite" the current active camera. 

That being said, I am going to make functions specific for each. I will do the same for the shutter sounds.

More things to look into:

When using one camera is it always set to `CamPort::CAM1`?
When using multiple are the ports set in the order they are activated?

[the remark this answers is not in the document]
SteveCookTU: The `IN1_OUT1` and other combination seems useless. Not sure if I'm missing something or it wasn't fully implemented or working in libctru to begin with. Need to test more with activating both outward to make sure it activates both.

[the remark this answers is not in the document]
Meziu: > The `IN1_OUT1` and other combination seems useless. Not sure if I'm missing something or it wasn't fully implemented or working in libctru to begin with. Need to test more with activating both outward to make sure it activates both.

In all honesty I have no idea how it could be using the inward and outward camera’s at the same time either, so that’s something strange.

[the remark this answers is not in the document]
SteveCookTU: Here are my findings:

Activating `IN1` deactivates `OUT1` and/or `OUT2`. (Similar to `NONE` deactivating the other options)
`CamPort::CAM1` always refers to `OUT1` when combined or whichever camera is activated when only one is.

With this in mind, what are your thoughts on a struct/module for each that manages each variation? 

`Inward1`
`OutwardLeft`
`OutwardRight`
`Outward3D` (for managing `OUT1` and `OUT2` and their ports combined)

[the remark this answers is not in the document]
Meziu: The enum variants look fine, and it depends on whether we have a separate function for each variation or not.

Regarding the cam selection for actually taking the pictures:
We should only really thinking about the case with both outward cameras, right? In any other case it will just use CAM1, so we can abstract it by just having a “take_picture” method in `Cam`.

When it comes to take two buffers, it is slightly different, but maybe we could pass an enum option (which holds either one or two Vec buffers). This means we would have to handle the memory allocation inside Cam, rather than passing a reference to a pre-built buffer, but I don’t know if that’s a bad idea or not.

ctru-rs/examples/camera-image.rs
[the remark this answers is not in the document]
Meziu: We should just have a method in `Cam` which stops the thread waiting for the picture (obviously this behaviour has to be documented in the API). It doesn’t make much sense to have a specific module to handle `svc` actions.

ian-h-chamberlain:
I think the API looks okay, but we could also consider something like we did for `Screen`, which is to have a trait/struct that implements common functionality and can be obtained _from_ the `Cam` service, logically representing a single camera (or maybe also multiple?). A rough sketch of what I'm thinking:

```rust
let cam = Cam::init().unwrap():
let mut rear_cams = cam.both_outward();
rear_cams.set_size(CamSize::CTR_TOP_LCD).unwrap();
rear_cams.set_output_format(CamOutputFormat::RGB_565).unwrap();

cam.activate(rear_cams).unwrap();
cam.start_capture(CamPort::BOTH).unwrap();

// etc...
```

I haven't look in great detail at the libctru C API, so maybe this doesn't make a ton of sense, but I think abstracting over the hardware like this to prevent accidentally crossing wires or mutating the wrong device state can be helpful, so just a suggestion. Feel free to ignore if you think it doesn't make sense or introduces too much complexity to the service (or, it could be implemented in another PR too).

ctru-rs/src/services/cam.rs
```diff
@@ -0,0 +1,1050 @@
...
+            let r = ctru_sys::CAMU_GetSleepCamera(&mut camera);
+            if r < 0 {
+                Err(r.into())
+            } else {
+                Ok(CamSelect::from_bits(camera).unwrap_or_default())
+            }
+        }
+    }
+
+    pub fn set_sleep_camera(&self, camera: CamSelect) -> crate::Result<()> {
```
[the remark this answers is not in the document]
SteveCookTU: They should, I'll change those. I think I originally had them but not sure why I changed it to immutable

SteveCookTU:
> ```rust
> let cam = Cam::init().unwrap():
> let mut rear_cams = cam.both_outward();
> rear_cams.set_size(CamSize::CTR_TOP_LCD).unwrap();
> rear_cams.set_output_format(CamOutputFormat::RGB_565).unwrap();
> 
> cam.activate(rear_cams).unwrap();
> cam.start_capture(CamPort::BOTH).unwrap();
> 
> // etc...
> ```

Something like this is what I was thinking as well in the comment I mentioned earlier. We could change each cam variant to a struct with their own implementation of a `Camera` trait.

The three singular cameras could use the default implementation and then the two outer cameras together can have its own. This would also reduce all the enums visible to the user to just the different camera settings.

Also something to consider is using something like `num_enum` for non-bitflag enums. I think it would help a lot with removing all the `From` implementation boilerplate.

SteveCookTU:
Working on an idea of this and was curious, is there a reason the screens are wrapped in a `RefCell` within the `gfx` module?

AzureMarker:
> Working on an idea of this and was curious, is there a reason the screens are wrapped in a `RefCell` within the `gfx` module?

There's some context here: https://github.com/Meziu/ctru-rs/pull/22

SteveCookTU:
Pushed an initial version of this to my local. Didn't want to commit to the current PR just in case. The initial idea can be found `here`. I can also open a separate PR for that branch if necessary.

AzureMarker:
> Also something to consider is using something like `num_enum` for non-bitflag enums. I think it would help a lot with removing all the `From` implementation boilerplate. I didn't want to add this without asking though because it's another dependency to tack on

I'm ok with this, it looks like it will cut down on the boilerplate by a lot.

ctru-sys/src/bin/docstring-to-rustdoc.rs
```diff
@@ -33,8 +33,7 @@ fn main() -> io::Result<()> {
         .map(|v| {
             // Only modify lines with the following structure: `` #[doc ... ] ``
             if v.trim_start().starts_with("#[doc") && v.trim_end().ends_with("]") {
```
AzureMarker, line 35: While you're at it, might as well fix this clippy warning:
```
warning: single-character string constant used as pattern
  --> ctru-sys/src/bin/docstring-to-rustdoc.rs:35:78
   |
35 |             if v.trim_start().starts_with("#[doc") && v.trim_end().ends_with("]") {
   |                                                                              ^^^ help: try using a `char` instead: `']'`
   |
   = note: `#[warn(clippy::single_char_pattern)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#single_char_pattern
```

ctru-rs/src/services/cam.rs
```diff
@@ -0,0 +1,1050 @@
...
+            let r = ctru_sys::CAMU_FlipImage(camera.bits(), flip.bits(), context.bits());
+            if r < 0 {
+                Err(r.into())
+            } else {
+                Ok(())
+            }
+        }
+    }
+
+    pub fn set_detail_size(
```
AzureMarker, line 641: Clippy doesn't like this. :paperclip:
Maybe use tuples for x/y pairs, or ignore the lint.
```
warning: this function has too many arguments (9/7)
   --> ctru-rs/src/services/cam.rs:641:5
    |
641 | /     pub fn set_detail_size(
642 | |         &self,
643 | |         camera: CamSelect,
644 | |         width: i16,
...   |
650 | |         context: CamContext,
651 | |     ) -> crate::Result<()> {
    | |__________________________^
    |
    = note: `#[warn(clippy::too_many_arguments)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#too_many_arguments
```

ctru-rs/src/services/cam.rs
```diff
@@ -0,0 +1,1050 @@
+use bitflags::bitflags;
+use ctru_sys::Handle;
+use crate::services::gspgpu::FramebufferFormat;
+
+pub struct Cam(());
```
AzureMarker, line 5: It would be great to have some documentation on the structs/enums and functions (and a quick line for the module to explain what it is for).

ctru-rs/src/services/cam.rs
```diff
@@ -0,0 +1,1050 @@
...
+            let r = ctru_sys::CAMU_ClearBuffer(port.bits());
+            if r < 0 {
+                Err(r.into())
+            } else {
+                Ok(())
+            }
+        }
+    }
+
+    pub fn get_vsync_interrupt_event(&self, port: CamPort) -> crate::Result<u32> {
```
AzureMarker, line 280: Maybe it would be clearer to return the `Handle` type?
(same comment for the other similar functions)

ctru-rs/src/services/cam.rs
```diff
@@ -0,0 +1,1050 @@
...
+        unsafe {
+            let mut data = 0;
+            let r = ctru_sys::CAMU_ReadMcuVariableI2cExclusive(&mut data, camera.bits(), addr);
+            if r < 0 {
+                Err(r.into())
+            } else {
+                Ok(data)
+            }
+        }
+    }
```
AzureMarker, line 894: Do you think these functions might need to be marked unsafe?

ctru-rs/src/services/cam.rs
```diff
@@ -0,0 +1,1050 @@
...
+    pub fn driver_finalize(&self) -> crate::Result<()> {
+        unsafe {
+            let r = ctru_sys::CAMU_DriverFinalize();
+            if r < 0 {
+                Err(r.into())
+            } else {
+                Ok(())
+            }
+        }
+    }
```
AzureMarker, line 995: Is it safe to call these functions multiple times? Is `driver_initialize` required for any of the other functions? If either answers are yes, maybe we should have another struct to enforce the init/deinit and function access.

Edit: looking at the docs for `camInit` and `camExit`, they call these functions automatically. Maybe we shouldn't expose them then?

AzureMarker (commented):
This is cool! Thanks for the PR.

ctru-rs/examples/camera-image.rs
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        let img = convert_image_to_rgb8(buf, 0, 0, WIDTH as usize, HEIGHT as usize);
+
+        unsafe {
+            gfx.top_screen
+                .borrow_mut()
+                .get_raw_framebuffer(Side::Left)
+                .ptr
+                .copy_from(img.as_ptr(), img.len());
```
AzureMarker, line 89: Isn't `img` the size of two screens? Therefore writing all of its contents into a framebuffer the size of one screen would mean you're writing past the buffer's limits?

ctru-rs/src/services/cam.rs
SteveCookTU: In favor of removing these possibly because they may not have much use and the unsafe factor. Thoughts?

ctru-rs/src/services/cam.rs
SteveCookTU: Same idea for this as the register functions

ctru-rs/examples/camera-image.rs
SteveCookTU: I'm going to change `BUF_SIZE` to represent a single image. I mentioned this before in the comment thread about the conversion function. The reason BUF_SIZE was the size of multiple was for 3D as the cameras would take two images. In that case, writing to the frame buffer would be a little different as well I believe

ctru-rs/src/services/cam.rs
SteveCookTU: With the talk about the `SVC` module, it seems it may be best to not expose handles like this for safety. If `SVC` functions stay internal, the user will never have a need for getting handles and therefore any handle related function should probably not be exposed or even removed if never used.

ctru-rs/examples/camera-image.rs
SteveCookTU: After playing around with this, it seems like the rgb8 format for the frame buffer may have something to do with how it's read as two images. I tried halving `BUF_SIZE` but it causes an out of bounds exception when converting. I'd have to look into 3D more probably to understand how that's actually being sent to the buffer cause it may not be as straight forward as sending the two halves.

SteveCookTU:
I decided to merge the structs branch into this one as it seems it would be a favored starting point to refactor what was originally there. I think it's a better idea to wrap the hardware with structs as @ian-h-chamberlain mentioned. 

There are definitely changes to be made for what functions to safely expose to the user and I have a way to do that but I want to ask first what everyone believes should be safe to provide. Currently I believe the only things that should be exported are functions related to camera settings (this excludes `Handle` results and other svc related functions such as vsync timings and synching handles).

ctru-rs/src/services/cam.rs
AzureMarker: I don't think it's a concern here to just use the type from `ctru_sys`. It doesn't change anything relevant to safety. But, this code might change anyway based on @Meziu's comments.

ctru-rs/src/services/cam.rs
AzureMarker: Yeah, it might be better to remove them since they seem unsafe and probably aren't used very often. The wrapper is simple so it shouldn't be hard to call the ctru_sys functions directly.

AzureMarker:
We should still expose some way of waiting for the photo to be taken. The details of how that waiting happens can be hidden behind a simple function, like @Meziu suggested.

SteveCookTU:
> We should still expose some way of waiting for the photo to be taken. The details of how that waiting happens can be hidden behind a simple function, like @Meziu suggested.

In the current version, `take_picture` starts and waits for the image under the hood. The returned result contains the image bytes.

ctru-sys/src/bin/docstring-to-rustdoc.rs
AzureMarker: Moved this comment to the CI PR:
https://github.com/Meziu/ctru-rs/pull/66/files#r929514001

ctru-rs/src/services/cam.rs
```diff
@@ -0,0 +1,1020 @@
...
+        const OUT1      = ctru_sys::SELECT_OUT1;
+        const IN1       = ctru_sys::SELECT_IN1;
+        const OUT2      = ctru_sys::SELECT_OUT2;
+        const OUT1_OUT2 = ctru_sys::SELECT_OUT1_OUT2;
+    }
+}
+
+bitflags! {
+    #[derive(Default)]
+    pub struct CamContext: u32 {
```
Meziu, line 36: The `CamContext` (aka `CAMU_Context` in the raw bindings) seems a difficult to implement/useless addition. As far as I understand, a context is simply a set of "options" which you can define for the camera environment. After setting the options for each context (of which there are only 2), you can switch between them at will, without having to mess with the options again. It's cool as a concept, but seeing how it has to be used for each cam and etc., it quite easily gives problems with the camera's internal state.

I'd suggest to avoid its use for now, and de-implement it. We can add it back once the API gets more stable. This also counts for the functions that require it.

Meziu (changes requested):
I'd like to get this PR merged soon, as it has been stuck here for a while now, but until some nits are solved, and that issue with 3D buffers is understood, we need some more work on it.

ctru-rs/src/services/cam.rs
```diff
@@ -0,0 +1,1020 @@
+use crate::services::gspgpu::FramebufferFormat;
+use bitflags::bitflags;
+use ctru_sys::Handle;
+use std::cell::RefCell;
+
+pub struct Cam {
```
Meziu, line 6: This should be `#[non_exhaustive]`, so that it can't be created by users manually.

ctru-rs/src/services/cam.rs
Meziu: Related to my `CamContext` comments.

ctru-rs/src/services/cam.rs
SteveCookTU: Would it be ok to instead force the use of `CamContext::A` since that's the default when the service is initialized? Then we don't have to remove the functions, just remove the context from the signatures and settings can still be used.

Meziu: > Would it be ok to instead force the use of `CamContext::A` since that's the default when the service is initialized? Then we don't have to remove the functions, just remove the context from the signatures and settings can still be used.

That’s what I was thinking.

Meziu (changes requested):

ctru-rs/examples/camera-image.rs
```diff
@@ -0,0 +1,140 @@
...
+        if keys_down.contains(KeyPad::KEY_START) {
+            break;
+        }
+
+        if keys_down.contains(KeyPad::KEY_R) {
+            println!("Capturing new image");
+            buf = camera
+                .take_picture(
+                    WIDTH.try_into().unwrap(),
+                    HEIGHT.try_into().unwrap(),
```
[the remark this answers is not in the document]
SteveCookTU: `WIDTH` and `HEIGHT` are re-used after setting the size to specify how many bytes from the picture are actually sent. To my understanding you can have a camera be a specified size but can take a picture of a section of the full view of the camera. They are re-used to calculate the maximum bytes transferred and also how many to actually transfer when taking the picture.

ctru-rs/examples/camera-image.rs
```diff
@@ -0,0 +1,140 @@
...
+                    transfer_unit,
```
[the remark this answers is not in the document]
SteveCookTU: Along with this I realized there are multiple functions within take_picture itself that were available in the public interface through the trait and moved them into it. There were related to steps in between the start and end of taking a picture such as `set_receiving`. None of the removed functions should be used outside of taking a picture (or video when that's eventually added) so they have been changed to be internal usage.

ctru-rs/examples/camera-image.rs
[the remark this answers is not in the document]
Meziu: Are these sizes calculated from the top-left corner of the image? Either way, this makes `set_size` quite misleading, maybe we should rename it `set_view` or `set_max_size` plus adding some comments to explain it.

ctru-rs/src/services/cam.rs
```diff
@@ -898,36 +774,87 @@ pub trait Camera {
...
+                transfer_unit,
+                width as i16,
+                height as i16,
+            );
+            if r < 0 {
+                return Err(r.into());
+            }
+        };
 
-        self.activate()?;
```
Meziu: Maybe this can be useful for other use cases? For example, obtaining a continous view of the camera, but I haven’t looked at it directly. Also, it looks like the camera is never deactivated at the end of this function, did you forget?

SteveCookTU: The libctru example of showing an active view of the camera uses other low level functions to do so including dealing with handles to detect interrupts. I think it may be better to have a separate function for previewing which is pretty much rendering an active video from the camera. [Here is the example.](https://github.com/devkitPro/3ds-examples/blob/master/camera/video/source/main.c)

Meziu: 1. From the looks of it it seems they are actually just taking “multiple” pictures and flushing them to the framebuffer (with the only difference of checking the svchandle for a prolonged time rather than a single wait). Still, in our case, at least right now, the camera should be deactivated at the end of the `take_picture` function.

ctru-rs/src/services/cam.rs
```diff
@@ -0,0 +1,899 @@
+use crate::services::gspgpu::FramebufferFormat;
+use bitflags::bitflags;
+use ctru_sys::Handle;
+use std::cell::RefCell;
+use std::time::Duration;
+
+#[non_exhaustive]
+pub struct Cam {
+    pub inner_cam: RefCell<InwardCam>,
```
Meziu, line 9: Are you sure these should be `RefCell`s? We use `RefCell` in `gfx` because in that case the two screens act as separate objects, but in this case I don’t think one can use multiple types of cameras at once. Maybe they should be just held by the struct, and the borrow rules will handle the safety automatically.

SteveCookTU: I think I ended up using them because I saw what was being done in gfx. It made thing easier but I know it's not necessarily the right thing to do. I switched back to just the structs and enclosed setting the camera settings in the example so the mutable borrow is dropped before the loop. Otherwise, the shutter sound wouldn't be able to play since it takes an immutable reference.

ctru-rs/examples/camera-image.rs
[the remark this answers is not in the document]
SteveCookTU: I plan on adding docs/comments very soon. What are your thoughts on naming it `set_view_size`?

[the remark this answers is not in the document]
Meziu: Good to me, but it clearly needs some comments to document it properly.

Meziu:
Other than documentation, I don’t know if there is anything else to check. @ian-h-chamberlain @AzureMarker could you re-review it?

SteveCookTU:
I have a final that I'm studying for to take tomorrow morning. Afterwards I can work on documentation.

ctru-rs/examples/camera-image.rs
```diff
@@ -0,0 +1,138 @@
...
+use ctru::gfx::{Screen, Side};
+use ctru::services::cam::{Cam, CamOutputFormat, CamShutterSoundType, CamSize, Camera};
+use ctru::services::hid::KeyPad;
+use ctru::services::{Apt, Hid};
+use ctru::Gfx;
+use std::time::Duration;
+
+const WIDTH: usize = 400;
+const HEIGHT: usize = 240;
+const BUF_SIZE: usize = WIDTH * HEIGHT * 2 * 2;
```
AzureMarker, line 11: As part of the documentation, please add a comment explaining why it's multiplied by 2 twice.

ctru-rs/examples/camera-image.rs
```diff
@@ -0,0 +1,138 @@
...
+        camera
+            .set_auto_exposure(true)
+            .expect("Failed to enable auto exposure");
+        camera
+            .set_auto_white_balance(true)
+            .expect("Failed to enable auto white balance");
+
+        camera
+            .set_trimming(false)
+            .expect("Failed to disable trimming");
```
AzureMarker, line 56: nit: All of these setting calls can be grouped together.

ctru-rs/src/services/cam.rs
```diff
@@ -0,0 +1,898 @@
...
+            let r = ctru_sys::CAMU_SetAutoExposure(self.camera_as_raw(), enabled);
+            if r < 0 {
+                Err(r.into())
+            } else {
+                Ok(())
+            }
+        }
+    }
+
+    fn is_auto_exposure(&self) -> crate::Result<bool> {
```
AzureMarker, line 450: nit: Maybe rename to `is_auto_exposure_enabled`?
Similar functions have the same issue.

AzureMarker (commented):
It looks pretty good, just needs docs and some small changes.

ian-h-chamberlain (approved):

ian-h-chamberlain:
LGTM

SteveCookTU:
Apologies for the delay, I finally had the time today to add probably the most basic comments.

Meziu (approved):
@AzureMarker i'd like a final opinion before merging the changes.

ctru-rs/src/services/cam.rs
```diff
@@ -0,0 +1,952 @@
...
+            if r < 0 {
+                Err(r.into())
+            } else {
+                Ok(data)
+            }
+        }
+    }
+
+    /// Sets the camera as the current sleep camera
+    // TODO: Explain sleep camera
```
AzureMarker, line 784: There's a to-do comment here

AzureMarker (commented):
Left some small comments, and there's a few existing comments in the example file. Implementation though looks good.

ctru-rs/src/services/cam.rs
SteveCookTU: I put a todo because I personally was unable to figure out what the sleep camera really does. If there is knowledge on this I can add the information.

ctru-rs/src/services/cam.rs
```diff
@@ -0,0 +1,952 @@
...
+                x_end,
+                y_end,
+            }
+        }
+    }
+}
+
+/// Represents data used by the camera to calibrate image quality
+#[derive(Default)]
+pub struct ImageQualityCalibrationData(ctru_sys::CAMU_ImageQualityCalibrationData);
```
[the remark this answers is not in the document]
SteveCookTU: Good catch. The inner should be public. Would it be preferred to allow direct access to the inner struct or repeat the fields and implement `From<ImageQualityCalibrationData>` for `CAMU_ImageQualityCalibrationData`?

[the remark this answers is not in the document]
AzureMarker: You can give direct access

ctru-rs/src/services/cam.rs
AzureMarker: Oh no problem, I'll resolve this then.

AzureMarker (approved):

PR #66, opened 2022-07-17
Initial attempt at CI with GitHub Actions

ian-h-chamberlain:
Since this project is starting to attract a few more people, I thought it would be a good idea to introduce some basic CI (even if we can't run e.g. unit tests, examples).

This relies on the std PR being merged, and unfortunately is broken since the `std::thread` PR hasn't been merged yet, but we could either `#[cfg]` out the broken example, or just wait on this until there is a new nightly that has been built with `std::thread`.

Either way, I thought I'd put up the PR to show a starting point, and also to see if I could test an actual actions run. @Meziu it might require you to allow Actions runs in order to test this, since there haven't been any up until now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,64 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+env:
+  # https://blog.rust-lang.org/2022/06/22/sparse-registry-testing.html
+  CARGO_UNSTABLE_SPARSE_REGISTRY: "true"
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    container: devkitpro/devkitarm
+    steps:
+      - if: ${{ env.ACT }}
+        name: Hack container for local development
+        run: |
+            curl -fsSL https://deb.nodesource.com/setup_18.x | sudo -E bash -
+            sudo apt-get install -y nodejs
+
+      - name: Checkout branch
+        uses: actions/checkout@v2
+
+      - name: Setup default Rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          components: clippy, rustfmt, rust-src
+          profile: minimal
+          toolchain: nightly
+          default: true
+
+      - name: Install build tools for host
+        run: sudo apt-get install -y build-essential
+
+      - name: Install cargo-3ds
+        uses: actions-rs/cargo@v1
+        with:
+          command: install
+          # TODO: this should probably just be a released version from crates.io
+          # once cargo-3ds gets published somewhere...
+          args: >-
+            --git https://github.com/rust3ds/cargo-3ds
+            --rev 913645665391ea715bc96910bda35f98a4536a6e
+
+      - name: Check formatting
+        run: cargo fmt --all --verbose -- --check
+
+      # For some reason, `cargo clippy` doesn't seem to work properly with
+      # -Zbuild-std (and thus with `cargo 3ds`). Maybe would be helped by
+      # https://github.com/rust3ds/cargo-3ds/pull/21 ?
+      - name: Cargo check
+        run: cargo 3ds check --color=always --workspace --verbose --all-targets
+        env:
+          RUSTFLAGS:
+            --deny=warnings
+
+  # TODO: it would be nice to actually build 3dsx for examples/tests, etc.
+  # and run it somehow, but exactly how remains to be seen. Also, we probably
+  # need the std::thread PR to land before a lot of the examples are runnable.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,15 @@ env:
 
 jobs:
   lint:
+    strategy:
+      matrix:
+        toolchain:
+          # Run against a "known good" nightly
+          - nightly-2022-07-18
+          # Check for breakage on latest nightly
+          - nightly
+    # But if latest nightly fails, allow the workflow to continue
+    continue-on-error: ${{ matrix.toolchain == 'nightly' }}
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


Meziu:
Good idea, but it's needless to say that having unit testing available would be golden. Is formatting all we may need right now?

ian-h-chamberlain:
I have been trying to devise a way to run unit / other tests, but so far the only options I've come up with are kind of hacky / not ideal for a CI workflow. It seems reasonable to me to start with just formatting/linting (hopefully can get `clippy` figured out) and then add proper tests / running stuff in the future.

Meziu:
@AzureMarker didn’t you manage to get `clippy` running? I personally never looked too much into it.

AzureMarker:
Yeah, I got clippy working on a custom rustc. You basically build it and then copy around the tools binaries into the right sysroot bin folder (where rustc is).

But hopefully soon we can just use nightly :).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,58 @@
...
+      # https://github.com/rust3ds/cargo-3ds/pull/21 ?
+      - name: Cargo check
+        run: cargo 3ds check --color=always --workspace --verbose --all-targets
+        env:
+          RUSTFLAGS:
+            --deny=warnings
+
+  # TODO: it would be nice to actually build 3dsx for examples/tests, etc.
+  # and run it somehow, but exactly how remains to be seen. Also, we probably
+  # need the std::thread PR to land before a lot of the examples are runnable.
```
AzureMarker, line 58: I tried using `--exclude` to ignore the thread-related examples, but it doesn't seem to work with examples :confused:.
https://doc.rust-lang.org/cargo/commands/cargo-build.html#package-selection

AzureMarker:
I did some tests locally trying to get clippy to work with nightly on the 3DS target. For some reason it doesn't work, even though we pass the build-std option.

SteveCookTU:
Thoughts on using something like [mocktopus](https://docs.rs/mocktopus/latest/mocktopus/) for tests? 

Also for clippy, without a prebuilt std I've been able to run it with `cargo clippy -Z build-std --target armv6k-nintendo-3ds` unless there is more to be checked that I'm missing. If you're running clippy through cargo-3ds, is there a chance the current build of `cargo-3ds` isn't passing the arguments correctly?

AzureMarker:
Most of what we're testing is how the 3DS reacts to the stuff we do (ex. does the networking stack accept this, are these words printed to the screen, etc), and also it's mostly a wrapper around libctru so there shouldn't be much more than wrapper code. I don't think mocking will help that much.

What toolchain are you using to run clippy? Nightly or a local build?

SteveCookTU:
> What toolchain are you using to run clippy? Nightly or a local build?

nightly-x86_64-pc-windows-msvc

ian-h-chamberlain:
I figured something out about clippy: https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/-Zbuild-std.20and.20clippy.3F/near/290622458

So I think from reading the code, we'd need either https://github.com/rust3ds/cargo-3ds/pull/21 or a change to reorder `-Z build-std` to be _after_ the subcommand, in order for `cargo +nightly 3ds clippy` to work as expected. In the meantime just passing the target + `build-std` manually should work, I'll see if I can push that change here as a workaround meanwhile.

AzureMarker:
I opened a PR to fix `cargo +nightly 3ds clippy`:
https://github.com/rust3ds/cargo-3ds/pull/24

ctru-sys/src/bin/docstring-to-rustdoc.rs
```diff
@@ -33,8 +33,7 @@ fn main() -> io::Result<()> {
         .map(|v| {
             // Only modify lines with the following structure: `` #[doc ... ] ``
             if v.trim_start().starts_with("#[doc") && v.trim_end().ends_with("]") {
```
AzureMarker, line 35: While you're at it, might as well fix this clippy warning:
```
warning: single-character string constant used as pattern
  --> ctru-sys/src/bin/docstring-to-rustdoc.rs:35:78
   |
35 |             if v.trim_start().starts_with("#[doc") && v.trim_end().ends_with("]") {
   |                                                                              ^^^ help: try using a `char` instead: `']'`
   |
   = note: `#[warn(clippy::single_char_pattern)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#single_char_pattern
```

.github/workflows/ci.yml
ian-h-chamberlain: Pushed a change with a feature to disable building the thread examples unless `--all-features` (or the explicit `std-threads` feature) is enabled.

AzureMarker:
You've probably run this CI somewhere, but I can't find where. Do you have a link, or does @Meziu need to enable it in the repo?

Meziu:
> You've probably run this CI somewhere, but I can't find where. Do you have a link, or does @Meziu need to enable it in the repo?

To run it in this branch I believe I would have to merge it first into master. Actions are already activated, it just isn’t present. Run it in a fork.

ian-h-chamberlain:
For testing, I have been using https://github.com/nektos/act (plus a little bit of local hackery to get `node` to work in the container, basically using what they posted in https://github.com/nektos/act/issues/917#issuecomment-1074421318).

@Meziu I believe the `on: pull_request` configuration should make it attempt to run the action before merging, but since I have never run actions on this repo before, you may need to manually grant the action as described in https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks ?

AzureMarker (approved):

Meziu:
I've opened https://github.com/Meziu/ctru-rs/pull/68/ to check.
Turns out the latest commits weren't formatted, so I had to do that too, but that means this works perfectly!

PR #67, opened 2022-07-20
Add SVC service wrapper

SteveCookTU:
This is a large PR which wraps the SVC service of ctru. Minimal documentation is here but there is at least some for a decent understanding. 

This is a slightly major/important module for creating services or applications on the 3DS. It allows for proper communication between services along with synchronizing them through SVC calls as seen in the example of #65. 

There are also 2 extra functions not given in libctru. One is Luma3DS only (which I may recommend removing only because Luma may be the custom OS of choice but it's not the only one). The other is to "steal" services as some have a restriction on how many handles can be owned at a time. I am unsure if this is just undocumented on libctru or is luma only as well.

### DIFF
--- a/ctru-rs/src/lib.rs
+++ b/ctru-rs/src/lib.rs
@@ -77,6 +77,7 @@ pub mod error;
 pub mod gfx;
 pub mod services;
 pub mod srv;
+pub mod svc;
 
 cfg_if::cfg_if! {
     if #[cfg(all(feature = "romfs", romfs_exists))] {
--- a/ctru-rs/src/svc.rs
+++ b/ctru-rs/src/svc.rs
@@ -1,0 +1,553 @@
+use std::arch::asm;
+use libc::c_void;
+use ctru_sys::{DebugEventInfo, svcAcceptSession, svcBreak, svcContinueDebugEvent, svcCreateEvent, svcCreateMemoryBlock, svcDebugActiveProcess, svcExitProcess, svcGetProcessDebugEvent, svcGetProcessInfo, svcGetProcessList, svcOpenProcess, svcQueryDebugProcessMemory, svcReadProcessMemory, svcReplyAndReceive, svcSignalEvent, svcSleepThread, svcUnmapMemoryBlock, svcWaitSynchronization, svcWriteProcessMemory};
+use crate::Error;
+
+#[derive(PartialEq, Debug)]
+pub struct Handle(u32);
+
+/// An abstraction on top of resource handles to enforce type safety.
+/// A Handle should only be made from a resource that is guaranteed to be a unique copy.
+/// When a Handle is dropped, the underlying resource handle is closed.
+/// Handles are intentionally non-copyable to avoid using Handles that have already been closed.
+impl Handle {
+    /// Returns the raw u32 handle
+    /// # Safety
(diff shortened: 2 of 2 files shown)


SteveCookTU:
The problem I see with not providing this to the user is that we cut off the capabilities of what could be created with the library. Background services or "sysmodules" wouldn't be possible and other advanced techniques that many existing applications use such as `PKSM`.

Meziu:
> The problem I see with not providing this to the user is that we cut off the capabilities of what could be created with the library. Background services or "sysmodules" wouldn't be possible and other advanced techniques that many existing applications use such as `PKSM`.

If the user is looking into the development of “sysmodules” they shouldn’t be trying to use a safe wrapper around `libctru`. This stuff is so low level and out of our scope (which mainly targets userland apps) that I don’t see the point in adding it. Why would we go out of our way to build an unsafe layer which is probably not good enough in the scope of a thin system module? They should (and probably will) just use `ctru-sys` instead.

For any services which have functionality requiring the use of `svc` we can provide basic functionality which ensures safety via very strong and explicit code measures (like having to hold a struct for the whole duration of the process to ensure the service exists).

I’m sorry to have asked you about this PR, as I wasn’t aware of the true implications of this set of functions. This is simply not what `ctru-rs` is meant to be doing.

SteveCookTU:
> I’m sorry to have asked you about this PR, as I wasn’t aware of the true implications of this set of functions. This is simply not what `ctru-rs` is meant to be doing.

No worries! This helps me understand the extent ctru-rs will be. The only things I might say are truly unsafe about SVC are the memory mapping functions otherwise nothing else would cause undefined behavior unless there are bugs within the OS of the system. Anything that fails just leads to a runtime error but shouldn't be causing any issues within the system itself. What keeps everything safe is guaranteeing the handles are closed which the Drop implementation would provide.

I am no expert on this personally but this is from an explanation I got from others who have done a bit of work on the 3DS

Meziu:
> The only things I might say are truly unsafe about SVC are the memory mapping functions otherwise nothing else would cause undefined behaviour unless there are bugs within the OS of the system. Anything that fails just leads to a runtime error but shouldn't be causing any issues within the system itself. What keeps everything safe is guaranteeing the handles are closed which the Drop implementation would provide.

The problems here isn’t really with memory safety (which is also a risk in any case), but more with breaking the user contract. This effectively is a way to avoid the mutability checks imposed by Rust’s compiler. Having this “shortcut” and access to system administration basically erases every way to keep track of the changes within the application. This type of control is useful, but only in an unsafe environment (and at that point, why make an API if the whole thing has to be unsafe either way).


What I want you to know is that `unsafe` in `ctru-rs` doesn’t have to be correlated to segmentation faults or memory management, but even just an “unsupervised” way to control the program. Hope this clears up your doubts.

AzureMarker:
> What keeps everything safe is guaranteeing the handles are closed which the Drop implementation would provide.

BTW, `drop` is not guaranteed to be called in all cases. Leaking memory is safe! :upside_down_face: 
https://doc.rust-lang.org/nomicon/leaking.html

SteveCookTU:
I understand now the usage for the library, definitely cleared things up. That being said I'm going to go ahead and delete the branch! Don't think it aligns with the library's goal

PR #70, opened 2022-09-18
Prelude module addition

Meziu:
Seeing the state of the many examples, and the general use of this library, I thought adding a simple `prelude` module would be  a welcomed addition to avoid many lines of repeating imports. For now I've added basic modules I would see fit in most applications (must-use like `Apt`, `Gfx` and `Hid`, plus some helpers like `Soc`, `KeyPad` and `Console`). Since the main use of any prelude module is to aid simple (usually single-file) projects, I've added only this basic functionality, but tell me if you think some other modules could be helpful.

@ian-h-chamberlain @AzureMarker

### DIFF
--- a/ctru-rs/examples/buttons.rs
+++ b/ctru-rs/examples/buttons.rs
@@ -1,7 +1,4 @@
-use ctru::console::Console;
-use ctru::gfx::Gfx;
-use ctru::services::apt::Apt;
-use ctru::services::hid::{Hid, KeyPad};
+use ctru::prelude::*;
 
 fn main() {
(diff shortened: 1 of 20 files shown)


AzureMarker:
The reasoning makes sense, though I have a dislike for preludes because they introduce a second import path for the same thing, and make it harder to tell where something comes from just by looking at the imports. But, if the prelude is kept reasonable enough then I'm ok with it.

AzureMarker (approved):

ian-h-chamberlain (approved):

PR #74, opened 2022-10-09
Improve the ergonomics of calling ``ctru_sys`` functions

Techie-Pi:
Right now, most calls to ``ctru_sys`` functions look like this:
```rs
let r = ctru_sys::aptInit();
if r < 0 {
    Err(r.into())
} else {
    Ok(Apt(()))
}
```

This PR adds the ``LibCtruResult`` struct (though the name can be changed), which implements the ``Try`` trait to allow the usage of the ``?`` operator. The example above now looks like this:
```rs
LibCtruResult(ctru_sys::appInit())?;
Ok(Apt(()))
```

This makes the code much more simpler, and both do exactly the same thing as it can be seen [here](https://github.com/Techie-Pi/ctru-rs-meziu/blob/28cb72addd2b9836df4e56ceeae416f734d83703/ctru-rs/src/error.rs#L23).

This could be implemented in a macro, but putting the ``i32`` returned from the ``ctru_sys`` functions into a struct defined by the crate allows the future implementation of further abstractions around the _raw_ errors provided by ``ctru_sys`` to cover cases [that can't be covered right now](https://github.com/Techie-Pi/ctru-rs-meziu/blob/28cb72addd2b9836df4e56ceeae416f734d83703/ctru-rs/src/services/fs.rs#L396).

CI is failing due to [this cam function](https://github.com/Techie-Pi/ctru-rs-meziu/blob/28cb72addd2b9836df4e56ceeae416f734d83703/ctru-rs/src/services/cam.rs#L278) not being used (maybe it was meant to be public?, @SteveCookTU)

### DIFF
--- a/ctru-rs/src/error.rs
+++ b/ctru-rs/src/error.rs
@@ -15,6 +15,7 @@ impl Try for LibCtruResult {
     type Output = ();
     // This type is passed to [FromResidual::from_residual] when the LibCtruResult is an error,
     // so this type implies "this is a result than CAN'T be `Ok`" (Infallible is the same as !)
+    // The purpose of this type is to _only_ bring information about the *Error*
     type Residual = crate::Result<core::convert::Infallible>;
 
     fn from_output(_: Self::Output) -> Self {
--- a/ctru-rs/src/error.rs
+++ b/ctru-rs/src/error.rs
@@ -1,10 +1,37 @@
 use std::error;
 use std::ffi::CStr;
 use std::fmt;
+use std::ops::{ControlFlow, FromResidual, Try};
 
 use ctru_sys::result::{R_DESCRIPTION, R_LEVEL, R_MODULE, R_SUMMARY};
 
 pub type Result<T> = ::std::result::Result<T, Error>;
+
+#[derive(Debug, Clone, PartialEq, PartialOrd)]
+#[repr(transparent)]
+pub(crate) struct LibCtruError(pub i32);
+
+impl Try for LibCtruError {
+    type Output = ();
+    type Residual = crate::Result<core::convert::Infallible>;
+
+    fn from_output(_: Self::Output) -> Self {
+        Self(0)
+    }
+
+    fn branch(self) -> ControlFlow<Self::Residual, Self::Output> {
+        match self.0 {
+            0 => ControlFlow::Continue(()),
+            _ => ControlFlow::Break(Err(self.into()))
+        }
+    }
+}
+
+impl FromResidual for LibCtruError {
+    fn from_residual(_: <Self as Try>::Residual) -> Self {
+        Self(1)
+    }
+}
 
 /// The error type returned by all libctru functions.
 #[non_exhaustive]
@@ -36,6 +63,12 @@ impl Error {
 impl From<ctru_sys::Result> for Error {
     fn from(err: ctru_sys::Result) -> Self {
         Error::Os(err)
+    }
+}
+
+impl From<LibCtruError> for Error {
+    fn from(err: LibCtruError) -> Self {
+        Self::Os(err.0)
     }
 }
 
--- a/ctru-rs/src/error.rs
+++ b/ctru-rs/src/error.rs
@@ -22,7 +22,7 @@ impl Try for LibCtruError {
     fn branch(self) -> ControlFlow<Self::Residual, Self::Output> {
         match self.0 {
             0 => ControlFlow::Continue(()),
-            _ => ControlFlow::Break(Err(self.into()))
+            _ => ControlFlow::Break(Err(self.into())),
         }
     }
 }
--- a/ctru-rs/src/error.rs
+++ b/ctru-rs/src/error.rs
@@ -13,6 +13,8 @@ pub(crate) struct LibCtruResult(pub i32);
 
 impl Try for LibCtruResult {
     type Output = ();
+    // This type is passed to [FromResidual::from_residual] when the LibCtruResult is an error,
+    // so this type implies "this is a result than CAN'T be `Ok`" (Infallible is the same as !)
     type Residual = crate::Result<core::convert::Infallible>;
 
     fn from_output(_: Self::Output) -> Self {
@@ -30,13 +32,9 @@ impl Try for LibCtruResult {
 
 impl FromResidual for LibCtruResult {
     fn from_residual(e: <Self as Try>::Residual) -> Self {
-        if let Some(e) = e.err() {
-            match e {
-                Error::Os(result) => Self(result),
-                _ => Self(-1),
-            }
-        } else {
-            Self(-1)
+        match e.err().unwrap() {
+            Error::Os(result) => Self(result),
+            _ => Self(-1),
         }
     }
 }
--- a/ctru-rs/src/error.rs
+++ b/ctru-rs/src/error.rs
@@ -9,7 +9,7 @@ pub type Result<T> = ::std::result::Result<T, Error>;
 
 #[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Eq, Ord)]
 #[repr(transparent)]
-pub(crate) struct ResultCode(pub i32);
+pub(crate) struct ResultCode(pub ctru_sys::Result);
 
 impl Try for ResultCode {
     type Output = ();
--- a/ctru-rs/src/error.rs
+++ b/ctru-rs/src/error.rs
@@ -20,9 +20,10 @@ impl Try for LibCtruError {
     }
 
     fn branch(self) -> ControlFlow<Self::Residual, Self::Output> {
-        match self.0 {
-            0 => ControlFlow::Continue(()),
-            _ => ControlFlow::Break(Err(self.into())),
+        if self.0 < 0 {
+            ControlFlow::Break(Err(self.into()))
+        } else {
+            ControlFlow::Continue(())
         }
     }
 }
--- a/ctru-rs/src/error.rs
+++ b/ctru-rs/src/error.rs
@@ -9,9 +9,9 @@ pub type Result<T> = ::std::result::Result<T, Error>;
 
 #[derive(Debug, Clone, PartialEq, PartialOrd)]
 #[repr(transparent)]
-pub(crate) struct LibCtruError(pub i32);
+pub(crate) struct LibCtruResult(pub i32);
 
-impl Try for LibCtruError {
+impl Try for LibCtruResult {
     type Output = ();
     type Residual = crate::Result<core::convert::Infallible>;
 
@@ -28,7 +28,7 @@ impl Try for LibCtruError {
     }
 }
 
-impl FromResidual for LibCtruError {
+impl FromResidual for LibCtruResult {
     fn from_residual(e: <Self as Try>::Residual) -> Self {
         if let Some(e) = e.err() {
             match e {
@@ -74,8 +74,8 @@ impl From<ctru_sys::Result> for Error {
     }
 }
 
-impl From<LibCtruError> for Error {
-    fn from(err: LibCtruError) -> Self {
+impl From<LibCtruResult> for Error {
+    fn from(err: LibCtruResult) -> Self {
         Self::Os(err.0)
     }
 }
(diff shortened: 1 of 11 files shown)


SteveCookTU:
Apologies, I would have removed the cam function in the PR but there were CI issues and I guess it went through accidentally. The function is used for when both cameras are taking a picture at the same time. Since this isn't implemented yet it can removed for now (whether a small addition to this PR or a small commit by an author to the main branch).

Techie-Pi:
> Since this isn't implemented yet it can removed for now (whether a small addition to this PR or a small commit by an author to the main branch).
Alright, I'll remove the function then. No problem

Techie-Pi:
The last commits were done to fix the CI, if this PR isn't merged I'll make them a separate PR

ctru-rs/src/error.rs
```diff
@@ -1,11 +1,46 @@
...
 
 pub type Result<T> = ::std::result::Result<T, Error>;
 
+#[derive(Debug, Clone, PartialEq, PartialOrd)]
+#[repr(transparent)]
+pub(crate) struct LibCtruResult(pub i32);
+
+impl Try for LibCtruResult {
+    type Output = ();
+    type Residual = crate::Result<core::convert::Infallible>;
```
Meziu, line 16: I got a little confused when I saw the `Residual` as `Infallible` (since its possible failure is quite literally the whole point), but this notation actually refers to a type which brings no information other than a possible error, which makes sense. I would still like a comment or two to explain why this is, just for readability.

ctru-rs/src/error.rs
```diff
@@ -1,11 +1,46 @@
...
+
+impl FromResidual for LibCtruResult {
+    fn from_residual(e: <Self as Try>::Residual) -> Self {
+        if let Some(e) = e.err() {
+            match e {
+                Error::Os(result) => Self(result),
+                _ => Self(-1),
+            }
+        } else {
+            Self(-1)
```
Meziu, line 39: This is a worst case scenario, right? In no normal case it should be an `Ok`, yet this is what it's implying. Maybe an `unreachable!` would be better? I didn't check if there are actual cases under which it is possible to run this line.

ctru-rs/src/lib.rs
```diff
@@ -2,6 +2,7 @@
 #![crate_name = "ctru"]
 #![feature(test)]
 #![feature(custom_test_frameworks)]
+#![feature(try_trait_v2)]
```
Meziu, line 5: I find the use of the new `Try` trait very elegant (especially compared to the use of a macro, which was my initially scrapped idea). Even so, I must say I don't really enjoy the use of even more unstable features, but this will probably be stabilized soon so it shouldn't be much of a problem.

ctru-rs/src/error.rs
```diff
@@ -1,11 +1,46 @@
...
 use std::fmt;
+use std::ops::{ControlFlow, FromResidual, Try};
 
 use ctru_sys::result::{R_DESCRIPTION, R_LEVEL, R_MODULE, R_SUMMARY};
 
 pub type Result<T> = ::std::result::Result<T, Error>;
 
+#[derive(Debug, Clone, PartialEq, PartialOrd)]
+#[repr(transparent)]
+pub(crate) struct LibCtruResult(pub i32);
```
Meziu, line 12: @ian-h-chamberlain @AzureMarker could you give your opinion on the name? I find `LibCtruResult` a bit confusing (since there is already a `Result` type), but I can't come up with a much better idea...

Meziu (commented):
Overall, this was a change I had already in mind for a while, but that you realized really nicely. I am positive to merging this, after receiving feedback from @ian-h-chamberlain and @AzureMarker .

ctru-rs/src/error.rs
Techie-Pi: Alright, I'll try my best to explain it

ctru-rs/src/error.rs
Techie-Pi: Maybe it would be a good idea to create a ``Result``-like enum, replace the current type-alias with that enum and implement: (i) the logic from ``LibCtruResult``, and (ii) a conversion method to Rust's ``Result``.

Though it feels like doing some gymnastics

ctru-rs/src/error.rs
Techie-Pi: Yeah, you're right, this shouldn't ever happen, I'll unwrap the error directly instead

ctru-rs/src/error.rs
Meziu: No, this new type is strictly internal to ease the raw `libctru` calls. It has no meaning externally. Maybe `RawResultWrapper` or something else would be better… I am not very creative 😅

Techie-Pi: I actually like that name, definitely better that the current one!

Techie-Pi: Should I change the current name, or should I wait for @ian-h-chamberlain and/or @AzureMarker opinions?

Meziu: You can change it already if you want, but this PR will stay on hold until I get some more opinions about it either way

AzureMarker: What do you think of `CtruSysResult` or `RawResult` instead? `RawResultWrapper` is a little long.

ctru-rs/src/error.rs
```diff
@@ -1,11 +1,45 @@
 use std::error;
 use std::ffi::CStr;
 use std::fmt;
+use std::ops::{ControlFlow, FromResidual, Try};
 
 use ctru_sys::result::{R_DESCRIPTION, R_LEVEL, R_MODULE, R_SUMMARY};
 
 pub type Result<T> = ::std::result::Result<T, Error>;
 
+#[derive(Debug, Clone, PartialEq, PartialOrd)]
```
AzureMarker: This should also derive `Copy`, `Eq`, and `Ord`.

ctru-rs/src/error.rs
Meziu: `CtruSysResult` can be mistaken for `ctru_sys::Result`. `RawResult` is the better option imo.

ctru-rs/src/error.rs
```diff
@@ -1,11 +1,45 @@
...
+#[repr(transparent)]
+pub(crate) struct LibCtruResult(pub i32);
+
+impl Try for LibCtruResult {
+    type Output = ();
+    // This type is passed to [FromResidual::from_residual] when the LibCtruResult is an error,
+    // so this type implies "this is a result than CAN'T be `Ok`" (Infallible is the same as !)
+    // The purpose of this type is to _only_ bring information about the *Error*
+    type Residual = crate::Result<core::convert::Infallible>;
```
AzureMarker: I think this implementation can be improved by looking at this example (which describes exactly what you're trying to do here!):
https://rust-lang.github.io/rfcs/3058-try-trait-v2.html#implementing-try-for-a-non-generic-type

ctru-rs/src/error.rs
```diff
@@ -1,11 +1,45 @@
...
+    }
+}
+
+impl FromResidual for LibCtruResult {
+    fn from_residual(e: <Self as Try>::Residual) -> Self {
+        match e.err().unwrap() {
+            Error::Os(result) => Self(result),
+            _ => Self(-1),
+        }
+    }
```
AzureMarker: Note: if we enable the [`exhaustive_patterns`](https://github.com/rust-lang/rust/issues/51085) feature this can be:
```suggestion
    fn from_residual(e: <Self as Try>::Residual) -> Self {
        match e {
            Err(Error::Os(result)) => Self(result),
            Err(_) => Self(-1),
        }
    }
```

ctru-rs/src/error.rs
AzureMarker: After reading through https://rust-lang.github.io/rfcs/3058-try-trait-v2.html#implementing-try-for-a-non-generic-type, `ResultCode` is another option.

ctru-rs/src/error.rs
Techie-Pi: Oh that's great! I'll check it out later!

ctru-rs/src/error.rs
Techie-Pi: I think ``ResultCode`` is the best option, to be honest

Meziu: > I think `ResultCode` is the best option, to be honest

Yes, I would go with that.

ctru-rs/src/error.rs
Techie-Pi: Mmm, I've not been able to create a working implementation with the example without making it more verbose or without having to use weird gymastics. I think maybe we should land this with the current impl (after I resolve the other concerns ofc), and make the implementation clearer in a follow-up PR

Edit: f896871b73f1b7d8dfa1e8c5c750637a4d913ca3

ctru-rs/src/error.rs
Techie-Pi: @Meziu What do you think about this?

Meziu: It is too small of a change to warrant the use an unstable feature. Obviously it isn’t a dangerous feature, but I’d like to keep things tidy.

ctru-rs/src/lib.rs
ian-h-chamberlain: I agree that we should be careful about adding new unstable features to depend on. Hopefully this one stabilizes before too long, but it would be nice to eventually build on stable.

It isn't written like this currently, but I think we could use `#[cfg_attr(test)]` for the other features here. `pthread-3ds` has some other unstable features too I guess, but it would be nice if we can reduce the scope to avoid becoming perma-unstable IMO.

ctru-rs/src/error.rs
ian-h-chamberlain: Can the residual type not just be `Error`, or would that end up requiring stuff like `?.into()` in the calling code? I am not super familiar with how the `Try` trait works so if that doesn't make any sense then I think this is probably fine as-is.

ian-h-chamberlain (approved):
Looks pretty good to me!

ctru-rs/src/error.rs
ian-h-chamberlain: Proof of concept, I think: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=4f40869939c1b94d22d877ebb6d1b1fa

Techie-Pi: Thank you @ian-h-chamberlain!

ctru-rs/src/error.rs
Techie-Pi: Changed to ``ResultCode`` (f896871b73f1b7d8dfa1e8c5c750637a4d913ca3)

Techie-Pi:
Hmmm CI fails to compile ``cargo-3ds`` with latest nightly

Meziu (approved):

ian-h-chamberlain:
> Hmmm CI fails to compile `cargo-3ds` with latest nightly

The CI on latest nightly failing shouldn't prevent merge, it's just there as a sort of canary to tell us whether stuff works on latest nightly. I think this instance is some kind of issue with the `CARGO_UNSTABLE_SPARSE_REGISTRY: "true"` flag we use, maybe it got destabilized or something.

ian-h-chamberlain (approved):

ctru-rs/src/error.rs
```diff
@@ -1,11 +1,48 @@
...
 use std::fmt;
+use std::ops::{ControlFlow, FromResidual, Try};
 
 use ctru_sys::result::{R_DESCRIPTION, R_LEVEL, R_MODULE, R_SUMMARY};
 
 pub type Result<T> = ::std::result::Result<T, Error>;
 
+#[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Eq, Ord)]
+#[repr(transparent)]
+pub(crate) struct ResultCode(pub i32);
```
AzureMarker: nit: use `ctru_sys::Result` instead of `i32`

AzureMarker (approved):

PR #64, opened 2022-07-10
Improve docs

Techie-Pi:
Following some suggestions from #32, I've made some changes to improve documentation.

- Added a simple Doxygen to Rustdoc script
- Applied some ``rustdoc`` suggestions
- Fixed doctests from ``Fs`` (mostly minor syntax issues)
- Removed ``--no-doc-comments`` from ``bindgen.sh`` to generate _kind of_ docs for users of ``ctru-sys``

Issue #32, opened 2022-01-29
General documentation improvements


ian-h-chamberlain:
Obviously, this library is still in early stages, and I don't expect there to be a ton of documentation at this point, but I think having good docs will be important for users and others looking to contribute.

A couple points I can think of right off the bat:
* Use [`rustdoc` lints](https://doc.rust-lang.org/rustdoc/lints.html?highlight=lint#lints) to check for missing docstrings and fix where relevant
* Investigate removing `--no-doc-comments` from `bindgen.sh`. There seem to be a pretty decent amount of doc comments in `libctru`, so having these readily available in Rust might be nice, even if they're written in a different format than `rustdoc` expects.
* Doctests? Unfortunately not much of a possibility for this library, although we might be able to use `no_run` and at least verify examples compile.
* Simply write docstrings for new code as it comes along. This alone will probably help a lot.

Long-term, if the crates are published to crates.io and then we can get nice docs on docs.rs! I have already been using `cargo 3ds doc` a bit to start out, with some success.

Edit: command to check doc lints looks like
```sh
cargo 3ds rustdoc -- -W rustdoc::all
```
Although that is a bit heavy handed – we may not want _all_ lints in the end.

Meziu:
Yeah, this is a good thing to keep in mind for the future of the crate.

Techie-Pi:
Maybe this is _somewhat_ off-topic, but I think it would be a good idea to generate and publish the Rustdocs of ``ctru-rs`` to GitHub Pages on push, that way is more accesible to users of the library. [Example of a workflow that does this](https://github.com/Techie-Pi/doxygen-rs/blob/main/.github/workflows/rustdoc.yml).

Meziu:
> Maybe this is _somewhat_ off-topic, but I think it would be a good idea to generate and publish the Rustdocs of `ctru-rs` to GitHub Pages on push, that way is more accesible to users of the library. [Example of a workflow that does this](https://github.com/Techie-Pi/doxygen-rs/blob/main/.github/workflows/rustdoc.yml).

I don't feel the need to do that. Once the crates get published (which should I hope will coincide with the merge of the thread-related features) [crates.io](https://www.crates.io) will handle the documentation hosting.

Techie-Pi:
> the crates get published (which I hope will coincide with the merge of the thread-related features)

In that case, you're right, that won't be needed

PeCoBe:
[...]

PR #75, opened 2022-10-09
Further improve ``bindings.rs`` documentation

Techie-Pi:
This adds a "_bin-dependency_" to ``ctru-sys`` to doxygen-rs, a project I've been working on to improve Doxygen to Rustdoc conversion with ``ctru-rs`` in mind: [Techie-Pi/doxygen-rs](https://github.com/Techie-Pi/doxygen-rs) .

Now, every single Doxygen tag used by the bindings is translated (except a few _markdown-like_ tags like ``p``).

Example of the new documentation (also check https://github.com/rust3ds/ctru-rs/pull/75#issuecomment-1276765784):
imagen

Maybe adding a dev-dependency is not desired, in that case I could send PRs only with the improvements to ``bindings.rs``, though it would not be easy to automate or to replicate on other machines.

Related: #32 #64 https://github.com/rust-lang/rust-bindgen/issues/955


Techie-Pi:
CI fails because the changes done in #74 have not been _back_-ported here

Techie-Pi:
This last commit adds support for the tags that remained untranslated. Further examples of tags that are now translated
> ### Return values
imagen

> ### Clickable references
imagen

> ### Clickable links
imagen

Techie-Pi:
(https://github.com/rust3ds/ctru-rs/pull/75#issuecomment-1272537447)

ctru-sys/src/bin/docstring-to-rustdoc.rs
```diff
@@ -7,17 +7,7 @@
 //!
```
AzureMarker: Update the usage to include the required feature

AzureMarker (approved):

ctru-sys/Cargo.toml
```diff
@@ -8,3 +8,11 @@ edition = "2021"
 
 [dependencies]
 libc = { version = "0.2.121", default-features = false }
+doxygen-rs = { git = "https://github.com/Techie-Pi/doxygen-rs.git", version = "0.2.2", rev = "573f483ad63ab4662650961998d3ed093a703983", optional = true }
```
ian-h-chamberlain: It might be preferable to split out `docstring-to-rustdoc` as a separate crate, just to make it very clear what is a "dev dependency" (tooling) vs a real dependency. This is the model used in e.g. [cargo-xtask](https://github.com/matklad/cargo-xtask), and I think I have seen in some other projects as well.

ian-h-chamberlain:
(Github won't let me comment directly on the file, it seems)

As a minor nit, the spacing for some of the `#[doc]` attributes seems a bit odd:
```rust
#[doc = "Undefined syscall."]
#[doc = ""]

pub const EXCEVENT_UNDEFINED_SYSCALL: ExceptionEventType = 8;
#[doc = "Reasons for an exception event."]
#[doc = ""]

pub type ExceptionEventType = ::libc::c_uint;
```

I would expect more like
```rs
#[doc = "Undefined syscall."]
#[doc = ""]
pub const EXCEVENT_UNDEFINED_SYSCALL: ExceptionEventType = 8;
// maybe no newline here, idk
#[doc = "Reasons for an exception event."]
#[doc = ""]
pub type ExceptionEventType = ::libc::c_uint;
```

I don't imagine people will read `bindings.rs` directly very often, so probably not a big deal in general.

ian-h-chamberlain (approved):
Looks good! I am trying to fix my local toolchain installation to build the docs locally and check them out but your screenshots look like a nice improvement over what was there.

Techie-Pi:
> Looks good! I am trying to fix my local toolchain installation to build the docs locally and check them out but your screenshots look like a nice improvement over what was there.

Try deleting the Cargo.lock file on root, that did it for me

ian-h-chamberlain:
> Try deleting the Cargo.lock file on root, that did it for me

Thanks, it was more of an OS-level issue but I got it solved and the docs look good!

Actually on a semi-related note... I wonder if we ought to check in a Cargo.lock for the sake of CI, and maybe add another CI job to the matrix that deletes Cargo.lock before running just to find breakage from upstream. Just a thought

I do see some weird rendering for docstrings on constants, but I'm guessing that's a rustdoc bug and nothing to do with your changes:

<details>
<summary>example screenshot</summary>

Screen Shot 2022-10-15 at 2 26 22 PM


</details>

Clicking into them seems to work fine so idk what's up with that.

Techie-Pi:
> I do see some weird rendering for docstrings on constants, but I'm guessing that's a rustdoc bug and nothing to do with your changes:

Yeah, that's an issue with Rustdoc, it also happened before #64

ctru-sys/Cargo.toml
Techie-Pi: I think this is what you meant(?): 95756f7b73cd82994a3257acf31a179b1e1af00b

Meziu (approved):
Looks great! I would still refer to the original docs, but it doesn't hurt to have them all here as well.

ctru-sys/bindgen.sh
```diff
@@ -29,6 +29,6 @@ bindgen "$DEVKITPRO/libctru/include/3ds.h" \
     -D__3DS__ \
 > src/bindings.rs
 
-cargo run --bin docstring-to-rustdoc --features="build-binary" -- src/bindings.rs
+cargo docstring-to-rustdoc src/bindings.rs
```
ian-h-chamberlain: Looks good to me, but this might need to be something like 
```sh
cargo run --package docstring-to-rustdoc -- src/bindings.rs
```

Unless you actually install the binary at some point

Edit: fixed the command to one that actually works

ian-h-chamberlain (approved):

ctru-sys/bindgen.sh
Techie-Pi: I tested it, and because of the alias added in the ``.cargo/config.toml`` it works without any problem

ian-h-chamberlain: Ah, I see. That file isn't checked in as part of the repository due to this `.gitignore` entry: https://github.com/rust3ds/ctru-rs/blob/master/.gitignore#L3

So I don't think it will work for anyone else unless the command is changed or the `config.toml` gets checked in (not my preference personally, since I tend to use the config file for random stuff like patches, etc.)

Techie-Pi: Ahhhhh I didn't realize that! I'll change it then

Techie-Pi: Done: dff7ee19c6ef04fd868e0c85585c242550260454

ian-h-chamberlain (approved):

Techie-Pi:
Wait a second, I think I need to check one thing

Techie-Pi:
Alright, done. I thought I had committed something I shouldn't, but that isn't the case

PR #76, opened 2022-10-15
First attempt at splitting top screen types

### LINKED ISSUE #69: Consider splitting up top screen into two types?
As I've been trying to work on https://github.com/ian-h-chamberlain/citro3d-rs/pull/4, it occurred to me that it might make sense for us to break out the top-left and top-right screen into their own structs. Currently, it's a bit challenging to write code that's generic for both screens, since you can't get the raw frame buffer, so you have to write something like:

```rust
impl Thing {
    pub fn new_top(screen: TopScreen, side: Side);
    pub fn new_bottom(screen: BottomScreen);
}
```

I was thinking, maybe we could separate the top screen into two types, and add frame buffer helpers to `Screen` instead, like:
```rust
pub trait Screen {
    /// Returns the libctru value for the Screen kind
    fn as_raw(&self) -> ctru_sys::gfxScreen_t;

    /// Returns the raw frame buffer for rendering to the Screen
    fn get_raw_framebuffer(&mut self) -> RawFrameBuffer;
}
```

This way, it would be a bit easier to have separate ownership of the top-left and top-right screens, in case e.g. you wanted to have separate render targets:
```rust
let top_left = render::Target::new(gfx.top_left_screen.borrow_mut());
let top_right = render::Target::new(gfx.top_right_screen.borrow_mut());
```

For simplicity, maybe we could also have a helper struct/method in case you want to render to both sides, although I don't think it could implement `Screen` so it may have limited uses. Perhaps a helper like that would be more useful in render code than in `gfx` itself.

I think this implementation would be more consistent with e.g. the CAM implementation in #65 and would make it a bit easier to write generic code against screens. @Meziu @AzureMarker what are your thoughts?

ian-h-chamberlain:
Closes #69

I'm not totally happy with this API, it feels like there's a bit of repetition but I think this is generally the gist of what I was thinking. I haven't updated all the examples yet since I wanted to get some opinions first.

I don't love having the extra inner type, but just using a `RefCell<TopScreen>` would force callers to use `Ref{Mut}::map` which is kind of a pain and also requires exposing the `left` and `right` members, so idk.

One thing that's kinda missing here is a type-safe way of working with a wide mode screen, for example it might be nice to have something like this:

```rs
struct TopWideScreen;
impl Screen for TopWideScreen {
	// use left-side frame buffer
}

// maybe no impl Screen for TopScreen, then?

impl TopScreen {
	pub fn wide_mode(self) -> TopWideScreen;
}

impl TopWideScreen {
	pub fn unset_wide_mode(self) -> TopScreen; // or some other way to disable wide mode
}
```


Meziu:
This is a big issue, especially when keeping in mind `citro-3d`. The idea of splitting the types sounds fine, but we must check for all safety measures. For example: When enabling wide-mode, both `TopScreen` sides should be available. Also, does rendering to the right side work when not in 3d-mode? Can we actually use all of the screens (left and right side counted as different) independently?

AzureMarker:
Is it legal/safe to enable/disable wide mode while already using the top screen instances? I think it works, so not sure if we should block it.

Meziu:
> Is it legal/safe to enable/disable wide mode while already using the top screen instances? I think it works, so not sure if we should block it.

I believe the RightSide would become invalid, or just do nothing. I haven’t tested it.

Meziu:
After looking at `libctru`'s code, I can firmly declare that: the Right side is a lie. There is no such thing as a Right side. The framebuffer size is *always* 2x the TopScreen pixelcount (depending on the format). When getting the RightSide's framebuffer, it only returns the LeftSide's framebuffer's (which we might as well call the "TopScreen framebuffer") pointer with an offset. The additional RAM is allocated even if the program never needs 3D or WIDE mode.

In my opinion, it's a bit wacky, but it also means that the RightSide is valid as long as the `gfx` service is running.
Furthermore, the RightSide isn't flushed to the screen *unless* 3D mode is activated (or its memory is used by wide mode): it cannot be used as a standalone screen.

At this point, I believe the best approach would be to remove the "get framebuffer based on side" thing and just split the types when asked by 3D mode activation. This way we can make safety checks for the use of `wide-mode` (or even just the left side alone) directly, and not worry too much about it.

Something like:
```Rust
impl TopScreen {
        // We get the 2 sides when needed for 3D
	pub fn enable_3d_mode(&mut self) -> (TopLeft, TopRight);
	
	// The 2 sides get dropped inside this function to assure the correct deactivation of 3D mode
	pub fn disable_3d_mode(&mut self, TopLeft: Side, TopRight: Side);
}
```

@ian-h-chamberlain @AzureMarker What `Side` (as in the trait/type) should be in this case is up to debate. It shouldn't be a "full" version of the screen, since the sides can't independently change framebuffer format or set double buffering, which would be instead handled by the already existing `TopScreen` instance. Tell me your ideas to implement this.

AzureMarker:
Those functions would need to take ownership of the TopScreen value, otherwise you could make multiple left and right values. It could return a third value representing the top screen in general.

ian-h-chamberlain:
Hey all, I've just pushed a change based on the discussion, which I think mostly accomplishes the goals of having a distinct type to split the top screen. It's not quite what I first envisioned, but keeps the `RefCell` somewhat like it was before.

I also added an example to make sure it works as expected when writing to the left/right frame buffers and it seems like it works well! Proper Citro3D stereoscopic might work a bit differently, but I think could end up taking ownership of a `RefMut<dyn Screen>` to prevent writing directly to the framebuffer while C3D is rendering to the screen.

ctru-rs/examples/gfx-3d-mode.rs
```diff
@@ -0,0 +1,72 @@
+use ctru::gfx::TopScreen3D;
+use ctru::prelude::*;
+
+/// See `graphics-bitmap.rs` for details on how the image is generated.
+///
+/// WARNING: this example uses 3D mode in a rather unnatural way, and should
+/// probably not be viewed for too long or at all if you are photosensitive.
+
+const IMAGE: &[u8] = include_bytes!("assets/ferris.rgb");
+const ZERO: &[u8] = &[0; IMAGE.len()];
```
AzureMarker: These should probably be `static` not `const`.

ctru-rs/src/gfx.rs
```diff
@@ -36,6 +54,21 @@ pub trait Screen {
...
+    // - RefMut: we don't have an easy way to obtain Ref<dyn Screen> for the left side.
+    //      maybe this one isn't as important since the use case is typically RefMut anyway.
+    //      we could just return &dyn Screen instead of Ref<dyn Screen> ?
+    left: &'a RefCell<TopScreen>,
+    right: RefCell<TopScreenRight>,
+}
+
+// TODO: it feels a little weird to have an asymmetric separate type like this,
+// but maybe if it's not `pub` it's not as weird...
+struct TopScreenRight;
```
AzureMarker, line 85: I think this is OK, since we just hide it behind `dyn Screen`. Alternatively I guess `TopScreen` could hold a `TopScreenLeft` and `TopScreenRight`, and then `TopScreen3D` would have references to each.

ctru-rs/examples/gfx-3d-mode.rs
ian-h-chamberlain: I could do that for `ZERO`, but `IMAGE` has to be `const` in order to use `IMAGE.len()` so I thought it would just be consistent that way. I can change it though

ian-h-chamberlain:
@rust3ds/active anyone have objections to merging this, or other comments?

ctru-rs/src/gfx.rs
```diff
@@ -34,9 +59,26 @@ pub trait Screen {
 }
 
 #[non_exhaustive]
```
AzureMarker: nit: this isn't needed now that there are private fields

ctru-rs/src/gfx.rs
```diff
@@ -139,80 +182,128 @@ impl Gfx {
     }
 }
 
-impl TopScreen {
-    /// Enable or disable the 3D stereoscopic effect
-    pub fn set_3d_enabled(&mut self, enabled: bool) {
+impl<'screen> RawFrameBuffer<'screen> {
+    fn for_screen_side(screen: &'screen mut (impl Screen + ?Sized), side: Side) -> Self {
```
AzureMarker: Why not `&'screen mut dyn Screen`?

ctru-rs/examples/gfx-3d-mode.rs
```diff
@@ -0,0 +1,72 @@
...
+    let gfx = Gfx::init().expect("Couldn't obtain GFX controller");
+    let hid = Hid::init().expect("Couldn't obtain HID controller");
+    let apt = Apt::init().expect("Couldn't obtain APT controller");
+    let _console = Console::init(gfx.bottom_screen.borrow_mut());
+
+    println!("Press Start to exit.\nPress A to switch which side is drawn to.");
+
+    let top_screen = TopScreen3D::from(&gfx.top_screen);
+
+    // TODO set double buffering for top screen
```
AzureMarker: Reminder to handle this TODO

ctru-rs/src/gfx.rs
ian-h-chamberlain: Actually ended up just inlining this function definition into the one place it was used, avoiding the `?Sized` issues that cropped up otherwise

ian-h-chamberlain: Actually ended up just inlining this function definition into the one place it was used, avoiding the `?Sized` issues that cropped up otherwise

AzureMarker (approved):

Techie-Pi (approved):

Meziu (approved):

PR #78, opened 2022-10-20
Implement Mii Selector applet

Techie-Pi:
I've implemented the Mii Selector applet, I have yet to add documentation, but I would like to see your opinions on the API.
Specially around the creation and usage of the enums ``MiiConfigIndex`` and ``MiiType``.


Techie-Pi:
Also, out of curiosity, why is [``MiiData`` an opaque type](https://github.com/rust3ds/ctru-rs/blob/9cf5e4de94e8d3711fd757e3464dd1880d91b92f/ctru-sys/bindgen.sh#L17)?
Nevermind, just saw the [issue](https://github.com/rust-lang/rust-bindgen/issues/1556), that's actually sad tbh

Techie-Pi:
I've been researching, and the representation of the MiiData given by the 3DS doesn't change at all and [is documented](https://www.3dbrew.org/wiki/Mii#Mii_format). The purpose of libctru's ``MiiData`` is to _imitate_ the representation so that the struct is generated directly from the 3DS' MiiData.

Instead of that, we could parse the raw data to a Rust-compatible struct. I'll do some testing

Techie-Pi:
Looks like it is possible, and quite simple tbh, maybe I could also add this to the PR.
Edit: Done... that was easier than what I had thought... I'll clean-up the code, fix a few things, and I'll incorporate the code into the PR

ctru-rs/src/mii.rs
```diff
@@ -0,0 +1,435 @@
...
+
+#[derive(Clone, Debug)]
+pub struct SelectorPosition {
+    pub page_index: u8,
+    pub slot_index: u8,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum OriginConsole {
+    ConsoleWii,
```
Meziu: Remove the `Console` prefix from all variants.

Techie-Pi: But Rust doesn't allow numerals at the start of a member (3DS)

ctru-rs/src/mii.rs
```diff
@@ -0,0 +1,435 @@
...
+    pub mouth_details: MouthDetails,
+    pub mustache_details: MustacheDetails,
+    pub beard_details: BeardDetails,
+    pub glass_details: GlassDetails,
+    pub mole_details: MoleDetails,
+
+    pub author_name: String,
+}
+
+impl From<[u8; 92]> for MiiData {
```
Meziu: Use the `ctru-sys` type rather than `[u8;92]`. Do this for all the cases in which it appears.

ctru-rs/src/mii.rs
Meziu: True, hadn't noticed... Still, does "WiiUSwitch" actually exist as an option? I don't see it mentioned in `libctru`'s docs.

ctru-rs/src/mii.rs
```diff
@@ -0,0 +1,479 @@
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum RegionLock {
+    NoLock,
```
Meziu: `None` would be less repetitive.

ctru-rs/src/mii.rs
Techie-Pi: The 3dbrew wiki says so.
[Ref](https://www.3dbrew.org/wiki/Mii#Mii_format) offset 0x3, bit 4-6

ctru-rs/src/mii.rs
Techie-Pi: The problem is that [there is no ``ctru-sys`` type for ``[u8; 92]``](https://github.com/rust3ds/ctru-rs/blob/9cf5e4de94e8d3711fd757e3464dd1880d91b92f/ctru-sys/src/bindings.rs#L17597), but I could make a type alias

Meziu:
My system's main mii is grayed out when selecting and thus can't be used. Is there an issue with the configuration?

ctru-rs/examples/mii-selector.rs
```diff
@@ -0,0 +1,45 @@
...
+    ctru::init();
+
+    let gfx = Gfx::init().expect("Couldn't obtain GFX controller");
+    let hid = Hid::init().expect("Couldn't obtain HID controller");
+    let apt = Apt::init().expect("Couldn't obtain APT controller");
+    let _console = Console::init(gfx.top_screen.borrow_mut());
+
+    let mut mii_selector = MiiSelector::init();
+    mii_selector.set_initial_index(3);
+    mii_selector.blacklist_user_mii(0.into());
```
Meziu, line 14: Why blacklist the user's mii?

ctru-rs/examples/mii-selector.rs
```diff
@@ -0,0 +1,45 @@
...
+    let result = mii_selector.launch();
+
+    println!("\x1b[0;0HIs Mii selected?: {:?}", result.is_mii_selected);
+    println!("\x1b[1;0HValid checksum?: {:?}", result.valid_checksum());
+    println!("\x1b[2;0HMii type: {:?}", result.mii_type);
+    println!("\x1b[3;0HMii checksum: {:?}", result.checksum);
+    println!("\x1b[4;0HName: {:?}", result.name());
+    println!("\x1b[5;0HAuthor: {:?}", result.author());
+    println!(
+        "\x1b[5;0HDoes it have moles?: {:?}",
```
Meziu: This message and the "author" message are both on row 5. In this order the "Moles" message wipes the "Author" one.

ctru-rs/examples/mii-selector.rs
Techie-Pi: That's just an example, [the 3ds-examples' example also does this](https://github.com/devkitPro/3ds-examples/blob/cde9dff910f1115de79752cf55c609a9616cdc94/mii_selector/source/main.c#L54)

ctru-rs/src/applets/mii_selector.rs
```diff
@@ -0,0 +1,193 @@
...
+        const MII_SELECTOR_GUEST_START = ctru_sys::MIISELECTOR_GUESTSTART;
+    }
+}
+
+pub struct MiiSelector {
+    config: Box<ctru_sys::MiiSelectorConf>,
+}
+
+pub struct MiiSelectorReturn {
+    raw_return: Box<ctru_sys::MiiSelectorReturn>,
```
Meziu: We shouldn't hold this pointer. It has no use after being consumed by our wrapper.

ctru-rs/src/applets/mii_selector.rs
```diff
@@ -0,0 +1,193 @@
...
+
+    pub fn launch(&mut self) -> MiiSelectorReturn {
+        let mut return_val = Box::<ctru_sys::MiiSelectorReturn>::default();
+        unsafe { ctru_sys::miiSelectorLaunch(self.config.as_mut(), return_val.as_mut()) }
+
+        return_val.into()
+    }
+}
+
+impl MiiSelectorReturn {
```
Meziu: These functions are duplicate features already present (name and author can be obtained by the `MiiData`).

Furthermore, the checksum (while useful in a C environment) is a case we just shouldn't expect the user to check. The case of a faulty checksum means that our `MiiSelectorReturn` wrapped value is invalid, which is one of the safety measures we should prevent *by default*.

The checksum should be checked before returning the mii data to the user, and if the check fails, it should return an error. This check should be handled in the `MiiSelector::launch` method.

ctru-rs/src/applets/mii_selector.rs
```diff
@@ -0,0 +1,193 @@
...
+    config: Box<ctru_sys::MiiSelectorConf>,
+}
+
+pub struct MiiSelectorReturn {
+    raw_return: Box<ctru_sys::MiiSelectorReturn>,
+    pub mii_data: MiiData,
+    pub is_mii_selected: bool,
+    pub mii_type: MiiType,
+    pub checksum: u16,
+}
```
Meziu, line 67: All of the values in this struct should be `read-only`, since this is a *result* type for our operations. Rather than being `pub`, they should each have a corresponding `get` method.

Also, as stated in another comment, the presence of `checksum` here is to be discussed (as its validity would be already checked), thought I'd like to hear reasons why it may actually be useful, since I've never played around with this part of `libctru`.

ctru-rs/src/mii.rs
Meziu: You should just use `ctru_sys::MiiData` for the conversion and take the inner buffer *here* rather than in `MiiSelectorResult`.

ctru-rs/src/mii.rs
```diff
@@ -0,0 +1,479 @@
...
+pub struct MoleDetails {
+    pub is_enabled: bool,
+    pub scale: u8,
+    pub x_position: u8,
+    pub y_position: u8,
+}
+
+#[derive(Clone, Debug)]
+pub struct MiiData {
+    pub mii_options: MiiDataOptions,
```
Meziu: I think you should remove all the `mii_` prefixes in the struct's members. Most of the time they will be called directly from the `mii_data.[member]` signature.

ctru-rs/src/mii.rs
AzureMarker: You can use `N3DS` instead

ctru-rs/examples/mii-selector.rs
```diff
@@ -0,0 +1,45 @@
...
+    let _console = Console::init(gfx.top_screen.borrow_mut());
+
+    let mut mii_selector = MiiSelector::init();
+    mii_selector.set_initial_index(3);
+    mii_selector.blacklist_user_mii(0.into());
+    mii_selector.set_title("Great Mii Selector!");
+
+    let result = mii_selector.launch();
+
+    println!("\x1b[0;0HIs Mii selected?: {:?}", result.is_mii_selected);
```
AzureMarker: I don't think the line positioning code is needed. The console automatically handles newlines.

Edit: Running this on my 3DS, the second line overwrites the first line until you remove the line positioning code.

Meziu (changes requested):

ctru-rs/src/applets/mii_selector.rs
```diff
@@ -0,0 +1,193 @@
...
+        unsafe {
+            ctru_sys::miiSelectorInit(config.as_mut());
+        }
+        Self { config }
+    }
+
+    pub fn set_title(&mut self, text: &str) {
+        // This can only fail if the text contains NUL bytes in the string... which seems
+        // unlikely and is documented
+        let c_text = CString::new(text).expect("Cstring::new failed");
```
AzureMarker: nit: based on the comment the `expect` error message could be more accurate/useful.

ctru-rs/src/applets/mii_selector.rs
```diff
@@ -0,0 +1,193 @@
...
+            MiiConfigIndex::Index(i) => unsafe {
+                ctru_sys::miiSelectorWhitelistGuestMii(self.config.as_mut(), i)
+            },
+            MiiConfigIndex::All => unsafe {
+                ctru_sys::miiSelectorWhitelistGuestMii(
+                    self.config.as_mut(),
+                    ctru_sys::MIISELECTOR_GUESTMII_SLOTS,
+                )
+            },
+        }
```
AzureMarker: nit: Could reduce duplication (this applies to multiple methods):
```suggestion
        let index = match mii_index {
            MiiConfigIndex::Index(i) => i,
            MiiConfigIndex::All => ctru_sys::MIISELECTOR_GUESTMII_SLOTS,
        };
        unsafe { ctru_sys::miiSelectorWhitelistGuestMii(self.config.as_mut(), index) };
```

Even better, move the `match` conversion to a method on `MiiConfigIndex`.

ctru-rs/src/applets/mii_selector.rs
AzureMarker: Why include this in the struct if it's private? This struct will only be used by users, who can't access this field.

ctru-rs/src/mii.rs
```diff
@@ -0,0 +1,479 @@
...
+        .chunks(2)
+        .collect::<Vec<&[u8]>>()
+        .iter()
+        .map(|v| u16::from_le_bytes([*v.first().unwrap_or(&0), *v.get(1).unwrap_or(&0)]))
+        .collect::<Vec<u16>>();
+
+    String::from_utf16(raw_utf16_composed.as_slice())
+        .unwrap()
+        .replace('\0', "")
+}
```
AzureMarker: nit: This could be simpler:
```suggestion
fn utf16_byte_pairs_to_string(data: &[u8]) -> String {
    let raw_utf16_composed = data
        .chunks_exact(2)
        .map(|chunk| u16::from_le_bytes([chunk[0], chunk[1]]))
        .collect::<Vec<u16>>();

    String::from_utf16_lossy(raw_utf16_composed.as_slice())
}
```

ctru-rs/src/mii.rs
```diff
@@ -0,0 +1,479 @@
...
+
+/// Gets the values from the slice and concatenates them
+fn get_and_concat_vec_bit(data: &[u8], get_values: &[usize]) -> Vec<bool> {
+    get_values
+        .iter()
+        .map(|v| vec_bit(data[*v]))
+        .collect::<Vec<[bool; 8]>>()
+        .concat()
+}
```
AzureMarker: nit: This could be simpler:
```suggestion
fn get_and_concat_vec_bit(data: &[u8], get_values: &[usize]) -> Vec<bool> {
    get_values
        .iter()
        .flat_map(|v| vec_bit(data[*v]))
        .collect()
}
```

ctru-rs/src/mii.rs
```diff
@@ -0,0 +1,479 @@
...
+            raw_mii_data[0x12],
+            raw_mii_data[0x13],
+            raw_mii_data[0x14],
+            raw_mii_data[0x15],
+        ];
+        let raw_details: [bool; 16] =
+            get_and_concat_vec_bit(raw_mii_data.as_slice(), &[0x18, 0x19])
+                .try_into()
+                .unwrap();
+        let raw_utf16_name = &raw_mii_data.as_slice()[0x1A..0x2D];
```
AzureMarker: This can just be `let raw_utf16_name = &raw_mii_data[0x1A..0x2D];` (`as_slice` isn't needed)

ctru-rs/src/mii.rs
```diff
@@ -0,0 +1,479 @@
...
+        let creator_mac = [
+            raw_mii_data[0x10],
+            raw_mii_data[0x11],
+            raw_mii_data[0x12],
+            raw_mii_data[0x13],
+            raw_mii_data[0x14],
+            raw_mii_data[0x15],
+        ];
+        let raw_details: [bool; 16] =
+            get_and_concat_vec_bit(raw_mii_data.as_slice(), &[0x18, 0x19])
```
AzureMarker: nit: it's more idiomatic to do `&raw_mii_data` instead of `raw_mii_data.as_slice()`

ctru-rs/src/mii.rs
```diff
@@ -0,0 +1,479 @@
...
+                .try_into()
+                .unwrap();
+        let raw_utf16_author = &raw_mii_data.as_slice()[0x48..0x5C];
+
+        let mii_name = utf16_byte_pairs_to_string(raw_utf16_name);
+        let author_name = utf16_byte_pairs_to_string(raw_utf16_author);
+
+        let options = MiiDataOptions {
+            is_copying_allowed: *raw_options.first().unwrap(),
+            is_profanity_flag_enabled: *raw_options.get(1).unwrap(),
```
AzureMarker:
```suggestion
            is_copying_allowed: raw_options[0],
            is_profanity_flag_enabled: raw_options[1],
```

ctru-rs/src/mii.rs
```diff
@@ -0,0 +1,479 @@
...
+                    RegionLock::None
+                } else if !first_bit && second_bit {
+                    // 0b01
+                    RegionLock::Japan
+                } else if first_bit && !second_bit {
+                    // 0b10
+                    RegionLock::USA
+                } else {
+                    RegionLock::Europe
+                }
```
AzureMarker: Do this refactoring for the other similar instances below as well.
```suggestion
                match (raw_options[3], raw_options[2]) {
                    (false, false) => RegionLock::None,
                    (false, true) => RegionLock::Japan,
                    (true, false) => RegionLock::USA,
                    (true, true) => RegionLock::Europe,
                }
```

ctru-rs/src/mii.rs
```diff
@@ -0,0 +1,479 @@
...
+        };
+
+        let details = Details {
+            sex: {
+                let first_bit = raw_details[0];
+                if first_bit {
+                    MiiSex::Female
+                } else {
+                    MiiSex::Male
+                }
```
AzureMarker:
```suggestion
                if raw_details[0] {
                    MiiSex::Female
                } else {
                    MiiSex::Male
                }
```

ctru-rs/src/mii.rs
```diff
@@ -0,0 +1,479 @@
...
+    let mut result: u8 = 0;
+    data.map(u8::from).iter().for_each(|&bit| {
+        result <<= 1;
+        result ^= bit;
+    });
+    result
+}
+
+/// The reverse allows to write things on a more _humane_ way, but return a LE u8
+fn partial_vec_to_u8_with_reverse(data: &[bool]) -> u8 {
```
AzureMarker: This function's functionality is still unclear to me, maybe a better name and more descriptive docs would help?

ctru-rs/src/mii.rs
```diff
@@ -0,0 +1,479 @@
...
+
+/// Transforms a [bool; 8] into an u8
+fn vec_bit_to_u8(data: [bool; 8]) -> u8 {
+    let mut result: u8 = 0;
+    data.map(u8::from).iter().for_each(|&bit| {
+        result <<= 1;
+        result ^= bit;
+    });
+    result
+}
```
AzureMarker: nit: This could be simplified:
```suggestion
fn vec_bit_to_u8(data: [bool; 8]) -> u8 {
    data.into_iter().fold(0, |result, bit| {
        (result << 1) ^ u8::from(bit)
    })
}
```

ctru-rs/src/mii.rs
```diff
@@ -0,0 +1,479 @@
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum RegionLock {
```
AzureMarker, line 8: It makes sense to derive `Copy` for most of these structs/enums.

ctru-rs/src/applets/mii_selector.rs
AzureMarker: I think it's fine to return the whole struct and let the user mess with it. It's not the library's problem after it gets returned to the user. They can muck around with it all they want, we don't need it back :).

ctru-rs/src/applets/mii_selector.rs
Techie-Pi: Oh yeah, that's something that I left from testing, I'll remove it
Edit: Actually, it was used. But I've removed it

ctru-rs/src/mii.rs
Techie-Pi: Ohhh that's actually great! I hadn't thought of doing that

ctru-rs/src/applets/mii_selector.rs
Techie-Pi: > Even better, move the match conversion to a method on MiiConfigIndex.

The problem with that is that, depending on the context, ``MiiConfigIndex::All`` returns ``MIISELECTOR_GUESTMII_SLOTS`` or ``MIISELECTOR_USERMII_SLOTS``

ctru-rs/src/applets/mii_selector.rs
Techie-Pi: Is it better now?

ctru-rs/src/mii.rs
Techie-Pi: Maybe I need help writing the docs for this function, so I'll write what's the purpose of this functions here so you'll be able to give some suggestions.
The 3DS is a LE system, but for humans it's _easier_ to read and understand BE bits. So, this function is given a BE sequence of bits (e.g. ``[true, true, false]``) that gets: (i) filled with _``false``s_ to fill the remaining space until 8 bits (``u8``) are filled (e.g. ``[false, false, false, false, false, true, true, false]``), and (ii) the bits returned are transformed to be LE (e.g. ``[false, true, true, false, false, false, false, false]`` so the 3DS can understand them.

``partial_vec`` -> A Vec\<bool> (_in reality it is a slice_) that doesn't fill the 8 bits of an ``u8``
``with_reverse`` -> Where the values being returned are LE (""_reversed_"")

Maybe something like ``partial_bits_to_le_u8`` is better?

ctru-rs/src/mii.rs
Techie-Pi: > You can use N3DS instead

Just to clarify (_sorry it is late here_), ``N3DS`` instead of ``WiiUSwitch`` right?

ctru-rs/src/applets/mii_selector.rs
Techie-Pi: I agree with AzureMaker, and the ``checksum`` thing has been solved

ctru-rs/src/mii.rs
Meziu: No, write `N3DS` instead of `Console3DS`. At that point you can remove all the `Console` prefixes.

Techie-Pi: Alright, that makes sense, idk why I thought that it was meant to replace ``WiiUSwitch``.
About replacing ``Console3DS`` with ``N3DS``, wouldn't people confuse ``N3DS`` with New 3DS?

ctru-rs/src/mii.rs
Techie-Pi: Alright, that function was _actually_ misused. The purpose of the function was writing BE bits, filling them until the length was 8 bits and making them LE.

This function was used this way before, but now the data is directly from the 3DS and this isn't needed + it caused some _big big_ bugs. I've fixed it and documented it

Techie-Pi:
(For some reason I can't request multiple reviews at the same time /shrug)

ctru-rs/src/mii.rs
AzureMarker: We can use `New3DS` or `NewN3DS` for that. `N3DS` could mean "Nintendo 3DS"

AzureMarker: Or use `O3DS` for "Original 3DS", but I don't think that's used often.

I named one of my repos [`n3ds-controller`](https://github.com/AzureMarker/n3ds-controller/) because of this issue.

ctru-rs/src/applets/mii_selector.rs
AzureMarker: In that case just doing the refactor in each method would be good.

ctru-rs/src/applets/mii_selector.rs
```diff
@@ -0,0 +1,190 @@
...
+    config: Box<ctru_sys::MiiSelectorConf>,
+}
+
+/// Return value from a MiiSelector's launch
+#[derive(Clone, Debug)]
+pub struct MiiSelectorReturn {
+    pub mii_data: MiiData,
+    pub is_mii_selected: bool,
+    pub mii_type: MiiType,
+    pub checksum: u16,
```
AzureMarker: Is the checksum relevant now that the original data structure isn't returned?

ctru-rs/src/applets/mii_selector.rs
```diff
@@ -0,0 +1,190 @@
...
+
+        if unsafe { ctru_sys::miiSelectorChecksumIsValid(return_val.as_mut()) } {
+            Ok(return_val.into())
+        } else {
+            Err(MiiLaunchError::InvalidChecksum)
+        }
+    }
+}
+
+impl From<Box<ctru_sys::MiiSelectorReturn>> for MiiSelectorReturn {
```
AzureMarker: nit: this should be `From<ctru_sys::MiiSelectorReturn>`. The `Box` part isn't needed/utilized.

ctru-rs/src/mii.rs
```diff
@@ -0,0 +1,461 @@
...
+            birthday_month: partial_u8_bits_to_u8(&raw_details[1..4]),
+            birthday_day: partial_u8_bits_to_u8(&raw_details[5..9]),
+            shirt_color: partial_u8_bits_to_u8(&raw_details[10..13]),
+            is_favorite: raw_details[14],
+        };
+
+        let face_details = FaceDetails {
+            style: FaceStyle {
+                is_sharing_enabled: !raw_face_style[1],
+                shape: partial_u8_bits_to_u8(&raw_face_style[1..4]),
```
AzureMarker: Are you sure `is_sharing_enabled` uses `raw_face_style[1]`? `shape` also uses that value.

ctru-rs/src/mii.rs
```diff
@@ -0,0 +1,461 @@
...
+        let details = Details {
+            sex: {
+                match raw_details[0] {
+                    true => MiiSex::Female,
+                    false => MiiSex::Male,
+                }
+            },
+            birthday_month: partial_u8_bits_to_u8(&raw_details[1..4]),
+            birthday_day: partial_u8_bits_to_u8(&raw_details[5..9]),
+            shirt_color: partial_u8_bits_to_u8(&raw_details[10..13]),
```
AzureMarker: I see some missed bits here. `1..4` is `1, 2, 3`, and `5..9` is `5, 6, 7, 8` (4 is missing).
This shows up a lot, is it expected?

Techie-Pi: Yeah this is expected, some bits are completely unused/padding according to 3dbrew

ctru-rs/src/applets/mii_selector.rs
Techie-Pi: Oh yeah, I forgot to remove that

ctru-rs/src/mii.rs
Techie-Pi: I think I'll use N3DS and clarify this in the documentation, seems like the best option to me

ctru-rs/src/applets/mii_selector.rs
```diff
@@ -0,0 +1,190 @@
...
+/// let result = mii_selector.launch().unwrap();
+/// ```
+#[derive(Clone, Debug)]
+pub struct MiiSelector {
+    config: Box<ctru_sys::MiiSelectorConf>,
+}
+
+/// Return value from a MiiSelector's launch
+#[derive(Clone, Debug)]
+pub struct MiiSelectorReturn {
```
Meziu, line 57: `MiiSelectorReturn` should be `non_exhaustive` to prevent the user from creating it without using the service.

Meziu (changes requested):

ctru-rs/src/mii.rs
Techie-Pi: Hmm now that I see that maybe you're right here, I had misread what you said :p

Techie-Pi: Turns out you were completely right! The values I were using for testing just didn't trigger any bug

Techie-Pi:
I've squashed some commits (mostly docs and cargo fmt/clippy)

Meziu (approved):
Looks good to me!

ctru-rs/src/applets/mii_selector.rs
```diff
@@ -0,0 +1,188 @@
...
+            Err(MiiLaunchError::InvalidChecksum)
+        }
+    }
+}
+
+impl From<ctru_sys::MiiSelectorReturn> for MiiSelectorReturn {
+    fn from(ret: ctru_sys::MiiSelectorReturn) -> Self {
+        let raw_mii_data = ret.mii;
+        let no_mii_selected = ret.no_mii_selected;
+        let guest_mii_index_clone = ret.guest_mii_index;
```
AzureMarker: nit: This variable isn't needed, you can directly reference `ret.guest_mii_index`

ctru-rs/src/mii.rs
```diff
@@ -0,0 +1,462 @@
...
+            y_position: partial_u8_bits_to_u8(&raw_eye_details[25..=29]),
+        };
+
+        let eyebrow_details = EyebrowDetails {
+            style: partial_u8_bits_to_u8(&raw_eyebrow_details[0..=4]),
+            color: partial_u8_bits_to_u8(&raw_eyebrow_details[5..=7]),
+            scale: partial_u8_bits_to_u8(&raw_eyebrow_details[8..=11]),
+            y_scale: partial_u8_bits_to_u8(&raw_eyebrow_details[12..=14]),
+            rotation: partial_u8_bits_to_u8(&raw_eyebrow_details[16..=19]),
+            x_spacing: partial_u8_bits_to_u8(&raw_eyebrow_details[21..=24]),
```
AzureMarker, line 359: Are there supposed to be gaps here?

ctru-rs/src/mii.rs
```diff
@@ -0,0 +1,462 @@
...
+    data.into_iter()
+        .fold(0, |result, bit| (result << 1) ^ u8::from(bit))
+}
+
+/// Given a series of LE bits, they are filled until a full LE u8 is reached
+fn partial_u8_bits_to_u8(data: &[bool]) -> u8 {
+    let leading_zeroes_to_add = 8 - data.len();
+    let leading_zeroes = vec![false; leading_zeroes_to_add];
+
+    vec_bit_to_u8([data, &leading_zeroes].concat().try_into().unwrap())
```
AzureMarker, line 450: If these are leading zeros, why aren't they at the start of the list?

ctru-rs/src/mii.rs
Techie-Pi: Yes, here it is

ctru-rs/src/mii.rs
Techie-Pi: Because of the way the bits are ordered by the 3ds, (this time I tested thoroughly to make sure the values returned are correct, so this should always work :p, but I'll double check later)

Techie-Pi: Yeah. Tested again and it works properly

ian-h-chamberlain (approved):
Sorry I meant to review this a while ago, but finally got a chance to take a look and it seems pretty cool! 

Some of `ctru::MiiData` might be worth documenting further, if we're going to maintain a totally separate struct vs the `ctru_sys::MiiData`, but I think this is good to start. There's lots of documentation stuff that needs to be done across the board so I don't think it should keep this from merging as-is.

Techie-Pi:
After current concerns about this PR are resolved, I plan on adding (at least) some basic documentation (and maybe references to the 3dbrew wiki) before merging, though I think the best thing would be to merge this PR with basic documentation and add more detailed documentation in a follow-up PR (I'm not a native English speaker, so I may need help in the follow-up PR, which may take some time to make because of that :p)

AzureMarker (approved):

Techie-Pi:
Is there anything that should be documented in this PR instead of the follow-up? Or is this ready to merge?

Meziu:
> Is there anything that should be documented in this PR instead of the follow-up? Or is this ready to merge?

I believe it’s fine for now. PR anything you want to add.

PR #81, opened 2022-11-22
Add LINEAR allocator module

Meziu:
LINEAR memory is one of the fundamental features of the 3DS architecture when using GPU and DSP (audio) processes.
While we could make all allocations ourselves inside of the wrappers for such services, we would be taking away from the user the capability of this RAM sector, as well as slowing down the whole process (we'd have to copy each "asset" from normal heap to linear before being able to actually use it).

I don't like having to add yet another unstable feature, but it's the best way to have instant support for all `std` structs, like native `Box`es on the LINEAR memory.

This PR will become extremely useful in the writing of `citro3d-rs` and the `ndsp` wrapper module that I'm already working on.

Edit:
For example, it would simplify (and make safe) this call: https://github.com/rust3ds/citro3d-rs/blob/6ba85149d84aa3bcc879ec16ef79079dc411643f/citro3d/examples/triangle.rs#L140

### DIFF
--- a/ctru-rs/examples/linear-memory.rs
+++ b/ctru-rs/examples/linear-memory.rs
@@ -1,7 +1,7 @@
 #![feature(allocator_api)]
 
-use ctru::prelude::*;
 use ctru::linear::LinearAllocator;
+use ctru::prelude::*;
 
 fn main() {
     ctru::init();
@@ -15,13 +15,16 @@ fn main() {
     // Normal `Box` on the heap
     let heap_box = Box::new(1492);
     // `Box` living on the linear memory sector
-    let linear_box: Box::<i32, LinearAllocator> = Box::new_in(2022, LinearAllocator);
+    let linear_box: Box<i32, LinearAllocator> = Box::new_in(2022, LinearAllocator);
 
     println!("This value is from the heap: {heap_box}");
     println!("This value is from the LINEAR memory: {linear_box}");
 
     println!("\nLINEAR space free before allocation: {linear_space_before}");
-    println!("LINEAR space free after allocation: {}", LinearAllocator::free_space());
+    println!(
+        "LINEAR space free after allocation: {}",
+        LinearAllocator::free_space()
+    );
 
     println!("\x1b[29;16HPress Start to exit");
 
--- a/ctru-rs/examples/linear-memory.rs
+++ b/ctru-rs/examples/linear-memory.rs
@@ -1,0 +1,43 @@
+#![feature(allocator_api)]
+
+use ctru::prelude::*;
+use ctru::linear::LinearAllocator;
+
+fn main() {
+    ctru::init();
+    let gfx = Gfx::init().expect("Couldn't obtain GFX controller");
+    let hid = Hid::init().expect("Couldn't obtain HID controller");
+    let apt = Apt::init().expect("Couldn't obtain APT controller");
+    let _console = Console::init(gfx.top_screen.borrow_mut());
+
+    let linear_space_before = LinearAllocator::free_space();
+
+    // Normal `Box` on the heap
+    let heap_box = Box::new(1492);
+    // `Box` living on the linear memory sector
+    let linear_box: Box::<i32, LinearAllocator> = Box::new_in(2022, LinearAllocator);
+
+    println!("This value is from the heap: {heap_box}");
+    println!("This value is from the LINEAR memory: {linear_box}");
+
+    println!("\nLINEAR space free before allocation: {linear_space_before}");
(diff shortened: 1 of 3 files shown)


ctru-rs/examples/linear-memory.rs
```diff
@@ -0,0 +1,46 @@
...
+    let hid = Hid::init().expect("Couldn't obtain HID controller");
+    let apt = Apt::init().expect("Couldn't obtain APT controller");
+    let _console = Console::init(gfx.top_screen.borrow_mut());
+
+    let linear_space_before = LinearAllocator::free_space();
+
+    // Normal `Box` on the heap
+    let heap_box = Box::new(1492);
+    // `Box` living on the linear memory sector
+    let linear_box: Box<i32, LinearAllocator> = Box::new_in(2022, LinearAllocator);
```
Techie-Pi, line 18: Is adding this explicit type needed?

Techie-Pi (approved):
Looks great! Just a nit ;p

ctru-rs/examples/linear-memory.rs
Meziu: Not really, but I believe it’s better to have it explicit since the `Box` type is usually believed to be just `Box<T>`. It’s just an example, Rust can infer the type obviously.

Techie-Pi: Alright, makes sense

ctru-rs/src/linear.rs
```diff
@@ -0,0 +1,42 @@
...
+        unsafe { ctru_sys::linearSpaceFree() }
+    }
+}
+
+unsafe impl Allocator for LinearAllocator {
+    fn allocate(
+        &self,
+        layout: std::alloc::Layout,
+    ) -> Result<std::ptr::NonNull<[u8]>, std::alloc::AllocError> {
+        let pointer = unsafe { ctru_sys::linearAlloc(layout.size() as u32) };
```
ian-h-chamberlain: It looks like the default `linearAlloc` behavior is to just align everything to `0x80` which might be wasteful, do you think it would make sense to use `linearMemAlign(size, alignment)` instead since we have the full `Layout` information?

ctru-rs/src/linear.rs
```diff
@@ -0,0 +1,42 @@
...
+        let slice: &mut [u8] =
+            unsafe { std::slice::from_raw_parts_mut(pointer as *mut u8, layout.size()) };
```
ian-h-chamberlain: I think it might be UB to build a struct like this before checking if the pointer is non-null — I'm not sure if the call to `NonNull::new` can properly detect a nullptr after this point.

If you're willing to add another unstable feature, we could probably use [`slice_from_raw_parts`](https://doc.rust-lang.org/std/ptr/struct.NonNull.html#method.slice_from_raw_parts) for this to do it without `unsafe`, like

```rs
NonNull::new(pointer)
    .map(|ptr| NonNull::slice_from_raw_parts(ptr, layout.size()))
    .ok_or(AllocError)
```

ctru-rs/src/linear.rs
```diff
@@ -0,0 +1,42 @@
...
+
+        std::ptr::NonNull::new(slice).ok_or(std::alloc::AllocError)
+    }
+
+    unsafe fn deallocate(&self, ptr: std::ptr::NonNull<u8>, _layout: std::alloc::Layout) {
+        ctru_sys::linearFree(ptr.as_ptr() as *mut _);
```
ian-h-chamberlain: nit: I think clippy recommends `.cast()` instead of `as` when converting from `*mut T` to `*mut U` (when checking `clippy::pedantic`)

ctru-rs/examples/linear-memory.rs
```diff
@@ -0,0 +1,46 @@
+#![feature(allocator_api)]
```
ian-h-chamberlain, line 1: It might be worth mentioning in the doc comment that users also need to opt-in to this feature to use the allocator.

ian-h-chamberlain (approved):
Looks good! I think this will definitely be useful for `citro3d` and hopefully reduce the unsafe surface area a bit.

ctru-rs/src/linear.rs
Meziu: It's not UB, but `from_raw_parts_mut` does cause a panic (not wanted since NULL values should be handled within the `Result`). I will change this.

ctru-rs/src/linear.rs
Meziu: I had actually forgot about the `Layout` type functionality! Thanks for the reminder.

ian-h-chamberlain (approved):

PR #83, opened 2023-01-04
Addition of the NDSP module and related examples

Meziu:
I've been working on this module for a while, being one of the most chaotic pieces of code I have ever had to read through. Even though this is actually `libctru`'s second revision of an "audio" service, the system can barely stand on its feet, and making a *safe* wrapper depending solely on the assumptions (or lack thereof) `libctru` makes is not a simple task.

Even though it looks like it works as I wrote it, there are many more things I have to check and fix and other I simply have to add. Furthermore, I'm having problems tuning the `audio-filters` example to behave exactly the same way as the one in [3ds-examples](https://github.com/devkitPro/3ds-examples/tree/master/audio/filters).

This is why I made this PR draft: I need feedback on the overall structure of the wrapper (before venturing into audio streaming), and I need help to get the example to actually work.

~Even though the examples "procedure" seems to be working, the sound produced has 2 issues: it's not continuous and it makes a weird pop in the background. The `3ds-example`'s code generates a clean continuous wave.~
Edit: found the main issue. I finally understood the inner workings of the samples, and why  `3ds-examples` made all that conversion efforts. It works *better*, but I still find a small issue in this version: it seems the audio doesn't finish properly or something, since it's possible to hear some overlap the longer the app runs. I'd investigate more but  my 3DS' battery ran out. To tomorrow :smile:.

It's hacky, it's unfinished, but that's why I want to hear your thoughts :smiley:


ctru-rs/examples/audio_filters.rs
```diff
@@ -0,0 +1,187 @@
+#![feature(allocator_api)]
```
AzureMarker: nit: rename example file to `audio-filters.rs` to remain consistent with other examples.

ctru-rs/src/services/ndsp/mod.rs
```diff
@@ -0,0 +1,243 @@
...
+#[derive(Copy, Clone, Debug)]
+#[repr(u32)]
+pub enum AudioFormat {
+    PCM8Mono = ctru_sys::NDSP_FORMAT_MONO_PCM8,
+    PCM16Mono = ctru_sys::NDSP_FORMAT_MONO_PCM16,
+    ADPCMMono = ctru_sys::NDSP_FORMAT_MONO_ADPCM,
+    PCM8Stereo = ctru_sys::NDSP_FORMAT_STEREO_PCM8,
+    PCM16Stereo = ctru_sys::NDSP_FORMAT_STEREO_PCM16,
+    FrontBypass = ctru_sys::NDSP_FRONT_BYPASS,
+    SurroundPreprocessed = ctru_sys::NDSP_3D_SURROUND_PREPROCESSED,
```
AzureMarker: These two are marked as flags in libctru, so they might not be full audio formats?

ctru-rs/src/services/ndsp/mod.rs
```diff
@@ -0,0 +1,243 @@
...
+#[repr(u32)]
+pub enum InterpolationType {
+    Polyphase = ctru_sys::NDSP_INTERP_POLYPHASE,
+    Linear = ctru_sys::NDSP_INTERP_LINEAR,
+    None = ctru_sys::NDSP_INTERP_NONE,
+}
+
+pub struct Channel {
+    id: i32,
+}
```
AzureMarker: nit: Maybe change to the below? Not sure if there's more than an ID needed for this.
```suggestion
pub struct Channel(i32);
```

ctru-rs/src/services/ndsp/mod.rs
```diff
@@ -0,0 +1,243 @@
...
+
+static NDSP_ACTIVE: Lazy<Mutex<usize>> = Lazy::new(|| Mutex::new(0));
```
AzureMarker: `Mutex::new` is now const, no `Lazy` required :)

ctru-rs/src/services/ndsp/mod.rs
```diff
@@ -0,0 +1,243 @@
...
+
+#[non_exhaustive]
```
AzureMarker: nit: This isn't needed since the struct fields are private

ctru-rs/src/services/ndsp/mod.rs
```diff
@@ -0,0 +1,243 @@
...
+    /// # Errors
+    ///
+    /// An error will be returned if the channel id is not between 0 and 23.
+    pub fn channel(&self, id: u8) -> crate::Result<Channel> {
+        if id > 23 {
+            return Err(crate::Error::InvalidChannel(id.into()));
+        }
+
+        Ok(Channel { id: id.into() })
+    }
```
AzureMarker: Maybe this can be moved to `Channel::new`? It doesn't seem require a NDSP service reference, unless channel operations will fail until `ndspInit` is called?.

ctru-rs/src/services/ndsp/wave.rs
```diff
@@ -0,0 +1,138 @@
+use ctru_sys::ndspChnWaveBufClear;
+
+use super::AudioFormat;
+use crate::linear::LinearAllocator;
+
+/// Base struct to represent audio wave data. This requires audio format information.
+pub struct WaveBuffer {
+    /// Buffer data. This data must be allocated on the LINEAR memory.
+    data: Box<[u8], LinearAllocator>,
```
AzureMarker: It's great that we're able to specify where this memory needs to be allocated in the type system!

ctru-rs/src/services/ndsp/wave.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+impl<'b> WaveInfo<'b> {
+    pub fn new(buffer: &'b mut WaveBuffer, looping: bool) -> Self {
+        let address = ctru_sys::tag_ndspWaveBuf__bindgen_ty_1 {
+            data_vaddr: buffer.data.as_ptr().cast(),
+        };
+
+        let raw_data = ctru_sys::ndspWaveBuf {
+            __bindgen_anon_1: address, // Buffer data virtual address
```
AzureMarker, line 44: Any way to make this field named better?

AzureMarker (commented):
Only did a shallow pass over the code, but it looks cool!

ctru-rs/src/services/ndsp/mod.rs
Meziu: I was thinking of getting rid of them entirely anyways, they don’t seem to blend in very well and don’t have any docs.

ctru-rs/src/services/ndsp/mod.rs
Meziu: I missed this! We have some cleanup to do… 😄

ctru-rs/src/services/ndsp/mod.rs
Meziu: They won’t fail, but they also won’t do anything either. Since `Channel` functionality is entirely dependent on NDSP, i’d keep it this way.

ctru-rs/src/services/ndsp/wave.rs
Meziu: It is! Having RAM sectors we can freely allocate to is very nice, and this also means we can do static memory checks without any runtime cost.

ctru-rs/src/services/ndsp/wave.rs
Meziu: I’m no expert in bindgen, but I think it’s needed since the original struct has a completely anonymous field (thing that doesn’t exist in Rust, for good reasons). I wonder if we can make a special case in the build script to give it a name ourselves…

Meziu: https://github.com/rust-lang/rust-bindgen/issues/1971 looks like it currently isn't an option

Meziu:
The example works fine now though I had to turn up the buffer length. My guess to why it didn't work is that due to `chunks_exact_mut` and the other wrapping operations my implementation of `fill_buffer` was actually slower when accessing memory, creating delays between the "ending" of a buffer and the start of the next one (which caused popping). I could make an unsafe version to see the results (basically copying 1:1 the one in `3ds_examples`) but the use of too much unsafe code in an example seems against its purpose.

Meziu:
@ian-h-chamberlain Since you made the GitHub Actions and test implementation, could you take a look at why the checks fail? They seem to fail in the test implementation, which isn't related to our commits...

Techie-Pi:
Looks like the test fails with the latest nightly because `TestOpts` recently added [`fail_fast`](https://doc.rust-lang.org/nightly/test/struct.TestOpts.html#structfield.fail_fast). And the 2022-07-18 nightly test fails because [`cast_mut`](https://doc.rust-lang.org/std/primitive.pointer.html#method.cast_mut) was added on 1.65 (and nightly-2022-07-18 is near 1.62). So maybe we should update the nightly used in the nightly-2022-07-18 test?

PS I'll review this code in a few weeks, but nothing stands out to me after the changes brought by @AzureMarker's review (though I just did a shallow pass over the code) :)

ctru-rs/src/services/ndsp/wave.rs
```diff
@@ -0,0 +1,138 @@
...
+                     // adpcm_data: AdpcmData, TODO: Requires research on how this format is handled.
+}
+
+/// Informational struct holding the raw audio data and playaback info. This corresponds to [ctru_sys::ndspWaveBuf]
+pub struct WaveInfo<'b> {
+    /// Data block of the audio wave (plus its format information).
+    buffer: &'b mut WaveBuffer,
+    // Holding the data with the raw format is necessary since `libctru` will access it.
+    pub(crate) raw_data: ctru_sys::ndspWaveBuf,
+    played_on_channel: Option<u8>,
```
ian-h-chamberlain, line 11: Should this be an `Option<super::Channel>`? Or is that a different type of channel than the one used here?

Edit: I see, this is only used internally so probably doesn't matter much

ctru-rs/src/services/ndsp/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+// As such, there is no need to hold channels to ensure correct mutability.
+// As such, this struct is more of a dummy than an actually functional block.
+impl Channel {
+    /// Reset the channel
+    pub fn reset(&self) {
+        unsafe { ctru_sys::ndspChnReset(self.0) };
+    }
+
+    /// Initialize the channel's parameters
+    pub fn init_parameters(&self) {
```
ian-h-chamberlain, line 129: Is this initialization something that you always have to do? Might it make sense to have a `Channel::new()` that calls this, and call `reset()` (and/or `clear_queue()`) when the channel is dropped?

Edit: if you decided to go this route, you could have something like:
```rust
struct Ndsp {
    channels: [Channel; 24]
}

impl Ndsp {
    pub fn channel(&self, id: u8) -> Result<Channel> {
        self.channels.get(id).ok_or("invalid channel")
    }
}
```

You mentioned that all those operations are thread-safe though so maybe just `init` and cleanup would be enough, although it might be weird if a program got two instances of the same channel and it ended up resetting while some other thing was using it...

ctru-rs/examples/audio-filters.rs
```diff
@@ -0,0 +1,175 @@
...
+fn fill_buffer(audio_data: &mut [u8], frequency: f32) {
+    let formatted_data = audio_data.chunks_exact_mut(4);
+
+    for (i, chunk) in formatted_data.enumerate() {
+        // This is a simple sine wave, with a frequency of `frequency` Hz, and an amplitude 30% of maximum.
+        let sample: f32 = (frequency * (i as f32 / SAMPLE_RATE as f32) * 2. * PI).sin();
+        let amplitude = 0.3 * i16::MAX as f32;
+
+        // This operation is safe, since we are writing to a slice of exactly 32 bits
+        let chunk_ptr: &mut [i16; 2] = unsafe { &mut *(chunk.as_mut_ptr() as *mut [i16; 2]) };
```
ian-h-chamberlain: Ah, I thought we had `bytemuck` in `dev-dependencies` already but maybe that was in `citro3d`. If you feel like using it they have nice helpers for this kind of thing:

```rs
let [left, right] = bytemuck::from_bytes_mut::<[i16; 2]>(chunk);
*left = 5;
*right = 5;
```

ian-h-chamberlain (approved):
Very cool! I don't have too many comments on the implementation itself, it seems pretty reasonable (although I am no audio expert lol). Now to see if there are some cool rust audio libraries that we can plug into this and play music and stuff 😈

ctru-rs/src/services/ndsp/mod.rs
Meziu: > You mentioned that all those operations are thread-safe though so maybe just `init` and cleanup would be enough, although it might be weird if a program got two instances of the same channel and it ended up resetting while some other thing was using it...

That's exactly why I did it this way. The other option would be to have references to the `Ndsp` object and to count for `Channel` references to avoid multiple instances.

I believe the reference is needed either way (right now `Channels` can outlive `Ndsp`, bad behaviour), but I was wondering if there were better ways to count channels than just using a 24bit flag... Also, the reference should be unmutable (so that we can have multiple channels running at once), but changing the flag (located in Ndsp) using only an unmutable reference causes other troubles.

ctru-rs/examples/audio-filters.rs
Meziu: i went for an approach to avoid `as_chunks_mut`. Thanks for the suggestion!

ctru-rs/src/services/ndsp/mod.rs
ian-h-chamberlain: Here's a possible kind of implementation? https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e136b260313bbca1e62f31dc4fddaf99

It allows aliasing on the channels with immutable refs, only initializes / resets each one once (although perhaps manually resetting would be trickyier), and does so lazily as-needed. It's probably not perfect and might be missing some use cases but it at least ties the lifetimes together.

ctru-rs/examples/audio-filters.rs
```diff
@@ -0,0 +1,174 @@
...
+        "Peaking",
+    ];
+
+    let mut filter: usize = 0;
+
+    // We set up two wave buffers and alternate between the two,
+    // effectively streaming an infinitely long sine wave.
+
+    let mut audio_data1 = Box::new_in([0u8; AUDIO_WAVE_LENGTH], LinearAllocator);
+    fill_buffer(&mut audio_data1[..], NOTEFREQ[4]);
```
AzureMarker: nit: `[..]` isn't needed

ctru-rs/examples/audio-filters.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    let mut audio_data2 = Box::new_in([0u8; AUDIO_WAVE_LENGTH], LinearAllocator);
+    fill_buffer(&mut audio_data2[..], NOTEFREQ[4]);
```
AzureMarker: nit: this could be copied from the first buffer instead of recalculated.

ctru-rs/examples/audio-filters.rs
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        let mut update_params = false;
+        if keys_down.intersects(KeyPad::KEY_LEFT) {
+            let wraps;
+            (filter, wraps) = filter.overflowing_sub(1);
+
+            if wraps {
+                filter = filter_names.len() - 1;
+            }
```
AzureMarker: I feel like a bunch of this +1/-1 logic can be replaced with [`rem_euclid`](https://doc.rust-lang.org/stable/std/primitive.i32.html#method.rem_euclid)

ctru-rs/examples/audio-filters.rs
```diff
@@ -0,0 +1,174 @@
...
+        } else if keys_down.intersects(KeyPad::KEY_RIGHT) {
+            filter += 1;
+            if filter >= filter_names.len() {
+                filter = 0;
+            }
+            update_params = true;
+        }
+
+        // Check for upper limit
+        note = std::cmp::min(note, NOTEFREQ.len() - 1);
```
AzureMarker: nit: This should be done closer to the code that modifies `note`. Ex:
```rust
note = std::cmp::min(note + 1, NOTEFREQ.len() - 1);
```

ctru-rs/src/services/ndsp/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+pub struct Channel(i32);
+
+static NDSP_ACTIVE: Mutex<usize> = Mutex::new(0);
+
+pub struct Ndsp {
+    _service_handler: ServiceReference,
+}
+
+impl Ndsp {
+    pub fn init() -> crate::Result<Self> {
```
AzureMarker, line 76: nit: add some documentation, especially noting that this initialization will fail if previously initialized.

ctru-rs/src/services/ndsp/mod.rs
AzureMarker: Maybe replace the `Cell` with an `AtomicBool` and use `swap(true, Ordering::SeqCst)` to avoid two threads both racing to initialize the channel (even though libctru makes it thread safe, we probably don't want to initialize twice and possibly lose one thread's settings).

ctru-rs/src/services/ndsp/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+    pub fn set_mix(&self, mix: &[f32; 12]) {
+        unsafe { ctru_sys::ndspChnSetMix(self.0, mix.as_ptr().cast_mut()) }
+    }
+
+    /// Set the channel's rate of sampling.
+    pub fn set_sample_rate(&self, rate: f32) {
+        unsafe { ctru_sys::ndspChnSetRate(self.0, rate) };
+    }
+
+    // TODO: find a way to wrap `ndspChnSetAdpcmCoefs`
```
AzureMarker: Reminder to handle this TODO

ctru-rs/src/services/ndsp/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    /// Add a wave buffer to the channel's queue.
+    /// If there are no other buffers in queue, playback for this buffer will start.
+    ///
+    /// # Warning
+    ///
+    /// `libctru` expects the user to manually keep the info data (in this case [WaveInfo]) alive during playback.
+    /// To ensure safety, checks within [WaveInfo] will clear the whole channel queue if any queued [WaveInfo] is dropped prematurely.
+    pub fn queue_wave(&self, wave: &mut WaveInfo) {
+        // TODO: Return an error for already queued/used WaveInfos.
```
AzureMarker: Reminder to handle this TODO

ctru-rs/src/services/ndsp/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+            AudioFormat::PCM8Mono | AudioFormat::ADPCMMono => 1,
+            AudioFormat::PCM16Mono | AudioFormat::PCM8Stereo => 2,
+            AudioFormat::PCM16Stereo => 4,
+        }
+    }
+}
+
+impl Drop for Ndsp {
+    fn drop(&mut self) {
+        for i in 0..24 {
```
AzureMarker: nit: maybe 24 can become a constant for the number of channels

ctru-rs/src/services/ndsp/wave.rs
```diff
@@ -0,0 +1,138 @@
...
+use super::AudioFormat;
+use crate::linear::LinearAllocator;
+
+/// Base struct to represent audio wave data. This requires audio format information.
+pub struct WaveBuffer {
+    /// Buffer data. This data must be allocated on the LINEAR memory.
+    data: Box<[u8], LinearAllocator>,
+    audio_format: AudioFormat,
+    nsamples: usize, // We don't use the slice's length here because depending on the format it may vary
+                     // adpcm_data: AdpcmData, TODO: Requires research on how this format is handled.
```
AzureMarker: Reminder to handle this TODO

ctru-rs/src/services/ndsp/wave.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+/// Informational struct holding the raw audio data and playaback info. This corresponds to [ctru_sys::ndspWaveBuf]
+pub struct WaveInfo<'b> {
+    /// Data block of the audio wave (plus its format information).
+    buffer: &'b mut WaveBuffer,
```
AzureMarker: Does this need to be a mutable buffer?

ctru-rs/src/services/ndsp/wave.rs
```diff
@@ -0,0 +1,138 @@
...
+    Playing = ctru_sys::NDSP_WBUF_PLAYING as u8,
+    Done = ctru_sys::NDSP_WBUF_DONE as u8,
+}
+
+impl WaveBuffer {
+    pub fn new(data: Box<[u8], LinearAllocator>, audio_format: AudioFormat) -> crate::Result<Self> {
+        let nsamples: usize = data.len() / (audio_format.sample_size() as usize);
+
+        unsafe {
+            let _r = ctru_sys::DSP_FlushDataCache(data.as_ptr().cast(), data.len() as u32);
```
AzureMarker: Would be good to have a comment explaining this line

ctru-rs/src/services/ndsp/wave.rs
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        Ok(Self {
+            data,
+            audio_format,
+            nsamples,
+        })
+    }
+
+    pub fn get_mut_data(&mut self) -> &mut [u8] {
```
AzureMarker: I think this is normally named `get_data_mut` (`mut` as a suffix)

ctru-rs/src/services/ndsp/wave.rs
```diff
@@ -0,0 +1,138 @@
...
+        &mut self.data
+    }
+
+    pub fn get_format(&self) -> AudioFormat {
+        self.audio_format
+    }
+
+    pub fn get_sample_amount(&self) -> usize {
```
AzureMarker: nit: maybe rename to `get_sample_count`?

ctru-rs/src/services/ndsp/wave.rs
```diff
@@ -0,0 +1,138 @@
...
+
+use super::AudioFormat;
+use crate::linear::LinearAllocator;
+
+/// Base struct to represent audio wave data. This requires audio format information.
+pub struct WaveBuffer {
+    /// Buffer data. This data must be allocated on the LINEAR memory.
+    data: Box<[u8], LinearAllocator>,
+    audio_format: AudioFormat,
+    nsamples: usize, // We don't use the slice's length here because depending on the format it may vary
```
AzureMarker: nit: change to a more readable name, like `sample_count`

ctru-rs/src/services/ndsp/wave.rs
```diff
@@ -0,0 +1,138 @@
...
+        };
+
+        Self {
+            buffer,
+            raw_data,
+            played_on_channel: None,
+        }
+    }
+
+    pub fn get_mut_wavebuffer(&mut self) -> &mut WaveBuffer {
```
AzureMarker: Same comment as above: rename to `get_wavebuffer_mut`

ctru-rs/src/services/ndsp/wave.rs
```diff
@@ -0,0 +1,138 @@
...
+pub struct WaveBuffer {
+    /// Buffer data. This data must be allocated on the LINEAR memory.
+    data: Box<[u8], LinearAllocator>,
+    audio_format: AudioFormat,
+    nsamples: usize, // We don't use the slice's length here because depending on the format it may vary
+                     // adpcm_data: AdpcmData, TODO: Requires research on how this format is handled.
+}
+
+/// Informational struct holding the raw audio data and playaback info. This corresponds to [ctru_sys::ndspWaveBuf]
+pub struct WaveInfo<'b> {
```
AzureMarker: Why do we need both `WaveBuffer` and `WaveInfo` instead of just one struct?

ctru-rs/src/services/ndsp/wave.rs
```diff
@@ -0,0 +1,138 @@
...
+        self.raw_data.status.try_into().unwrap()
+    }
+
+    pub(crate) fn set_channel(&mut self, id: i32) {
+        self.played_on_channel = Some(id as u8)
+    }
+}
+
+impl TryFrom<u8> for WaveStatus {
+    type Error = String;
```
AzureMarker: nit: maybe change to `&'static str`

ctru-rs/examples/audio-filters.rs
Meziu: It was that way earlier, but it got split when testing the offset. Thanks for noticing. 😄

ctru-rs/src/services/ndsp/wave.rs
Meziu: Before closing up the example I had in mind a different system for WaveBuffers and WaveInfos, so that they could remain separate things (helping with clones etc). That didn't come to be because of the difficulties `ndsp` gives with lifetimes. I'll just merge them.

ctru-rs/src/services/ndsp/wave.rs
Meziu: Fixed thanks to the type merge.

ctru-rs/src/services/ndsp/wave.rs
Meziu: Indeed. Depending on how I'll solve the `Channel` lifetime issue it could be hard to hold it in there, so I see no reason to hold anything more than a single byte.

ctru-rs/src/services/ndsp/mod.rs
Meziu: > Maybe replace the `Cell` with an `AtomicBool` and use `swap(true, Ordering::SeqCst)` to avoid two threads both racing to initialize the channel (even though libctru makes it thread safe, we probably don't want to initialize twice and possibly lose one thread's settings).

That can't happen (I think?). Our `Ndsp` struct is blocked by `ServiceReference` which locks it to only one thread.

ian-h-chamberlain (approved):

Meziu:
Seems like ADPCM isn't worth the effort. I've seen some people around be able to reverse engineer it and use it, but it's overly complex and inaccessible. The only reason why someone would use that nowadays (in non-licensed software) is for modding, since many Nintendo games use that format for compressed audio.

Meziu:
Seems like most of the work on the backend is done. This makes me very happy :smile:. I will soon start working on some decoding examples, so we can play some real :notes: tunes :notes:  on the console.

ctru-rs/src/services/ndsp/mod.rs
Meziu: The `Cell` approach wasn’t doable, but I managed to “merge” the lifetime check and the exclusivity check by using `RefCell` in `Ndsp`. As always, it’s a bit heavy on the system, but it’s needed for safety…

ctru-rs/src/services/ndsp/mod.rs
```diff
@@ -209,13 +276,35 @@ impl AudioFormat {
...
+        match self {
+            Self::InvalidChannel(id) => write!(f, "Audio Channel with id {id} doesn't exist. Valid channels have an id between 0 and 23."),
+            Self::ChannelAlreadyInUse(id) => write!(f, "Audio Channel with id {id} is already being used. Drop the other instance if you want to use it here."),
+            Self::WaveAlreadyQueued(id) => write!(f, "The selected WaveInfo is already playing on channel {id}.")
+        }
+    }
+}
+
+impl error::Error for NdspError {
+    fn description(&self) -> &str {
```
ian-h-chamberlain: Any reason to define this when already using `Display`? It looks [to be deprecated](https://doc.rust-lang.org/std/error/trait.Error.html#method.description) for a while

ian-h-chamberlain (approved):

ctru-rs/src/services/ndsp/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+        unsafe { ctru_sys::ndspChnSetFormat(self.id.into(), format as u16) };
+    }
+
+    /// Set the channel's interpolation mode.
+    pub fn set_interpolation(&self, interp_type: InterpolationType) {
+        unsafe { ctru_sys::ndspChnSetInterp(self.id.into(), interp_type as u32) };
+    }
+
+    /// Set the channel's volume mix.
+    /// Docs about the buffer usage: <https://libctru.devkitpro.org/channel_8h.html#a30eb26f1972cc3ec28370263796c0444>
```
Techie-Pi: Maybe we could move the documentation about `mix` here?

ctru-rs/src/services/ndsp/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+#[repr(u32)]
+pub enum InterpolationType {
+    Polyphase = ctru_sys::NDSP_INTERP_POLYPHASE,
+    Linear = ctru_sys::NDSP_INTERP_LINEAR,
+    None = ctru_sys::NDSP_INTERP_NONE,
+}
+
+#[derive(Copy, Clone, Debug)]
+pub enum NdspError {
+    InvalidChannel(u8),      // channel id
```
Techie-Pi: Maybe documenting that the values here are channel IDs for users of the library would be a good idea
```suggestion
    /// Channel ID
    InvalidChannel(u8),
```

or something like
```rust
/// The values contained inside the members of this enum are the Channel IDs
pub enum NdspError {
...
```

ctru-rs/src/services/ndsp/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+            _service_handler,
+            channel_flags: Default::default(),
+        })
+    }
+
+    /// Return a representation of the specified channel.
+    ///
+    /// # Errors
+    ///
+    /// An error will be returned if the channel id is not between 0 and 23 or if the specified channel is already being used.
```
Techie-Pi:
```suggestion
    /// An error will be returned if the channel ID is not between 0 and 23 or if the specified channel is already being used.
```

ctru-rs/src/services/ndsp/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+        unsafe { ctru_sys::ndspChnWaveBufAdd(self.id.into(), &mut wave.raw_data) };
+
+        Ok(())
+    }
+}
+
+/// Functions to handle audio filtering.
+/// Refer to [libctru](https://libctru.devkitpro.org/channel_8h.html#a1da3b363c2edfd318c92276b527daae6) for more info.
+impl Channel<'_> {
+    pub fn iir_mono_set_enabled(&self, enable: bool) {
```
Techie-Pi, line 230: Same as before, maybe it's a good idea to move the documentation of the functions here to improve DX of users of the library

ctru-rs/src/services/ndsp/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+            AudioFormat::PCM16Mono | AudioFormat::PCM8Stereo => 2,
+            AudioFormat::PCM16Stereo => 4,
+        }
+    }
+}
+
+impl fmt::Display for NdspError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Self::InvalidChannel(id) => write!(f, "Audio Channel with id {id} doesn't exist. Valid channels have an id between 0 and 23."),
```
Techie-Pi:
```suggestion
            Self::InvalidChannel(id) => write!(f, "Audio Channel with ID {id} doesn't exist. Valid channels have an id between 0 and 23."),
```

Same for the others

Techie-Pi (approved):
Seems great! Just a few nits :)

AzureMarker (approved):

Meziu:
Edit: On second thought, adding weight to this PR is unneeded, and more complex examples should come once multi-threading enables optimizations. I just added some docs and put a new error case in, it should be fine now!

PR #86, opened 2023-01-14
Add an ir:USER service wrapper and Circle Pad Pro example

AzureMarker:
I think this is the first time anyone from homebrew has connected to the Circle Pad Pro (or they never talked about it online). It took a while, but I figured out you just have to be persistent during the connection phase (do immediate retries).

This PR adds a service wrapper for the ir:USER service, and a example/demo of using it to connect to the Circle Pad Pro (the main usage of the service as far as I can tell).

Note that the New 3DS internally emulates the Circle Pad Pro, so using `ir:USER` on a New 3DS system should give you the c-stick inputs. The `ir:rst` service is slightly different, where it only works on the New 3DS (doesn't give CPP inputs). Most retail 3DS games use `ir:USER` to be compatible with both old and new 3DS systems.

Reference links:
* https://www.3dbrew.org/wiki/IR_Services
* https://www.3dbrew.org/wiki/Circle_Pad_Pro
* https://www.3dbrew.org/wiki/IRUSER_Shared_Memory

BTW if you're curious what I want to use this for:
https://github.com/AzureMarker/n3ds-controller

Demo:
CPP demo photo

Note: the screen no longer flickers after enabling double buffering.

https://user-images.githubusercontent.com/4417660/212457545-5f12e5b4-85f3-4faa-9552-74ce43bacce8.mp4


ctru-rs/src/services/ir_user.rs
```diff
@@ -0,0 +1,504 @@
...
+    pub fn get_recv_event(&self) -> crate::Result<Handle> {
+        let response = self.send_service_request(vec![GET_RECEIVE_EVENT_COMMAND_HEADER], 4)?;
+        let recv_event = response[3] as Handle;
+
+        Ok(recv_event)
+    }
+
+    /// Wait for an event to fire. If the timeout is reached, an error is returned. You can use
+    /// [`Error::is_timeout`] to check if the error is due to a timeout.
+    pub fn wait_for_event(event: Handle, timeout: Duration) -> crate::Result<()> {
```
AzureMarker: This should probably be moved into a new `svc` module

Meziu: During the `cam` module addition, we deemed `svc` not well suited for user management yet. If you can find a way to make it completely safe, then we can work to re-write it.

AzureMarker: We can have a limited API surface, like just supporting `svcWaitSynchronization`

Meziu:
I honestly thought circle pad pro was automatically recognised by the system (I don’t have one, since I own a New 3DS XL with a built-in C-Stick)

ctru-rs/src/services/ir_user.rs
Meziu: That’s a good idea for now 👌

AzureMarker:
> I honestly thought circle pad pro was automatically recognised by the system (I don’t have one, since I own a New 3DS XL with a built-in C-Stick)

I thought the same thing, and initially started working on wrapping the `ir:rst` service. However, I found that the service only supports the New 3DS, and libctru already automatically initializes/uses it in the HID implementation:
https://github.com/devkitPro/libctru/blob/e253c2c00543ab96067dc241de8dd32d72e6ec99/libctru/source/services/hid.c#L65-L67

They leave the hard work of getting the CPP input to the user :upside_down_face:

Meziu:
> They leave the hard work of getting the CPP input to the user 🙃

I guess that makes sense. They noticed the difficulty to try to find a CPP and decided to make `Hid` more “stable” and not include it (since Hid runs in all apps regardless during the initial setup). Also… not many people miss that accessory 😂

AzureMarker:
Yeah I don't blame them, especially since it seems like no one knew how to get a connection working anyway. The Circle Pad Pro is pretty niche, but I really want that second joystick for my PC controller emulation :smiling_imp:.

SteveCookTU:
Really neat!

With the idea of how IR is receiving packets back and forth, is it worth possibly creating almost like an "IRServer" or "IRThread", etc that internally manages  waiting for packets? This way svc doesn't have to be exposed to the user similar to the `cam` service.

My thought is it would be more user safe if a mpsc receiver channel was returned from the initialization and you could use `recv` instead of a more raw call to `svc`.

The connect function could be moved to the `IrUser` module and pass in an `&Hid`, the device id (if there are more such as the pokewalker), and a button combination that could cancel the connection loop. At the different error points return a specific error and at the end return a `mpsc::receiver` or something. Thoughts?

I would write up an example but I'm currently on mobile.

AzureMarker:
ctru-rs is supposed to be a thin or minimal wrapper around libctru which makes it nicer and safer to use than raw C calls. With that in mind, we don't want to introduce more overhead or complexity than necessary, so the user application (or another library) should be where that IR thread and other connection logic goes. Especially when it involves handling user input. We don't want to force developers to handle initialization of the service/CPP in a specific way like the demo (i.e. taking in `Hid` and blocking execution). The suggestions are interesting and would be fine in their own crate though.

I think it's fine to expose some svc calls, especially when it's almost harmless like the one used here.

SteveCookTU:
In that case, if some of the SVC functions are added then the `cam` module could be simplified.

I guess the main concern would be the safety implications of the exposed functions. Ex. An invalid, null, or non-event handle being passed

ctru-rs/src/services/ir_user.rs
AzureMarker: Updated

ctru-rs/src/services/srv.rs
```diff
@@ -0,0 +1,20 @@
...
+//! Not all APIs are wrapped in this module, since a lot are fundamentally unsafe.
+//! Most APIs should be used directly from `ctru-sys`.
+
+use crate::error::ResultCode;
+use ctru_sys::Handle;
+use std::time::Duration;
+
+/// Wait for an event to fire. If the timeout is reached, an error is returned. You can use
+/// [`Error::is_timeout`] to check if the error is due to a timeout.
+pub fn wait_for_event(event: Handle, timeout: Duration) -> crate::Result<()> {
```
Meziu: This should be a method of `Handle`, rather than a static of the module.

Meziu (changes requested):
> Note that the New 3DS internally emulates the Circle Pad Pro, so using ir:USER on a New 3D system should give you the c-stick inputs. The ir:rst service is slightly different, where it only works on the New 3DS (doesn't give CPP inputs). Most retail 3DS games use ir:USER to be compatible with both old and new 3DS systems.

My N3DS XL won't recognize the c-stick as a CPP, it seems. It tries to check the infrared instead (I used my iPhone X's FaceID to shoot infrared in the ir sensor and it returns an error! :laughing:).

I can't really review the example/module if I can't run it :thinking:

SteveCookTU:
> My N3DS XL won't recognize the c-stick as a CPP, it seems. It tries to check the infrared instead (I used my iPhone X's FaceID to shoot infrared in the ir sensor and it returns an error! 😆).

Found this on the libctru docs. `Note that the main service for New3DS HID is ir:rst, and these two service are mutually exclusive: when one is initialized and reading data from New3DS HID, the other cannot access it.`

The example is initializing `hid` which in turn initializes `ir:rst` within `libctru`. My guess as to why it isn't getting inputs from the c-stick.

Meziu:
> Found this on the libctru docs. `Note that the main service for New3DS HID is ir:rst, and these two service are mutually exclusive: when one is initialized and reading data from New3DS HID, the other cannot access it.`
> 
> The example is initializing `hid` which in turn initializes `ir:rst` within `libctru`. My guess as to why it isn't getting inputs from the c-stick.

Ok, great, we just have to make a runtime check for New3DS and state the example won’t work without the actual CPP.

SteveCookTU:
Something to note:

There are a handful of devices that utilize `ir` including the NFC Reader, PokeWalker, and even some games directly access it (old Pokemon games syncing for local play). I'm curious how `ir` is utilized to handle the different packets since everything past hardware probably doesn't have a hardcoded device ID.

AzureMarker:
> Found this on the libctru docs. `Note that the main service for New3DS HID is ir:rst, and these two service are mutually exclusive: when one is initialized and reading data from New3DS HID, the other cannot access it.`
> 
> The example is initializing `hid` which in turn initializes `ir:rst` within `libctru`. My guess as to why it isn't getting inputs from the c-stick.

Maybe if the ir:USER service gets initialized before HID it will work. I don't have a New 3DS to test with (closest is Citra), but I'll try pushing that change to the example.

ctru-rs/src/services/srv.rs
AzureMarker: Updated

ctru-rs/src/services/ir_user.rs
```diff
@@ -0,0 +1,491 @@
...
+    service_handle: Handle,
+    shared_memory_handle: Handle,
+    shared_memory: &'static [u8],
+    shared_memory_layout: Layout,
+    recv_buffer_size: usize,
+    recv_packet_count: usize,
+}
+
+// ir:USER syscall command headers
+const REQUIRE_CONNECTION_COMMAND_HEADER: u32 = 0x00060040;
```
SteveCookTU: Could these headers be created by a `const fn` rather than hard coding? If created similarly to how `libctru` does it, it is easier to understand what the header and it's bits actually represent.

From libctru:

```
static inline u32 IPC_MakeHeader(u16 command_id, unsigned normal_params, unsigned translate_params)
{
	return ((u32) command_id << 16) | (((u32) normal_params & 0x3F) << 6) | (((u32) translate_params & 0x3F) << 0);
}
```

ctru-rs/src/services/ir_user.rs
```diff
@@ -0,0 +1,491 @@
...
+                let packet_data = |idx| -> u8 {
+                    let data_buffer_offset = offset_to_data_buffer + idx;
+                    shared_mem[header_size + data_buffer_offset % data_buffer_size]
+                };
+
+                // Find out how long the payload is (payload length is variable-length encoded)
+                let (payload_length, payload_offset) = if packet_data(2) & 0x40 != 0 {
+                    // Big payload
+                    (
+                        ((packet_data(2) as usize & 0x3F) << 8) + packet_data(3) as usize,
```
SteveCookTU, line 294: I believe this should be `((packet_data(2) as usize & 0x3F) << 6) + packet_data(3) as usize`

The first bit contains upper 6 bits rather than half of the payload size.

ctru-sys/src/lib.rs
```diff
@@ -17,3 +19,13 @@ pub use result::*;
 pub unsafe fn errno() -> s32 {
     (*__getreent())._errno
 }
+
+pub unsafe fn getThreadLocalStorage() -> *mut libc::c_void {
+    let return_value: *mut libc::c_void;
+    asm!("mrc p15, 0, {}, c13, c0, 3", out(reg) return_value);
+    return_value
+}
```
ian-h-chamberlain: Huh, I thought this was exposed some other way but I guess it's more of an implementation detail of `libctru` for the most part... 

I guess it's fine to expose like this since these are `unsafe` and part of `-sys` but it still feels a little funky to me, lol

ctru-rs/src/services/ir_user.rs
```diff
@@ -0,0 +1,491 @@
...
+                ResultCode(ctru_sys::srvGetServiceHandle(
+                    &mut service_handle,
+                    service_name.as_ptr(),
+                ))?;
+
+                // Calculate the shared memory size.
+                // Shared memory length must be page aligned.
+                let minimum_shared_memory_len =
+                    SHARED_MEM_INFO_SECTIONS_SIZE + recv_buffer_size + send_buffer_size;
+                let shared_memory_len = round_up(minimum_shared_memory_len, PAGE_SIZE);
```
ian-h-chamberlain, line 84: I would have guessed that `Layout` + `alloc` would do the rounding for us here, is that not the case?

ctru-rs/src/services/ir_user.rs
```diff
@@ -0,0 +1,491 @@
...
+                // Set up our service state
+                let user_state = IrUserState {
+                    service_handle,
+                    shared_memory_handle,
+                    shared_memory,
+                    shared_memory_layout,
+                    recv_buffer_size,
+                    recv_packet_count,
+                };
+                *IR_USER_STATE.lock().unwrap() = Some(user_state);
```
ian-h-chamberlain: could this `unwrap` instead be translated to an `Err` in the return value of this function?

ctru-rs/src/services/ir_user.rs
```diff
@@ -0,0 +1,491 @@
...
+        let shared_mem_guard = IR_USER_STATE.lock().unwrap();
+        let shared_mem = shared_mem_guard.as_ref().unwrap().shared_memory;
+
+        IrUserStatusInfo {
+            recv_err_result: i32::from_ne_bytes([
+                shared_mem[0],
+                shared_mem[1],
+                shared_mem[2],
+                shared_mem[3],
+            ]),
```
ian-h-chamberlain: `bytemuck::from_bytes(&shared_mem[..4])` might also work here, if it uses native endianness like this.

Actually, looking at how this gets parsed out it might be feasible to [`#[derive(AnyBitPattern)]`](https://docs.rs/bytemuck/latest/bytemuck/trait.AnyBitPattern.html) on `IrUserStatusInfo` and cast it directly from the bytes. 

I guess this would require making `bytemuck` a true dependency instead of a dev dependency but maybe it's worth it?

ctru-rs/src/services/ir_user.rs
```diff
@@ -0,0 +1,491 @@
...
+                    // Small payload
+                    ((packet_data(2) & 0x3F) as usize, 3)
+                };
+
+                // Check our payload length math against what the packet info contains
+                assert_eq!(data_length, payload_offset + payload_length + 1);
+
+                // IR packets start with a magic number, so double check it
+                let magic_number = packet_data(0);
+                assert_eq!(magic_number, 0xA5);
```
ian-h-chamberlain: Is this assertion assuming that the OS would have discarded if the magic didn't match? Otherwise (especially since this is coming from a physical input device), it might be safer to return an error here?

ctru-rs/src/services/ir_user.rs
```diff
@@ -0,0 +1,491 @@
...
+    ) -> crate::Result<Vec<u32>> {
+        let mut shared_mem_guard = IR_USER_STATE.lock().unwrap();
+        let shared_mem = shared_mem_guard.as_mut().unwrap();
+
+        // Set up the request
+        let cmd_buffer = unsafe {
+            &mut *(slice_from_raw_parts_mut(
+                ctru_sys::getThreadCommandBuffer(),
+                max(request.len(), expected_response_len),
+            ))
```
ian-h-chamberlain: Is it sound to create an `&mut[T]` from a pointer like this (since theoretically other parts of the application could modify it while the `&mut` is held)?

Example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=bab65a45e045afd8c3f22bc9d5322149
Both `y` and `z` claim unique `&mut` access to the slice at the same time which I think is technically undefined behavior (works in this case but not always).

I guess we might make similar assumptions in other places sometimes but I'm not positive that we should... `ptr::copy` on the other hand I think should normally work fine.

ctru-rs/src/services/ir_user.rs
```diff
@@ -0,0 +1,491 @@
...
+}
+
+// Internal helper for rounding up a value to a multiple of another value.
+fn round_up(value: usize, multiple: usize) -> usize {
+    if value % multiple != 0 {
+        (value / multiple) * multiple + multiple
+    } else {
+        (value / multiple) * multiple
+    }
+}
```
ian-h-chamberlain, line 363: While working on the `libctru` thread-local PR I learned a neat bit twiddling trick for aligning: https://github.com/devkitPro/libctru/pull/504/files#diff-06cc21a44744e726cdc344183858d0f060e0fc6ff4355d1a1290a068c6d9c6c9R58

This version is probably more readable but maybe slightly less efficient since it has a branch in it (depending how good the optimizer is)

ctru-rs/src/services/ir_user.rs
```diff
@@ -0,0 +1,491 @@
...
+        match *self {
+            IrDeviceId::CirclePadPro => 1,
+            IrDeviceId::Custom(id) => id,
+        }
+    }
+}
+
+/// This struct holds a parsed copy of the ir:USER service status (from shared memory).
+#[derive(Debug)]
+pub struct IrUserStatusInfo {
```
ian-h-chamberlain, line 387: Do these structs need to have `#[repr(C)]` to be compatible with the underlying OS and/or `libctru`?

ctru-rs/examples/ir-user.rs
```diff
@@ -0,0 +1,251 @@
...
+        loop {
+            hid.scan_input();
+            if hid.keys_held().contains(KeyPad::KEY_START) {
+                return ConnectionResult::Canceled;
+            }
+
+            // Send a request for input to the CPP
+            if let Err(e) = self
+                .ir_user
+                .request_input_polling(CPP_CONNECTION_POLLING_PERIOD_MS)
```
ian-h-chamberlain: it might be nice to have these APIs take a `std::time::Duration` instead, and convert (or fail) to ms internally?

ctru-rs/examples/ir-user.rs
```diff
@@ -0,0 +1,251 @@
...
+                .connection_status_event
+                .wait_for_event(Duration::from_millis(100))
+            {
+                if !e.is_timeout() {
+                    panic!("Couldn't initialize circle pad pro connection: {e}");
+                }
+            }
+
+            self.print_status_info();
+            if self.ir_user.get_status_info().connection_status == 2 {
```
ian-h-chamberlain: could `2` be defined (in `ir_user`, I suppose) as a `#[repr(u8)] enum Status` or something like that?

ctru-rs/examples/ir-user.rs
```diff
@@ -0,0 +1,251 @@
...
+
+        ConnectionResult::Connected
+    }
+
+    fn handle_packets(&self) {
+        let packets = self.ir_user.get_packets();
+        let packet_count = packets.len();
+        let Some(last_packet) = packets.last() else { return };
+
+        // Use a buffer to avoid flickering the screen (write all output at once)
```
ian-h-chamberlain: you might also be able to simply turn on double buffering for whichever screen this is getting printed to

ian-h-chamberlain (approved):
Very nice! I didn't deep dive on all the packet processing code and had some minor nits but for the most part this looks good!

I wonder if there are any other interesting use cases for IR (honestly, I didn't even realize there was a sensor there until you posted this PR 😆)

ctru-rs/src/services/ir_user.rs
Meziu: We shouldn't. This goes against Rust's safety standards. Moreover, we should wrap these methods with a proper (even if just internal) ′svc′ wrapper.

Meziu (approved):
LGTM, just solve the suggestions and pull the latest changes in master.

Meziu:
@AzureMarker I understand if you don't have time to work on this, but this PR has been stale for a while. Do you intend on working on this? I would take care of it myself if I could, but I don't own a CirclePadPro.

AzureMarker:
Yeah I definitely plan on merging this! I've just been busy with work and life recently 🙃. Thanks for keeping up the momentum!

wheremyfoodat:
Hi! I work on https://github.com/wheremyfoodat/Panda3DS and have been having some trouble with CPP emulation. Trying to set up Rust and this fork of ctru-rs has been giving me some headaches, so is there perhaps any simple CPP test ROM you have using this? Preferably something that does output via svcOutputDebugString too.

Thanks for your work! It's nice to see someone making CPP homebrew a thing. Also ctru-rs is very cool too.

Meziu:
> Hi! I work on https://github.com/wheremyfoodat/Panda3DS and have been having some trouble with CPP emulation. 

I'm happy to see you here! I've had more than a look at your project so I'm glad to see our work has sparked your interest.

> Trying to set up Rust and this fork of ctru-rs has been giving me some headaches, so is there perhaps any simple CPP test ROM you have using this? Preferably something that does output via svcOutputDebugString too.

Sadly this PR has stagnated for quite a while and isn't up-to-date with our current build systems and integration. I would have liked to finish working on this myself, however I'm not in possession of an Old3DS with a CirclePadPro, so I can't test and develop this module.

Luckily, the fixes needed to get the example working are easy for who's used to the Rust toolchain, so I quickly built it for you.

Here's a [google drive folder](https://drive.google.com/drive/folders/1gT_w1-gMa1oE5ax4Tut2c3X70fZOEkrW?usp=sharing) with the .3dsx files.
The provided examples do not use `svcOutputDebugString`, rather printing to the `stdout`.

Again, I'm incredibly sorry as this is the most I can do. If there are further issues, I can guide you in the process of building the examples yourself (which is already very well documented in our [wiki](https://github.com/rust3ds/ctru-rs/wiki) and for which you'll need [cargo-3ds](https://github.com/rust3ds/cargo-3ds)).

wheremyfoodat:
Thank you so so much ❤❤❤

AzureMarker:
> Sadly this PR has stagnated for quite a while and isn't up-to-date with our current build systems and integration. I would have liked to finish working on this myself, however I'm not in possession of an Old3DS with a CirclePadPro, so I can't test and develop this module.

I'm really looking forward to getting time to work on this and merge it, eventually! It definitely will be merged by end of this year. I'll have more time over winter vacation :). Unfortunately lots of life stuff is higher priority than having fun with side projects :/.

Meziu:
Took the freedom of merging the `master` branch into this. At least this PR should result more usable and up-to-date than before, even though I still can't really work on it.

wheremyfoodat:
> > Sadly this PR has stagnated for quite a while and isn't up-to-date with our current build systems and integration. I would have liked to finish working on this myself, however I'm not in possession of an Old3DS with a CirclePadPro, so I can't test and develop this module.
> 
> I'm really looking forward to getting time to work on this and merge it, eventually! It definitely will be merged by end of this year. I'll have more time over winter vacation :). Unfortunately lots of life stuff is higher priority than having fun with side projects :/.

:D

ctru-sys/src/lib.rs
AzureMarker: Resolving since this is now exposed by ctru-sys automatically.

ctru-rs/src/services/ir_user.rs
AzureMarker: The size of the allocation must be a multiple of the page size, not just aligned by page size.
I'll update the comment to clarify this.

ctru-rs/src/services/ir_user.rs
AzureMarker: I'll leave it as-is since this is only used once during initialization, and readability is more important here than performance (the optimizer may even compile it down to the optimized version).

ctru-rs/src/services/ir_user.rs
AzureMarker: No, this is a copy of the data after parsing.

ctru-rs/examples/ir-user.rs
AzureMarker: I'll leave it as-is because we have a very limited range of period values allowed (one unsigned byte). Doing it this way ensures that only valid inputs are given.

AzureMarker:
Did some work on the PR, but it needs a little more still. The double buffering change seems to have actually increased the stuttering/flashing, so not sure what I'm doing wrong there.
Edit: fixed!

I'll also need to open a PR on test-runner since it depends on Console, and I had to change the signature to make double buffering work.

ctru-rs/examples/ir-user.rs
AzureMarker: Switched to use double buffering (required some refactoring of Console).

ctru-rs/examples/ir-user.rs
AzureMarker: Done

ctru-rs/src/services/ir_user.rs
AzureMarker: Switched to use `std::ptr::copy` instead of making a slice.

ctru-rs/src/services/ir_user.rs
AzureMarker: Updated. Had to add a new variant to Error.

ctru-rs/src/services/ir_user.rs
AzureMarker: Updated

ctru-rs/src/console.rs
```diff
@@ -58,12 +58,12 @@ pub enum Dimension {
 /// you can try using [`Soc::redirect_to_3dslink`](crate::services::soc::Soc::redirect_to_3dslink) while activating the `--server` flag for `3dslink` (also supported by `cargo-3ds`).
 /// More info in the [`cargo-3ds` docs](https://github.com/rust3ds/cargo-3ds#running-executables).
 #[doc(alias = "PrintConsole")]
-pub struct Console<'screen> {
+pub struct Console<'screen, S: Screen> {
     context: Box<PrintConsole>,
-    screen: RefMut<'screen, dyn Screen>,
+    screen: RefMut<'screen, S>,
 }
```
ian-h-chamberlain, line 68: I seem to remember some long discussion about why this ended up as `RefMut<dyn Screen>` rather than something like the `Console<'s, S>` you have here... I don't remember all the reasoning or necessarily have a strong opinion, but I think it should still be possible to use `dyn` here if we want, with something like this:

```rs
pub struct Console<'screen> {
    context: Box<PrintConsole>,
    screen: RefMut<'screen, dyn FlushScreen>,
}

trait FlushScreen: Flush + Swap + Screen {}
impl<T: Flush + Swap + Screen> FlushScreen for T {}
```

The only caveat I can think of might be that `new()` would then require either `RefMut<dyn FlushScreen>`, or something like `pub fn new<S: Flush + Screen + Swap>(screen: RefMut<'screen, S>)` and convert the concrete type with `RefMut::map`.

I'm not really sure what is simpler and I don't have a strong preference, although keeping it as `dyn` would probably make https://github.com/rust3ds/test-runner/pull/15 unnecessary, at least.

ctru-rs/src/console.rs
```diff
@@ -322,9 +322,38 @@ impl<'screen> Console<'screen> {
             _ => unreachable!(),
         }
     }
+
+    /// Run a function with access to the underlying screen.
+    pub fn with_screen(&mut self, action: impl FnOnce(&mut RefMut<'_, S>)) {
+        action(&mut self.screen);
+    }
```
ian-h-chamberlain: Is this still needed now that you `impl Swap for Console`?

ctru-rs/src/services/srv.rs
```diff
@@ -0,0 +1,27 @@
...
+    /// [`Error::is_timeout`] to check if the error is due to a timeout.
+    fn wait_for_event(self, timeout: Duration) -> crate::Result<()>;
+}
+
+impl HandleExt for Handle {
+    fn wait_for_event(self, timeout: Duration) -> crate::Result<()> {
+        unsafe {
+            ResultCode(ctru_sys::svcWaitSynchronization(
+                self,
+                timeout.as_nanos() as i64,
```
ian-h-chamberlain: This might be good to have a `try_into()?` for, since I think it's technically possible to pass something here that would wrap (not sure how that would behave in reality): https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=17885a20236d20e78d9a63df6aad11c7

ian-h-chamberlain (commented):
Still planning to take a more in-depth look at `ir_user` but mostly looks okay to me — although there is another discussion to be had re: the test-runner / console changes. I'll post something over on that PR / issue later on probably.

ctru-rs/src/console.rs
Meziu: Yeah, I don't like this since it opens the door for weird behaviour. `Console`  should hold complete control over the screen.

AzureMarker: Yeah I think this isn't needed anymore

ctru-rs/src/services/ir_user.rs
AzureMarker: Updated

ctru-rs/src/services/ir_user.rs
AzureMarker: I simplified this part of the code and after that I don't think it's worthwhile to add the dependency (and I'm parsing the data in a special way for the connection status so that would complicate bytemuck integration).

ctru-rs/src/console.rs
AzureMarker: Thanks, this is a better approach. The only downside is that it prevents us from using TopScreenLeft/Right as the screen type, but from looking at the libctru code, it implicitly expects the screen side to be left (from `consoleInit`: `console->frameBuffer = (u16*)gfxGetFramebuffer(screen, GFX_LEFT, NULL, NULL);`).

ctru-rs/src/services/srv.rs
AzureMarker: Updated

Jhynjhiruu:
Wow, this is very fortuitous timing - I'm putting together a project that uses citro3d, and I've done some of the work to get bindings for the texture code in nice Rust. The last thing it needs in order to be a good demonstration for using the bindings to render a proper model is C-Stick support for moving the camera around; I've made do with the gyroscope so far, but it's not ideal.

AzureMarker:
@Jhynjhiruu If you have a New 3DS you can use the ir:rrst service which is initialized by the hid component in libctru automatically. It will give the New 3DS c-stick position (roughly, just the main direction):
https://libctru.devkitpro.org/hid_8h.html#a2f80701c36e79c0640d91c788feee0b3a557de4e103c1e2c316b9e0962c221ab6

Jhynjhiruu:
> @Jhynjhiruu If you have a New 3DS you can use the ir:rrst service which is initialized by the hid component in libctru automatically. It will give the New 3DS c-stick position (roughly, just the main direction): https://libctru.devkitpro.org/hid_8h.html#a2f80701c36e79c0640d91c788feee0b3a557de4e103c1e2c316b9e0962c221ab6

I do have a New 3DS, but I also have an Old 3DS and a Circle Pad Pro, and I don't want to limit my code to just working on one of them. Plus, not having analogue controls for the camera wouldn't be so fun.

AzureMarker:
Nice! If you find any issues with this code and your Circle Pad Pro or New 3DS just let me know.

Jhynjhiruu:
> Nice! If you find any issues with this code and your Circle Pad Pro or New 3DS just let me know.

Should I implement it using this branch and report back?

AzureMarker:
Yeah that should work

Jhynjhiruu:
> Yeah that should work

Alright, I'll give it a go

ctru-rs/src/console.rs
ian-h-chamberlain: yeah, AFAIK the console is not intended to work with left/right or anything (basically just wide mode or standard). I think at one point I asked the devkitPro folks about supporting buffering / more advanced rendering (e.g. for TUI roguelikes or something), and they were not interested in supporting those kinds of use cases (which prompted me to try myself with https://github.com/ian-h-chamberlain/console_3ds although I have since stopped working on it).

Jhynjhiruu:
Interesting; I appear to have written code that works completely fine on my real Circle Pad Pro but doesn't support my New 3DS. Odd.

Jhynjhiruu:
I'm not entirely sure what's going wrong here. IrUser is initialised before Hid, and yet it still doesn't receive any data from the C-Stick. My code also seems to be running very slowly, so it almost seems like one of the delays is wrong.

SteveCookTU:
> I'm not entirely sure what's going wrong here. IrUser is initialised before Hid, and yet it still doesn't receive any data from the C-Stick. My code also seems to be running very slowly, so it almost seems like one of the delays is wrong.

Unless there is another source, 3dbrew mentions that only one can read from it when both are initialized but nothing about the order. `ir:rst` may override completely or whichever is initialized last takes full control. This may have to be tested unless already done. (Unless you aren't using `ir:rst` at all then this can be ignored!)

Jhynjhiruu:
> > I'm not entirely sure what's going wrong here. IrUser is initialised before Hid, and yet it still doesn't receive any data from the C-Stick. My code also seems to be running very slowly, so it almost seems like one of the delays is wrong.
> 
> Unless there is another source, 3dbrew mentions that only one can read from it when both are initialized but nothing about the order. `ir:rst` may override completely or whichever is initialized last takes full control. This may have to be tested unless already done. (Unless you aren't using `ir:rst` at all then this can be ignored!)

I don't... think I'm using `ir:rst`?
Either way, the example in this PR (`ir-user-circle-pad-pro.rs`) doesn't work on my New 3DS. The input response is the same as in my project:
```rust
CirclePadProInputResponse {
    c_stick_x: 0x800,
    c_stick_y: 0x800,
    battery_level: 0x1f,
    zl_pressed: false,
    zr_pressed: false,
    r_pressed: false,
    unknown_field: 0x0,
}
```
The `ReceiveBuffer` seems to be doing something, it's just not what it's supposed to be. Comparing to my real Circle Pad Pro, the `ReceiveBuffer` seems to update much more slowly, and the data seems to be `f3 a5 20 06 10 00 08 80 ff 00` on repeat.
I notice that the real CPP's `IrUserStatusInfo.connection_role` is 1, whereas on my New 3DS it's 2. Otherwise, all of the info appears to be identical or equivalent. I'm a bit stumped.

SteveCookTU:
According to a comment in the example:

```
// Initialize HID after ir:USER because libctru also initializes ir:rst,
// which is mutually exclusive with ir:USER. Initializing HID before ir:USER
// on New 3DS causes ir:USER to not work.
```

I haven't tested the example yet on N3DS myself to see if there are any issues. Seems like complications will be particular to the N3DS though so playing around with initialization ordering or other small things may offer ideas since this doesn't seem to be too well documented.

Jhynjhiruu:
> According to a comment in the example:
> 
> ```
> // Initialize HID after ir:USER because libctru also initializes ir:rst,
> // which is mutually exclusive with ir:USER. Initializing HID before ir:USER
> // on New 3DS causes ir:USER to not work.
> ```
> 
> I haven't tested the example yet on N3DS myself to see if there are any issues. Seems like complications will be particular to the N3DS though so playing around with initialization ordering or other small things may offer ideas since this doesn't seem to be too well documented.

My code does initialise HID after ir:USER, I think, and the example definitely does. Neither works on my console.

AzureMarker:
Have you tried not initializing HID via libctru at all?

Jhynjhiruu:
> Have you tried not initializing HID via libctru at all?

No, I haven't, since then I wouldn't have an easy way to exit the app. I'll try it though.

Jhynjhiruu:
> > Have you tried not initializing HID via libctru at all?
> 
> No, I haven't, since then I wouldn't have an easy way to exit the app. I'll try it though.

Nope, didn't work. Still doesn't work.

Meziu:
> > Have you tried not initializing HID via libctru at all?
>>
> Nope, didn't work. Still doesn't work.

Friendly reminder that `libctru` initialises a handle to `hid` by default [here](https://github.com/devkitPro/libctru/blob/d0936b879bd2088cae61b69707fe70cb66ede651/libctru/source/system/appInit.c#L16).

You may have to close Hid or do something else before being able to use the service properly.

Jhynjhiruu:
> > > Have you tried not initializing HID via libctru at all?
> > 
> > 
> > Nope, didn't work. Still doesn't work.
> 
> Friendly reminder that `libctru` initialises a handle to `hid` by default [here](https://github.com/devkitPro/libctru/blob/d0936b879bd2088cae61b69707fe70cb66ede651/libctru/source/system/appInit.c#L16).
> 
> You may have to close Hid or do something else before being able to use the service properly.

I had totally forgotten about that!
However, disabling this (by defining a new `#[no_mangle] unsafe extern "C" fn __appInit()`, which calls `srvInit()` and nothing else) _still_ doesn't fix the issue. Eh??

Meziu:
> > > > Have you tried not initializing HID via libctru at all?
> > > 
> > > 
> > > Nope, didn't work. Still doesn't work.
> > 
> > 
> > Friendly reminder that `libctru` initialises a handle to `hid` by default [here](https://github.com/devkitPro/libctru/blob/d0936b879bd2088cae61b69707fe70cb66ede651/libctru/source/system/appInit.c#L16).
> > You may have to close Hid or do something else before being able to use the service properly.
> 
> I had totally forgotten about that! However, disabling this (by defining a new `#[no_mangle] unsafe extern "C" fn __appInit()`, which calls `srvInit()` and nothing else) _still_ doesn't fix the issue. Eh??

Here, found it. You need to re-implement this weak function https://github.com/devkitPro/libctru/blob/d0936b879bd2088cae61b69707fe70cb66ede651/libctru/source/services/hid.c#L35. If you set the return to false  any use of `hid` will not use `iirst`.

Jhynjhiruu:
> > > > > Have you tried not initializing HID via libctru at all?
> > > > 
> > > > 
> > > > Nope, didn't work. Still doesn't work.
> > > 
> > > 
> > > Friendly reminder that `libctru` initialises a handle to `hid` by default [here](https://github.com/devkitPro/libctru/blob/d0936b879bd2088cae61b69707fe70cb66ede651/libctru/source/system/appInit.c#L16).
> > > You may have to close Hid or do something else before being able to use the service properly.
> > 
> > 
> > I had totally forgotten about that! However, disabling this (by defining a new `#[no_mangle] unsafe extern "C" fn __appInit()`, which calls `srvInit()` and nothing else) _still_ doesn't fix the issue. Eh??
> 
> Here, found it. You need to re-implement this weak function https://github.com/devkitPro/libctru/blob/d0936b879bd2088cae61b69707fe70cb66ede651/libctru/source/services/hid.c#L35. If you set the return to false any use of `hid` will not use `iirst`.

That _still_ doesn't fix the issue. At this point I'm convinced there's a bug of some kind.
I think it might be worth adding a default `__appInit()` function somewhere in `ctru-rs` that does what I've just done, since we generally want to get our own handles to `Hid` etc. and there's not much point initialising more than we need.

wheremyfoodat:
I obtained the following log from Panda3DS, running Majora's Mask 3D, logging only `srv::GetServiceHandle` and any calls to HID and ir:user functions:

```
...
srv::getServiceHandle (Service: hid:USER, nameLength: 8, flags: 0)
HID::GetIPCHandles
HID::EnableAccelerometer
HID::EnableGyroscopeLow
HID::GetGyroscopeLowCalibrateParam
HID::GetGyroscopeLowRawToDpsCoefficient
HID::GetGyroscopeLowCalibrateParam
HID::GetGyroscopeLowRawToDpsCoefficient
...

Later on, when loading a save and the game attempts to initialize CPP:

srv::getServiceHandle (Service: ir:USER, nameLength: 7, flags: 0)
IR:USER: InitializeIrnopShared (shared mem size = 00001000, sharedMemHandle = 2C) 
IR:USER: GetConnectionStatusEvent
IR:USER: RequireConnection (device: 1)
```

So initializing HID before ir:USER seems to be fine, but it does indeed seem that libctru initializing `ir:rst` during `hidInit` on New 3DS might be problematic.

Jhynjhiruu:
> I obtained the following log from Panda3DS, running Majora's Mask 3D, logging only `srv::GetServiceHandle` and any calls to HID and ir:user functions:
> 
> ```
> ...
> srv::getServiceHandle (Service: hid:USER, nameLength: 8, flags: 0)
> HID::GetIPCHandles
> HID::EnableAccelerometer
> HID::EnableGyroscopeLow
> HID::GetGyroscopeLowCalibrateParam
> HID::GetGyroscopeLowRawToDpsCoefficient
> HID::GetGyroscopeLowCalibrateParam
> HID::GetGyroscopeLowRawToDpsCoefficient
> ...
> 
> Later on, when loading a save and the game attempts to initialize CPP:
> 
> srv::getServiceHandle (Service: ir:USER, nameLength: 7, flags: 0)
> IR:USER: InitializeIrnopShared (shared mem size = 00001000, sharedMemHandle = 2C) 
> IR:USER: GetConnectionStatusEvent
> IR:USER: RequireConnection (device: 1)
> ```
> 
> So initializing HID before ir:USER seems to be fine, but it does indeed seem that libctru initializing `ir:rst` during `hidInit` on New 3DS might be problematic.

Don't suppose you can determine the parameters MM uses for initialising the CPP? Timings etc.
Just overriding `hidShouldUseIrrst()` to return `false` doesn't change anything, irritatingly.

wheremyfoodat:
I have a half-decompilation of the Majora's Mask IR code on my computer, though it's rather complicated. "Timing" is done via the IR events (eg once the game calls IR::RequireConnection to pair with the CPP, it uses WaitSync1 on the IR connection event. This way it sleeps until the connection attempts has ended and IR shared memory is properly initialized. After that, I think it uses the send/recv events for synchronizing package transfers)

Jhynjhiruu:
> I have a half-decompilation of the Majora's Mask IR code on my computer, though it's rather complicated. "Timing" is done via the IR events (eg once the game calls IR::RequireConnection to pair with the CPP, it uses WaitSync1 on the IR connection event. This way it sleeps until the connection attempts has ended and IR shared memory is properly initialized. After that, I think it uses the send/recv events for synchronizing package transfers)

@AzureMarker Any chance of trying something like this out?

AzureMarker:
I'm pretty sure that's how it already works. I based my implementation on Majora's Mask (even got into quite a few gdb sessions).

Jhynjhiruu:
> I'm pretty sure that's how it already works. I based my implementation on Majora's Mask (even got into quite a few gdb sessions).

Welp. Is there anything else I can do? Info to capture?

AzureMarker:
Is there any way to know if the ir:rst service is in use? Maybe something else is initializing it. Or maybe calling the close function could help kill whatever active session there might be? I don't have a New 3DS to test with.

Jhynjhiruu:
> Is there any way to know if the ir:rst service is in use? Maybe something else is initializing it. Or maybe calling the close function could help kill whatever active session there might be? I don't have a New 3DS to test with.

The handy variable that says whether `ir:rst` is in use is unfortunately static, so I can't read it. However, the handles and the pointer to shared memory aren't static, so I can read those just fine, and they're all 0s, which strongly implies that `ir:rst` isn't in use.

AzureMarker:
@ian-h-chamberlain @Meziu I think all the comments have been addressed, so let me know if we can merge the PR or if you have more comments. Thanks for taking a look! It's almost been a year since I opened the PR :sweat_smile:.

Jhynjhiruu:
> @ian-h-chamberlain @Meziu I think all the comments have been addressed, so let me know if we can merge the PR or if you have more comments. Thanks for taking a look! It's almost been a year since I opened the PR 😅.

Well, it would be nice to get the example code working on New 3DS first...

AzureMarker:
I don't have a New 3DS to test with, and it works in Citra. We can merge this and work on improving New 3DS support as a follow-up. The PR has been open for a very long time and is close to merging.

ctru-rs/src/services/svc.rs
```diff
@@ -0,0 +1,76 @@
...
+//!
+//! Not all APIs are wrapped in this module, since a lot are fundamentally unsafe.
+//! Most APIs should be used directly from `ctru-sys`.
+
+use crate::error::ResultCode;
+use ctru_sys::Handle;
+use std::time::Duration;
+
+/// Extension trait for [Handle]
+pub trait HandleExt {
```
ian-h-chamberlain, line 11: We might want to consider making this `Sealed`, but I don't think it _needs_ to be and it definitely doesn't need to block merging.

ian-h-chamberlain (approved):

ian-h-chamberlain:
> @ian-h-chamberlain @Meziu I think all the comments have been addressed, so let me know if we can merge the PR or if you have more comments. Thanks for taking a look! It's almost been a year since I opened the PR 😅.

LGTM! As you said, N3DS issues could probably be fixed in a follow up just to get this merged since it has been so long.

ctru-rs/examples/ir-user-circle-pad-pro.rs
```diff
@@ -0,0 +1,259 @@
+//! A demo of using the ir:USER service to connect to the Circle Pad Pro.
```
Meziu, line 3: The header comment should be changed to more closely resemble the other example headers (Name of the example followed by a simple description).

ctru-rs/examples/ir-user-circle-pad-pro.rs
```diff
@@ -0,0 +1,259 @@
...
+use ctru::services::svc::HandleExt;
+use ctru_sys::Handle;
+use std::time::Duration;
+
+const PACKET_INFO_SIZE: usize = 8;
+const MAX_PACKET_SIZE: usize = 32;
+const PACKET_COUNT: usize = 1;
+const PACKET_BUFFER_SIZE: usize = PACKET_COUNT * (PACKET_INFO_SIZE + MAX_PACKET_SIZE);
+const CPP_CONNECTION_POLLING_PERIOD_MS: u8 = 0x08;
+const CPP_POLLING_PERIOD_MS: u8 = 0x32;
```
Meziu, line 18: Nit: Might seem a bit too much, but you should probably specify in a comment that this configuration is specifically made for the CPP, and shouldn't be used in all IR use-cases.

ctru-rs/examples/ir-user-circle-pad-pro.rs
```diff
@@ -0,0 +1,259 @@
...
+    let apt = Apt::new().unwrap();
+    let gfx = Gfx::new().unwrap();
+    let top_console = Console::new(gfx.top_screen.borrow_mut());
+    let bottom_console = Console::new(gfx.bottom_screen.borrow_mut());
+    let mut demo = CirclePadProDemo::new(top_console, bottom_console);
+    demo.print_status_info();
+
+    // Initialize HID after ir:USER because libctru also initializes ir:rst,
+    // which is mutually exclusive with ir:USER. Initializing HID before ir:USER
+    // on New 3DS causes ir:USER to not work.
```
Meziu, line 37: Either you add the definition for the weak function, or specify that this example (currently) works only on Old3DS. Otherwise, this comment becomes false and misleading.

ctru-rs/src/services/gfx.rs
```diff
@@ -14,6 +14,7 @@ use crate::services::ServiceReference;
 
 mod private {
     use super::{BottomScreen, TopScreen, TopScreen3D, TopScreenLeft, TopScreenRight};
+    use crate::console::Console;
```
Meziu: Nit: If we are planning on putting more structs into this field, we might have to consider moving this into `lib.rs` or into a separate module.

ctru-rs/src/services/ir_user.rs
```diff
@@ -0,0 +1,496 @@
...
+                2,
+            )?;
+        }
+
+        Ok(())
+    }
+
+    /// Mark the last `packet_count` packets as processed, so their memory in
+    /// the receive buffer can be reused.
+    pub fn release_received_data(&self, packet_count: u32) -> crate::Result<()> {
```
Meziu: Should this function require `&mut self`? Anything that changes internal states should require `mut`.

ctru-rs/src/services/ir_user.rs
```diff
@@ -0,0 +1,496 @@
...
+            self.send_service_request(
+                vec![REQUIRE_CONNECTION_COMMAND_HEADER, device_id.get_id()],
+                2,
+            )?;
+        }
+        Ok(())
+    }
+
+    /// Close the current IR connection.
+    pub fn disconnect(&self) -> crate::Result<()> {
```
Meziu: Must require `&mut self`.

ctru-rs/src/services/ir_user.rs
```diff
@@ -0,0 +1,496 @@
...
+            },
+        )?;
+
+        Ok(IrUser {
+            _service_reference: service_reference,
+        })
+    }
+
+    /// Try to connect to the device with the provided ID.
+    pub fn require_connection(&self, device_id: IrDeviceId) -> crate::Result<()> {
```
Meziu: Must require `&mut self`.

Meziu (changes requested):
Nice PR, this is very much ready for merging!

I've also done a bit of testing with `ir:rst` and `ir:USER` on the side, and I can say a couple of things to help the discussion.

First of all, the `ir:rst` service is NOT initialized when implementing the `libctru` weak function as such:
```rust
#[no_mangle]
unsafe extern "C" fn hidShouldUseIrrst() -> bool {
    false
}
```
I could test various uses of the `irrst` service and they actually stopped working after using this implementation.

Secondly... to me this function seems to make it work! Not for long though. I can clearly see the C-Stick readings to be correct for some fraction of a second (not much more than 6 packets in a row), right before dying out and never reading anything again. This behavior (which is quite consistent, it seems) clearly means that there is something else missing for the implementation to work correctly, though I'm pretty much unable to read any further into it.

This PR can be merged, but you should open an issue since more research needs to be made.

ctru-rs/src/services/svc.rs
AzureMarker: I don't think it needs to be sealed, we only use it to implement more functions (we don't take it as input).

ctru-rs/examples/ir-user-circle-pad-pro.rs
AzureMarker: Added the function export

ctru-rs/src/services/gfx.rs
AzureMarker: Moved to separate file

Meziu (approved):

Meziu:
@AzureMarker The PR is ready, but I want you to press the Merge button 😁

AzureMarker:
Opened https://github.com/rust3ds/ctru-rs/issues/151

PR #92, opened 2023-02-14
Remove init function

Meziu:
Removed the ´ctru::init´ functions since it's not needed for linking anymore!

Other changes include:
- Renamed the panic handler code (I want it to be optional for the user and not included in some pre-made package)
- Added ´new´ and ´drop´ for the ´Ps´struct due to https://github.com/rust3ds/rust-linker-fix-3ds/pull/16

### DIFF
--- a/ctru-rs/Cargo.toml
+++ b/ctru-rs/Cargo.toml
@@ -15,8 +15,8 @@ name = "ctru"
 cfg-if = "1.0"
 ctru-sys = { path = "../ctru-sys", version = "0.4" }
 const-zero = "0.1.0"
-linker-fix-3ds = { path = "../../rust-linker-fix-3ds" }
-pthread-3ds = { path = "../../pthread-3ds" }
+linker-fix-3ds = { git = "https://github.com/rust3ds/rust-linker-fix-3ds.git" }
+pthread-3ds = {git = "https://github.com/rust3ds/pthread-3ds.git" }
 libc = "0.2.121"
 bitflags = "1.0.0"
 widestring = "0.2.2"
--- a/ctru-rs/Cargo.toml
+++ b/ctru-rs/Cargo.toml
@@ -15,8 +15,8 @@ name = "ctru"
 cfg-if = "1.0"
 ctru-sys = { path = "../ctru-sys", version = "0.4" }
 const-zero = "0.1.0"
-linker-fix-3ds = { git = "https://github.com/rust3ds/rust-linker-fix-3ds.git" }
-pthread-3ds = { git = "https://github.com/rust3ds/pthread-3ds.git" }
+linker-fix-3ds = { path = "../../rust-linker-fix-3ds" }
+pthread-3ds = { path = "../../pthread-3ds" }
 libc = "0.2.121"
 bitflags = "1.0.0"
 widestring = "0.2.2"
--- a/ctru-rs/Cargo.toml
+++ b/ctru-rs/Cargo.toml
@@ -16,7 +16,7 @@ cfg-if = "1.0"
 ctru-sys = { path = "../ctru-sys", version = "0.4" }
 const-zero = "0.1.0"
 linker-fix-3ds = { git = "https://github.com/rust3ds/rust-linker-fix-3ds.git" }
-pthread-3ds = {git = "https://github.com/rust3ds/pthread-3ds.git" }
+pthread-3ds = { git = "https://github.com/rust3ds/pthread-3ds.git" }
 libc = "0.2.121"
 bitflags = "1.0.0"
 widestring = "0.2.2"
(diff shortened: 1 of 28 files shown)


ctru-rs/src/test_runner.rs
```diff
@@ -15,8 +15,6 @@ use crate::services::Apt;
 /// runs all tests in series, "failing" on the first one to panic (really, the
 /// panic is just treated the same as any normal application panic).
 pub(crate) fn run(tests: &[&TestDescAndFn]) {
-    crate::init();
```
ian-h-chamberlain, line 18: we probably want to keep panic setup here, otherwise it will be hard to read output for any failing test

ctru-rs/src/lib.rs
```diff
@@ -7,40 +7,28 @@
...
-    unsafe {
-        ctru_sys::psExit();
-    }
-}
+// These functions are imported to assure ´cargo´ we need to link the crates.
+// These don't need to run to link the functions properly!
+#[allow(unused_imports)]
+use linker_fix_3ds::init as link_init;
+#[allow(unused_imports)]
+use pthread_3ds::init as pthread_init;
```
ian-h-chamberlain: could these perhaps be 
```rs
extern crate linker_fix_3ds;
extern crate pthread_3ds;
```

I guess I'm not sure exactly what criteria `cargo` uses to decide to discard an entire crate during linking...

I wonder if [this Cargo feature](https://rust-lang.github.io/rfcs/3028-cargo-binary-dependencies.html) would ideally allow us to depend on them as `artifact = "staticlib"` so they participate in linking, but avoid any extra stuff in the code?

ctru-rs/src/services/ps.rs
```diff
@@ -30,23 +30,44 @@ pub enum AESKeyType {
     Keyslot39Nfc,
 }
 
-pub fn local_friend_code_seed() -> crate::Result<u64> {
-    let mut seed: u64 = 0;
+pub struct Ps(());
```
ian-h-chamberlain, line 33: I'm assuming you decided we don't need the service reference machinery here because of the atomic counter internal to `ps` ?

ian-h-chamberlain (approved):

ctru-rs/src/services/ps.rs
Meziu: The `ServiceReference` is only useful for services that shouldn’t be initialised more than once at a time (especially in multi-threaded programs). `Ps` is an extremely simple service that just give access to a couple of functions and keeps no internal state. There shouldn’t be much of an issue.

ctru-rs/src/test_runner.rs
Meziu: The panic setup isn’t used for tests. The testing suite already implements similar code, mainly to read info about the test execution. You should know that, you wrote it! 😂

ctru-rs/src/lib.rs
Meziu: I didn’t try just importing the crate, I forgot about that possibility. Though yes, cargo seems to not compile/pass to the linker crates which have zero references. I imported those functions just because they existed already.

ctru-rs/src/test_runner.rs
ian-h-chamberlain: I thought that we specifically called `init()` here to enforce the custom panic setup, particularly by the comment:

> (really, the panic is just treated the same as any normal application panic).

But I guess I'm remembering now that the output + panic get captured by the runner, then dumped when there's a failure (then loop and wait for user input). It has been a while since I wrote this stuff 😄

Meziu: I’m pretty sure we had an `init` there just for the `ps` stuff. Don’t worry, I was just teasing 😜

ctru-rs/Cargo.toml
```diff
@@ -16,7 +16,7 @@ cfg-if = "1.0"
 ctru-sys = { path = "../ctru-sys", version = "0.4" }
 const-zero = "0.1.0"
 linker-fix-3ds = { git = "https://github.com/rust3ds/rust-linker-fix-3ds.git" }
-pthread-3ds = { git = "https://github.com/rust3ds/pthread-3ds.git" }
```
AzureMarker: nit: this change can be reverted

AzureMarker (approved):

ctru-rs/src/lib.rs
```diff
@@ -7,12 +7,9 @@
...
 
-// These functions are imported to assure ´cargo´ we need to link the crates.
-// These don't need to run to link the functions properly!
-#[allow(unused_imports)]
-use linker_fix_3ds::init as link_init;
-#[allow(unused_imports)]
-use pthread_3ds::init as pthread_init;
+// Nothing is imported from these crates but their inclusion here assures correct linking of the missing implementations.
+extern crate linker_fix_3ds;
+extern crate pthread_3ds;
```
ian-h-chamberlain, line 12: 🎊

Meziu: 🥳

ian-h-chamberlain (approved):

PR #94, opened 2023-02-17
Fix non-working example and improve the camera module

Meziu:
The camera-image example didn't work (it was just a typo in the timeout wait). I took the chance to improve performance and catch some errors.

Changes:
- No more conversion from RGB565 to RGB8 (now we only rotate the image directly to the framebuffer).
- The example now only flushes the image once it gets shot, and not every frame (CPU rendering created a lot of latency).
- Removed all needless allocations. We allocated around 384000 bytes _**each time**_ a photo was shot. Now we allocate only once at the start.
- The `svc` timeout wasn't considered an error (it took me some time to debug it). We should probably handle those cases in a separate `svc` module PR (which we were already talking about), but for now it correctly returns the error.
- I've also documented some edge cases that cause ARM exceptions when using `svc`, it was a learning experience!

Tell me what you think 😄

### DIFF
--- a/ctru-rs/examples/camera-image.rs
+++ b/ctru-rs/examples/camera-image.rs
@@ -9,7 +9,7 @@ const WIDTH: usize = 400;
 const HEIGHT: usize = 240;
 
 // The screen size is the width and height multiplied by 2 (RGB565 store pixels in 2 bytes)
-const BUF_SIZE: usize = WIDTH * HEIGHT * 2;
+// const BUF_SIZE: usize = WIDTH * HEIGHT * 2;
 
 const WAIT_TIMEOUT: Duration = Duration::from_micros(300);
 
@@ -55,9 +55,8 @@ fn main() {
             .set_trimming(false)
             .expect("Failed to disable trimming");
     }
-    let mut buf = vec![128u8; BUF_SIZE / 2];
-    let mut lol = vec![255u8; BUF_SIZE / 2];
-    buf.append(&mut lol);
+
+    let mut buf;
 
     println!("\nPress R to take a new picture");
     println!("Press Start to exit to Homebrew Launcher");
@@ -72,8 +71,6 @@ fn main() {
 
         if keys_down.contains(KeyPad::KEY_R) {
             println!("Capturing new image");
-            cam.play_shutter_sound(CamShutterSoundType::NORMAL)
-                .expect("Failed to play shutter sound");
 
             let camera = &mut cam.outer_right_cam;
 
@@ -84,21 +81,24 @@ fn main() {
                     WAIT_TIMEOUT,
                 )
                 .expect("Failed to take picture");
+
+            cam.play_shutter_sound(CamShutterSoundType::NORMAL)
+                .expect("Failed to play shutter sound");
+
+            let img = rotate_image(&buf, WIDTH, HEIGHT);
+
+            unsafe {
+                gfx.top_screen
+                    .borrow_mut()
+                    .get_raw_framebuffer()
+                    .ptr
+                    .copy_from(img.as_ptr(), img.len());
+            }
+
+            gfx.flush_buffers();
+            gfx.swap_buffers();
+            gfx.wait_for_vblank();
         }
-
-        let img = rotate_image(&buf, WIDTH, HEIGHT);
-
-        unsafe {
-            gfx.top_screen
-                .borrow_mut()
-                .get_raw_framebuffer()
-                .ptr
-                .copy_from(img.as_ptr(), img.len());
-        }
-
-        gfx.flush_buffers();
-        gfx.swap_buffers();
-        gfx.wait_for_vblank();
     }
 }
 
--- a/ctru-rs/examples/camera-image.rs
+++ b/ctru-rs/examples/camera-image.rs
@@ -1,16 +1,15 @@
-use ctru::console::Console;
-use ctru::gfx::{Gfx, Screen};
+use ctru::gfx::Screen;
+use ctru::prelude::*;
 use ctru::services::cam::{Cam, CamOutputFormat, CamShutterSoundType, CamSize, Camera};
-use ctru::services::hid::KeyPad;
-use ctru::services::{Apt, Hid};
+use ctru::services::gspgpu::FramebufferFormat;
+
 use std::time::Duration;
 
 const WIDTH: usize = 400;
 const HEIGHT: usize = 240;
 
-// The screen size is the width and height multiplied by 2 and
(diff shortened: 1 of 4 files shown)


ctru-rs/examples/camera-image.rs
```diff
@@ -69,69 +71,58 @@ fn main() {
...
-            rgb8[draw_index + 2] = b;
+            let read_index = (j * width + i) * 2;
+
+            // Initial index of where to draw in the frame buffer based on y and x coordinates
+            let draw_index = (draw_x * height + draw_y) * 2; // This 2 stands for the number of bytes per pixel (16 bits)
+
+            unsafe {
+                let pixel_pointer = framebuf.offset(draw_index as isize);
+                *pixel_pointer = src[read_index];
+                *pixel_pointer.offset(1) = src[read_index + 1];
```
ian-h-chamberlain: tiny nit: this could be `pixel_pointer.copy_from(src.as_ptr(), 2)` or something similer with `std::ptr::copy()`

ctru-rs/src/services/cam.rs
```diff
@@ -744,6 +745,11 @@ pub trait Camera {
             ))?;
         };
 
+        let screen_size = u32::from(width) * u32::from(height) * 2;
+        if buffer.len() < screen_size as usize {
+            panic!("Provided buffer's length is shorter than the desired width and height matrix.")
```
ian-h-chamberlain: Could it maybe just return an `Err` here instead of panicking (since this function already returns `Result`)?

ctru-rs/src/services/cam.rs
```diff
@@ -754,7 +760,7 @@ pub trait Camera {
             let mut completion_handle: Handle = 0;
             ResultCode(ctru_sys::CAMU_SetReceiving(
                 &mut completion_handle,
-                buf.as_mut_ptr() as *mut ::libc::c_void,
+                buffer.as_mut_ptr() as *mut ::libc::c_void,
```
ian-h-chamberlain: nit: I think this can be `buffer.as_mut_ptr().cast()`

ian-h-chamberlain (approved):
Nothing major, just small style type stuff I think.

ctru-rs/src/services/cam.rs
Meziu: That was my initial approach as well, but the Result returns only our own `Error` type, so I’d need a new enum variant. I’ll see what we can do 👍

Meziu: Made a new error variant with some info.

ctru-rs/src/error.rs
```diff
@@ -55,6 +55,8 @@ pub enum Error {
     Libc(String),
     ServiceAlreadyActive,
     OutputAlreadyRedirected,
+    /// The first member is the length of the buffer provided by the user, the second parameter is the size of the requested data (in bytes).
+    BufferTooShort(usize, usize),
```
ian-h-chamberlain, line 59: tiny nit, if you want to make these values self-describing this could be something like
```rs
BufferTooShort {
    provided: usize,
    wanted: usize,
}
```

ctru-rs/src/error.rs
```diff
@@ -119,8 +126,9 @@ impl fmt::Display for Error {
             Self::Libc(err) => write!(f, "{err}"),
             Self::ServiceAlreadyActive => write!(f, "Service already active"),
             Self::OutputAlreadyRedirected => {
-                write!(f, "output streams are already redirected to 3dslink")
+                write!(f, "Output streams are already redirected to 3dslink")
```
ian-h-chamberlain: I guess this is personal preference but I would normally default to start error messages with lowercase, in case the user wanted to use it as part of a sentence:
```rs
println!("Internal error: {err}"); // "Internal error: output streams already redirected"
```
That said maybe a user that really cares about the string representation would write their own custom format handler anyway so this isn't a big deal IMO

ian-h-chamberlain (approved):

ctru-rs/src/error.rs
Meziu: It seems like Rust’s `std` uses lowercase error strings. I believe we should probably stand by that standard. Thanks for the suggestion.

ctru-rs/src/error.rs
Meziu: ~~I don't believe this is correct Rust code (?). The struct would just be a member of the enum, and **_not_** a variant of the aforementioned. I've seen some postponed RFCs about this kind of use, but it's not yet available.~~

Edit: I had just read it wrong, I'm sorry for the misunderstanding.

Techie-Pi: The change @ian-h-chamberlain suggested is correct Rust code (unless I don't understand what you're referring to). [example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e8908f1f4219776f8c28aa99fafb9f56)

Meziu: Oops, yeah i worded it wrongly. IT IS correct rust code, it's just not what we want it to do, I think (it'd be a different type, not a variant of the enum).

Techie-Pi: If I'm not wrong, both ``BufferTooShort(usize, usize)`` and ``BufferTooShort { provided: usize, wanted: usize }`` are variants of the enum. (Or maybe I still don't understand where is the problem 😅)

---
[From The Rust Reference](https://doc.rust-lang.org/reference/items/enumerations.html):

Enum constructors can have either named or unnamed fields:

```rust
enum Animal {
    Dog(String, f64),
    Cat { name: String, weight: f64 },
}

let mut a: Animal = Animal::Dog("Cocoa".to_string(), 37.2);
a = Animal::Cat { name: "Spotty".to_string(), weight: 2.7 };
```
In this example, Cat is a _struct-like **enum variant**_, whereas Dog is simply called an enum variant.

Meziu: You are absolutely correct, I had just misread a "struct" before the variant, which is definitely not there. Maybe I'm just tired 😅

AzureMarker: Catching up on some PRs, is there a reason this change wasn't incorporated? I see the `Debug` implementation pretends like it has named fields.

Techie-Pi: It was implemented, but it was directly pushed to main after the PR was merged (unless I don't remember correctly)

Meziu: Yeah it’s in `master`. I don’t really remember why it doesn’t show here. It’s possible I pushed it afterwards.

PR #95, opened 2023-02-28
Add a basic AM service wrapper and an example that uses it

Maccraft123:
This PR adds basic AM wrapper, for use by programs that read title data, and a basic example.
I was unsure where to put FsMediaType enum, initially it was in am.rs, but in libctru it was defined in fs.h, so it went into fs.rs
I'm also unsure about the prelude, whether Am struct and FsMediaType enum should be added to it.
As it usually makes sense to get all titles installed, instead of random n of them, get_title_list returns a Vec of all title ids on a given FsMediaType.
Should TitleId be a newtype or struct so that it could have some helper methods like TitleId::upper() etc?

This PR was tested with the 'title-info' example on a New 3DS with Luma3DS v11.0 and system firmware v11.16.0-49E

### DIFF
--- a/ctru-rs/examples/title-info.rs
+++ b/ctru-rs/examples/title-info.rs
@@ -59,18 +59,18 @@ fn main() {
         }
 
         if refresh {
-            let mut selected_title = 0;
+            let mut selected_title = cur_list.iter().skip(offset).next().unwrap();
             // Clear top screen and write title ids to it
             top_screen.select();
             print!("\x1b[2J");
 
             // Top screen seems to have only 30 rows
-            for (i, id) in cur_list.iter().skip(offset).take(29).enumerate() {
+            for (i, title) in cur_list.iter().skip(offset).take(29).enumerate() {
                 if i == 0 {
-                    selected_title = *id;
-                    println!("=> {id:x}");
+                    selected_title = title;
+                    println!("=> {:x}", title.id());
                 } else {
-                    println!("   {id:x}");
+                    println!("   {:x}", title.id());
                 }
             }
 
@@ -79,20 +79,18 @@ fn main() {
             println!("\x1b[2J");
             // Move cursor to top left
             println!("\x1b[1;1");
-            let media = if use_nand {
-                FsMediaType::Nand
-            } else {
-                FsMediaType::Sd
-            };
-            match am.get_title_info(media, &mut [selected_title]) {
+
+            match selected_title.get_title_info() {
                 Ok(info) => {
-                    // Vec returned by Am::get_title_info always has same length as inputed slice
-                    let info = info[0];
                     println!("Size: {} KB", info.size_bytes() / 1024);
                     println!("Version: 0x{:x}", info.version());
                     println!("Type: 0x{:x}", info.type_());
                 }
                 Err(e) => println!("Failed to get title info: {}", e),
+            }
+            match selected_title.get_product_code() {
+                Ok(code) => println!("Product code: \"{code}\""),
+                Err(e) => println!("Failed to get product code: {}", e),
             }
 
             println!("\x1b[26;0HPress START to exit");
--- a/ctru-rs/examples/title-info.rs
+++ b/ctru-rs/examples/title-info.rs
@@ -9,19 +9,26 @@ fn main() {
     let hid = Hid::init().expect("Couldn't obtain HID controller");
     let apt = Apt::init().expect("Couldn't obtain APT controller");
     let am = Am::init().expect("Couldn't obtain AM controller");
-    let _console = Console::init(gfx.top_screen.borrow_mut());
+    let top_screen = Console::init(gfx.top_screen.borrow_mut());
+    let bottom_screen = Console::init(gfx.bottom_screen.borrow_mut());
 
-    let title_count = am
+    let sd_count = am
         .get_title_count(FsMediaType::Sd)
-        .expect("Failed to get title count");
-    println!("This 3DS has {title_count} titles on its SD Card:");
-
-    let title_list = am
+        .expect("Failed to get sd title count");
+    let sd_list = am
         .get_title_list(FsMediaType::Sd)
-        .expect("Failed to get title list");
-    for id in title_list {
-        println!("{id:x}");
-    }
+        .expect("Failed to get sd title list");
+
+    let nand_count = am
+        .get_title_count(FsMediaType::Nand)
+        .expect("Failed to get nand title count");
+    let nand_list = am
+        .get_title_list(FsMediaType::Nand)
+        .expect("Failed to get nand title list");
+
+    let mut offset = 0;
+    let mut refresh = true;
+    let mut use_nand = false;
 
     // Main loop
     while apt.main_loop() {
@@ -31,6 +38,77 @@ fn main() {
         if hid.keys_down().contains(KeyPad::KEY_START) {
             break;
         }
+        if hid.keys_down().contains(KeyPad::KEY_SELECT) {
+            refresh = true;
+            offset = 0;
+            use_nand = !use_nand;
+        }
+
+        let cur_list = if use_nand { &nand_list } else { &sd_list };
+
+        if hid.keys_down().intersects(KeyPad::KEY_DOWN) {
+            if offset + 1 < cur_list.len() {
+                offset = offset + 1;
+                refresh = true;
+            }
+        } else if hid.keys_down().intersects(KeyPad::KEY_UP) {
+            if offset > 0 {
+                offset = offset - 1;
+                refresh = true;
+            }
+        }
+
+        if refresh {
+            let mut selected_title = 0;
+            // Clear top screen and write title ids to it
+            top_screen.select();
+            print!("\x1b[2J");
+
+            // Top screen seems to have only 30 rows
+            for (i, id) in cur_list.iter().skip(offset).take(29).enumerate() {
+                if i == 0 {
+                    selected_title = *id;
+                    println!("=> {id:x}");
+                } else {
+                    println!("   {id:x}");
+                }
+            }
+
+            // Clear bottom screen and write properties of selected title to it
+            bottom_screen.select();
+            println!("\x1b[2J");
+            // Move cursor to top left
+            println!("\x1b[1;1");
+            let media = if use_nand {
+                FsMediaType::Nand
+            } else {
+                FsMediaType::Sd
+            };
+            match am.get_title_info(media, &mut [selected_title]) {
+                Ok(info) => {
+                    // Vec returned by Am::get_title_info always has same length as inputed slice
+                    let info = info[0];
+                    println!("Size: {} KB", info.size_bytes() / 1024);
+                    println!("Version: 0x{:x}", info.version());
+                    println!("Type: 0x{:x}", info.type_());
+                }
+                Err(e) => println!("Failed to get title info: {}", e),
+            }
+
+            println!("\x1b[26;0HPress START to exit");
+            if use_nand {
+                println!("Press SELECT to choose SD Card");
+                println!("Current medium: NAND");
+                println!("Title count: {}", nand_count);
+            } else {
+                println!("Press SELECT to choose NAND");
+                println!("Current medium: SD Card");
+                println!("Title count: {}", sd_count);
+            }
+
+            refresh = false;
+        }
+
         // Flush and swap framebuffers
         gfx.flush_buffers();
         gfx.swap_buffers();
--- a/ctru-rs/examples/title-info.rs
+++ b/ctru-rs/examples/title-info.rs
@@ -84,7 +84,6 @@ fn main() {
                 Ok(info) => {
                     println!("Size: {} KB", info.size_bytes() / 1024);
                     println!("Version: 0x{:x}", info.version());
-                    println!("Type: 0x{:x}", info.type_());
                 }
                 Err(e) => println!("Failed to get title info: {}", e),
             }
--- a/ctru-rs/examples/title-info.rs
+++ b/ctru-rs/examples/title-info.rs
@@ -11,10 +11,14 @@ fn main() {
     let am = Am::init().expect("Couldn't obtain AM controller");
     let _console = Console::init(gfx.top_screen.borrow_mut());
 
-    let title_count = am.get_title_count(FsMediaType::Sd).expect("Failed to get title count");
+    let title_count = am
+        .get_title_count(FsMediaType::Sd)
+        .expect("Failed to get title count");
     println!("This 3DS has {title_count} titles on its SD Card:");
 
-    let title_list = am.get_title_list(FsMediaType::Sd).expect("Failed to get title list");
+    let title_list = am
+        .get_title_list(FsMediaType::Sd)
+        .expect("Failed to get title list");
     for id in title_list {
         println!("{id:x}");
     }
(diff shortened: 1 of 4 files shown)


ctru-rs/examples/title-info.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    let title_count = am
+        .get_title_count(FsMediaType::Sd)
+        .expect("Failed to get title count");
+    println!("This 3DS has {title_count} titles on its SD Card:");
+
+    let title_list = am
+        .get_title_list(FsMediaType::Sd)
+        .expect("Failed to get title list");
+    for id in title_list {
```
Meziu: I have a lot of downloaded titles (e-shop shutdown panic) so this loop actually covers the original message. Furthermore, it's not really useful (nor fun) for a user to get listed a bunch of hex numbers. Can we retrieve the title's names instead?

ctru-rs/src/services/am.rs
```diff
@@ -0,0 +1,43 @@
...
+            Ok(count)
+        }
+    }
+
+    pub fn get_title_list(&self, mediatype: FsMediaType) -> crate::Result<Vec<u64>> {
+        unsafe {
+            let count = self.get_title_count(mediatype)?;
+            let mut buf = Vec::with_capacity(count as usize);
+            let mut read_amount = 0;
+            ResultCode(ctru_sys::AM_GetTitleList(
```
Meziu, line 84: nit: Just put a newline before and after this `ResultCode`, it's more readable 👍

Meziu (approved):
Seems like you understood our standards well just by looking at our code! It looks really nice, I just wish there was a little more functionality added. On the other hand, looking at `libctru`'s documentation, I don't see easy/safe stuff to port, so it's not a big deal.

ctru-rs/examples/title-info.rs
Maccraft123: I tried to get the actual names and couldn't figure it out on my own(also rust3ds is my first time doing any kind of homebrew for 3ds)

ctru-rs/src/services/am.rs
Maccraft123: Done

Maccraft123:
> Seems like you understood our standards well just by looking at our code! It looks really nice, I just wish there was a little more functionality added. On the other hand, looking at `libctru`'s documentation, I don't see easy/safe stuff to port, so it's not a big deal.

Thanks!
I wanted to implement more functionality of `am`, however I don't really know how to use it beyond what I did and I don't want to push any code that I haven't tested myself.

Also, I noticed that getting lower and higher 32 bits of TitleIds is quite useful, and doing it manually every time or using a custom function is kinda ugly. What do you think about adding a TitleId tuple struct with a u64, that would also have other methods other than `lower` and `higher`, like `title_category`?

I also have PTMU wrapper in `feature/ptmu` branch, [here](https://github.com/Maccraft123/ctru-rs/commits/feature/ptmu) that does pretty much everything that libctru does, however there are some minor changes that I think would make it easier to use, but that all should just go in another PR that i'll do after this one is merged

ctru-rs/src/services/am.rs
```diff
@@ -0,0 +1,45 @@
...
+    pub fn get_title_count(&self, mediatype: FsMediaType) -> crate::Result<u32> {
+        unsafe {
+            let mut count = 0;
+            ResultCode(ctru_sys::AM_GetTitleCount(mediatype as u32, &mut count))?;
+            Ok(count)
+        }
+    }
+
+    pub fn get_title_list(&self, mediatype: FsMediaType) -> crate::Result<Vec<u64>> {
+        unsafe {
```
ian-h-chamberlain, line 83: as a minor nit, I'd generally recommend keeping the `unsafe` block as small as possible, in this case it can probably just be the calls to `AM_GetTitleList` and `buf.set_len`

ian-h-chamberlain (approved):

ian-h-chamberlain:
> Also, I noticed that getting lower and higher 32 bits of TitleIds is quite useful, and doing it manually every time or using a custom function is kinda ugly. What do you think about adding a TitleId tuple struct with a u64, that would also have other methods other than `lower` and `higher`, like `title_category`?

Haveing a `TitleId` wrapper (or maybe just `Title`? it could store the ID only and get other stuff on-demand) makes sense to me, then we could also have things like `Title::product_code()` that calls `AM_GetTitleProductCode`, as well as some wrapper functions around `AM_TitleEntry`. I don't think it needs to be this PR but you could add here if you wanted.

Maccraft123:
> > Also, I noticed that getting lower and higher 32 bits of TitleIds is quite useful, and doing it manually every time or using a custom function is kinda ugly. What do you think about adding a TitleId tuple struct with a u64, that would also have other methods other than `lower` and `higher`, like `title_category`?
> 
> Haveing a `TitleId` wrapper (or maybe just `Title`? it could store the ID only and get other stuff on-demand) makes sense to me, then we could also have things like `Title::product_code()` that calls `AM_GetTitleProductCode`, as well as some wrapper functions around `AM_TitleEntry`. I don't think it needs to be this PR but you could add here if you wanted.

So, `TitleId` struct, type alias... or what?
I'm not sure how to handle `AM` service init and exit, should it be done by `Title::product_code()`? Or should it have a reference to `Am` passed to it via an argument?
And where that `TitleId`/`Title` type would be defined in? `ctru-rs/src/service/am.rs`? `ctru-rs/src/titleid.rs`? Or where else?

In any case, I'll modify this PR so that the bare minimum of `TitleId`/`Title` struct is implemented to not have to make changes to do it later.

ian-h-chamberlain:
> So, `TitleId` struct, type alias... or what?
> I'm not sure how to handle `AM` service init and exit, should it be done by `Title::product_code()`? Or should it have a reference to `Am` passed to it via an argument?
> And where that `TitleId`/`Title` type would be defined in? `ctru-rs/src/service/am.rs`? `ctru-rs/src/titleid.rs`? Or where else?

Based on what we've done in other modules for similar functionality, this is what I'd propose:

* Create a `struct Title{id: u64}` or something like that. The only way to get it should be by calling `Am` methods which always ensures `Am` is initialized
* All this stuff can live in `am.rs` since it's so closely tied together

`Title::get_product_code` etc would just call `AM_whatever` under the hood, but as I write this I'm realizing you could do something like this:
```rs
let am = Am::init().unwrap();
let title = am.get_title_list(FsMediaType::Sd).unwrap()[0];
drop(am);
let _ = title.get_product_code(); // what happens here???
```
So maybe instead we should functions like `Am::get_product_code(&self, title: TitleId)` or something like that. The other alternative would be having something like `struct Title<'a>{id: u64, _am: PhantomData<&'a Am>` but that might not work super well if you need to pass title IDs around all over the place.

For now, probably just having `struct TitleId(u64)` is easiest and then any functions that need to call `ctru_sys::AM*` functions should be defined on `Am` itself, and take in a `TitleId` if relevant.

Maccraft123:
> Based on what we've done in other modules for similar functionality, this is what I'd propose:
> 
>     * Create a `struct Title{id: u64}` or something like that. The only way to get it should be by calling `Am` methods which always ensures `Am` is initialized
> 
>     * All this stuff can live in `am.rs` since it's so closely tied together
> 
> 
> `Title::get_product_code` etc would just call `AM_whatever` under the hood, but as I write this I'm realizing you could do something like this:
> 
> ```rust
> let am = Am::init().unwrap();
> let title = am.get_title_list(FsMediaType::Sd).unwrap()[0];
> drop(am);
> let _ = title.get_product_code(); // what happens here???
> ```
> 
> So maybe instead we should functions like `Am::get_product_code(&self, title: TitleId)` or something like that. The other alternative would be having something like `struct Title<'a>{id: u64, _am: PhantomData<&'a Am>` but that might not work super well if you need to pass title IDs around all over the place.
> 
> For now, probably just having `struct TitleId(u64)` is easiest and then any functions that need to call `ctru_sys::AM*` functions should be defined on `Am` itself, and take in a `TitleId` if relevant.

Why not have all methods of `TitleId` struct take a reference to `Am` struct and use that reference to indirectly call `ctru_sys::AM*`? Or maybe even storing a reference to `Am` in `TitleId` struct?

Meziu:
> Why not have all methods of `TitleId` struct take a reference to `Am` struct and use that reference to indirectly call `ctru_sys::AM*`? Or maybe even storing a reference to `Am` in `TitleId` struct?

The `Am` service will mutate the internal state only in functions related to title installation (which we currently don't plan on supporting). As such, I suggest saving a `PhantomData<&Am>` inside of `Title` for zero-sized lifetime checking. We wouldn't really need the reference regardless, correct?

Maccraft123:
> > Why not have all methods of `TitleId` struct take a reference to `Am` struct and use that reference to indirectly call `ctru_sys::AM*`? Or maybe even storing a reference to `Am` in `TitleId` struct?
> 
> The `Am` service will mutate the internal state only in functions related to title installation (which we currently don't plan on supporting). As such, I suggest saving a `PhantomData<&Am>` inside of `Title` for zero-sized lifetime checking. We wouldn't really need the reference regardless, correct?

I... think we won't need, as all title-related `ctru_sys::AM_*` functions would be used in `TitleId::*

I experimented with implementing `Title` struct [in this branch of my fork](https://github.com/Maccraft123/ctru-rs/commits/feature/titleid), to not affect this PR

ctru-rs/examples/title-info.rs
Meziu: > rust3ds is my first time doing any kind of homebrew for 3ds

Well, we always encourage using just devkitpro's toolchain first since it gives a better idea of how the system is going to actually handle the requests.

> I tried to get the actual names and couldn't figure it out on my own

Either way, let's at least not use the newline (just `print!()`with a comma after each title?)

Maccraft123: I'm thinking that implementing some basic scrolling would be good enough, as in just having offset variable, changing that offset with dpad/circle pad and on every change iterate through `title_list.skip(offset)`. That would also allow displaying more info of selected title id on bottom display

Maccraft123: I changed the interface quite a bit, what do you think now?

Maccraft123: > Can we retrieve the title's names instead?

In `feature/smdh` branch of my fork I implemented basic SMDH reading support and added it to `title-info` example.
I'll open PR with it after this one is merged.

ctru-rs/src/services/am.rs
```diff
@@ -0,0 +1,121 @@
+use crate::error::ResultCode;
+use crate::services::fs::FsMediaType;
+use std::marker::PhantomData;
+use std::mem::MaybeUninit;
+
+#[derive(Copy, Clone, Debug)]
+#[repr(C)]
+pub struct TitleInfo {
+    id: u64,
```
Meziu: This struct shouldn't be a copy of already existing (and auto-updating) types. You should simply hold a `ctru_sys::AM_TitleEntry` in here and reference that instead.

ctru-rs/src/services/am.rs
```diff
@@ -0,0 +1,121 @@
...
+    size: u64,
+    version: u16,
+    pad: u16,
+    type_: u32,
+}
+
+// Make sure TitleInfo is correct size
+const _TITLEINFO_SIZE_CHECK: [u8; 0x18] = [0; std::mem::size_of::<TitleInfo>()];
```
Meziu: We don't need size checks if we avoid hard-coded allocations. Refer to the comment on `TitleInfo`.

ctru-rs/src/services/am.rs
```diff
@@ -0,0 +1,121 @@
...
+    pub fn id(&self) -> u64 {
+        self.id
+    }
+    pub fn size_bytes(&self) -> u64 {
+        self.size
+    }
+    pub fn version(&self) -> u16 {
+        self.version
+    }
+    pub fn type_(&self) -> u32 {
```
Meziu: This isn't referenced in `libctru`'s docs, could you please explain what it does (and maybe link some references)?

ctru-rs/src/services/am.rs
```diff
@@ -0,0 +1,121 @@
...
+        let mut read_amount = 0;
+        unsafe {
+            ResultCode(ctru_sys::AM_GetTitleList(
+                &mut read_amount,
+                mediatype as u32,
+                count,
+                buf.as_mut_ptr(),
+            ))?;
+
+            buf.set_len(read_amount as usize);
```
Meziu: Simply make a vector with `vec![0; count]`. The data is filled out in every case so this seems like an useless unsafe call.

ctru-rs/src/services/am.rs
```diff
@@ -1,5 +1,75 @@
...
+
+#[derive(Copy, Clone, Debug)]
+#[repr(C)]
+pub struct TitleInfo {
+    id: u64,
+    size: u64,
+    version: u16,
+    pad: u16,
+    type_: u32,
+}
```
ian-h-chamberlain: Any reason not to just use `ctru_sys::AM_TitleEntry` instead of defining a new `#[repr(C)]` struct like this? You could avoid pointer casts and the `_TITLEINFO_SIZE_CHECK`.

I suppose you might still want a wrapper type like 
```rs
#[repr(transparent)]
pub struct TitleInfo(ctru_sys::AM_TitleEntry);
```

that way you could still have the helper methods and stuff like that. `type_()` would have to cast from bytes or something, I guess (`bytemuck::from_bytes(&unk[2..])` or similar)

ian-h-chamberlain (approved):
I think this looks good, but I would recommend using the existing `ctru_sys` structs over defining a new one, just to avoid potential ABI problems (and catch issues if the upstream struct ever changes).

Meziu (changes requested):

ctru-rs/src/services/am.rs
Maccraft123: Done

ctru-rs/src/services/am.rs
Maccraft123: I was sorta using Citra's sources as a reference to some structs and forgot to remove this once i noticed it's not really used anywhere else

ctru-rs/src/services/am.rs
Maccraft123: Done

ctru-rs/src/services/am.rs
Maccraft123: Done

ctru-rs/src/services/am.rs
Maccraft123: I just haven't thought about this... Fixed it, and removed `type_` stuff from TitleInfo

ctru-rs/src/services/am.rs
Meziu: No problem. I suggest giving a look at [libctru.devkitpro.org](https://libctru.devkitpro.org/) for `libctru` references and at [3dbrew.org](https://3dbrew.org) for specific system knowledge (it's helpful to learn how the hardware works). Citra doesn't really seem like the best source, especially since it's an inaccurate emulator mainly used for games, and not for development.

Maccraft123: I usually look at `ctru-sys/src/bindings.rs` and 3dbrew.org for documentation, as I find doxygen confusing to use

Meziu (approved):

ian-h-chamberlain (approved):

PR #96, opened 2023-03-04
Add Smdh struct

Maccraft123:
This MR adds Smdh struct to read smdh data from ExeFS of CXI files and some related data, like flags and region lockout data.
I wasn't really sure where to put it, so it went into `src/smdh.rs` and reading it went to `src/services/am.rs` as it's closely related to titles.
I'm not sure what can I do to make the actual-reading of data in `Title::get_smdh` look nicer.
I had an idea to make a somewhat-generic code for reading/writing from/to weird ArchiveIds with custom binary paths but I'm not sure how to implement it. The other part that looks sorta ugly-ish is verifying that smdh's magic data is 'SMDH', as I can't directly access magic field of `Smdh` from `am.rs` and it doesn't really make sense to add a public getter for it, so I added a public-to-crate getter for it.

I used https://www.3dbrew.org/wiki/SMDH and https://github.com/J-D-K/JKSM as references.

### DIFF
--- a/ctru-rs/src/services/am.rs
+++ b/ctru-rs/src/services/am.rs
@@ -74,7 +74,7 @@ impl<'a> Title<'a> {
         let archive_path_data: [u32; 4] = [
             self.low_u32(),
             self.high_u32(),
-            self.media_type() as u32,
+            self.mediatype as u32,
             0x0,
         ];
         let smdh_path_data: [u32; 5] = [0x0, 0x0, 0x2, u32::from_le_bytes(*b"icon"), 0x0];
--- a/ctru-rs/src/services/am.rs
+++ b/ctru-rs/src/services/am.rs
@@ -1,6 +1,9 @@
 use crate::error::ResultCode;
+use crate::services::fs::Fs;
 use crate::services::fs::FsMediaType;
+use crate::smdh::Smdh;
 use std::marker::PhantomData;
+use std::mem::size_of;
 use std::mem::MaybeUninit;
 
 #[derive(Copy, Clone, Debug)]
@@ -18,6 +21,7 @@ impl TitleInfo {
     }
 }
 
+#[derive(Copy, Clone, Debug)]
 pub struct Title<'a> {
     id: u64,
     mediatype: FsMediaType,
@@ -27,6 +31,14 @@ pub struct Title<'a> {
 impl<'a> Title<'a> {
     pub fn id(&self) -> u64 {
         self.id
+    }
+
+    pub fn low_u32(&self) -> u32 {
+        (self.id & 0x0000_0000_ffff_ffff) as u32
+    }
+
+    pub fn high_u32(&self) -> u32 {
+        ((self.id & 0xffff_ffff_0000_0000) >> 32) as u32
     }
 
     pub fn get_product_code(&self) -> crate::Result<String> {
@@ -55,6 +67,59 @@ impl<'a> Title<'a> {
 
             Ok(info.assume_init())
         }
+    }
+
+    pub fn get_smdh(&self) -> crate::Result<Smdh> {
+        // i have no idea how to make this look better
+        let archive_path_data: [u32; 4] = [
+            self.low_u32(),
+            self.high_u32(),
+            self.media_type() as u32,
+            0x0,
+        ];
+        let smdh_path_data: [u32; 5] = [0x0, 0x0, 0x2, u32::from_le_bytes(*b"icon"), 0x0];
+
+        let archive_path = ctru_sys::FS_Path {
+            type_: ctru_sys::PATH_BINARY,
+            size: size_of::<[u32; 4]>() as u32,
+            data: archive_path_data.as_ptr() as *const libc::c_void,
+        };
+        let smdh_path = ctru_sys::FS_Path {
+            type_: ctru_sys::PATH_BINARY,
+            size: size_of::<[u32; 5]>() as u32,
+            data: smdh_path_data.as_ptr() as *const libc::c_void,
+        };
+
+        let _fs = Fs::init();
+        let mut smdh_handle = 0;
+
+        let smdh: Smdh = unsafe {
+            let mut ret = MaybeUninit::zeroed();
+            ResultCode(ctru_sys::FSUSER_OpenFileDirectly(
+                &mut smdh_handle,
+                ctru_sys::ARCHIVE_SAVEDATA_AND_CONTENT,
+                archive_path,
+                smdh_path,
+                ctru_sys::FS_OPEN_READ,
+                0x0,
+            ))?;
+
+            ResultCode(ctru_sys::FSFILE_Read(
+                smdh_handle,
+                std::ptr::null_mut(),
+                0x0,
+                ret.as_mut_ptr() as *mut libc::c_void,
+                size_of::<Smdh>() as u32,
+            ))?;
+
+            ResultCode(ctru_sys::FSFILE_Close(smdh_handle))?;
+
+            ret.assume_init()
+        };
+
+        assert_eq!(&smdh.magic(), b"SMDH");
+
+        Ok(smdh)
     }
 }
 
--- a/ctru-rs/src/services/am.rs
+++ b/ctru-rs/src/services/am.rs
@@ -71,12 +71,8 @@ impl<'a> Title<'a> {
 
     pub fn get_smdh(&self) -> crate::Result<Smdh> {
         // i have no idea how to make this look better
-        let archive_path_data: [u32; 4] = [
-            self.low_u32(),
-            self.high_u32(),
-            self.mediatype as u32,
-            0x0,
-        ];
+        let archive_path_data: [u32; 4] =
+            [self.low_u32(), self.high_u32(), self.mediatype as u32, 0x0];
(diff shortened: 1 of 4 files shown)


Techie-Pi (commented):
I haven't checked the code in depth (I have exams next week, so I don't think I will be able to), but you should remove most `#[derive(Copy)]`.
I don't think that copying of big structs (`Smdh`, `Title`, etc) is a good idea, specially on a system like the 3DS

Maccraft123:
> I haven't checked the code in depth (I have exams next week, so I don't think I will be able to), but you should remove most `#[derive(Copy)]`. I don't think that copying of big structs (`Smdh`, `Title`, etc) is a good idea, specially on a system like the 3DS

Sure, will do.
Sorta got used to doing it whenever I can as I don't really care that much about performance until it becomes an issue.
Will do my best to remember to not do that in future.

Maccraft123:
Tbh it might be a good idea to move `Smdh`, and other data formats(NCCH, CIA etc), into their own crate, so that it could be used with tools that don't run on 3ds

Meziu:
> Tbh it might be a good idea to move `Smdh`, and other data formats(NCCH, CIA etc), into their own crate, so that it could be used with tools that don't run on 3ds

That is a very interesting point, especially since our tools (mainly `cargo-3ds`) would benefit immensely from having custom file format support rather than depending on installed binaries.

Maccraft123:
> That is a very interesting point, especially since our tools (mainly `cargo-3ds`) would benefit immensely from having custom file format support rather than depending on installed binaries.

If you want, I can try making a proof of concept for Rust `smdhtool` alternative and a Rust library to read/write SMDH data

Maccraft123:
Yeah, I can see it work,
https://github.com/Maccraft123/ctr-formats
https://github.com/Maccraft123/ctru-rs/commit/02d32b1cf81ce8a962e3c5ce2f89d80027827be9

ctru-rs/src/smdh.rs
```diff
@@ -0,0 +1,99 @@
+use crate::services::cfgu::Language;
+use bitflags::bitflags;
+
+#[derive(Copy, Clone, Debug)]
+#[repr(C)]
+pub struct Smdh {
```
ian-h-chamberlain, line 6: A doc comment here linking to 3dbrew.org might be useful.

Also, this is a huge struct so deriving `Copy` is probably not ideal. I think we'd want people to opt-in to explicit `clone()` rather than automatically copying

ctru-rs/src/smdh.rs
```diff
@@ -0,0 +1,99 @@
...
+    magic: [u8; 4],
+    version: u16,
+    titles: [SmdhTitle; 16],
+    age_ratings: [u8; 16],
```
ian-h-chamberlain, line 10: Age rating might be deserving of its own struct / bitflags, from reading the docs about it

ctru-rs/src/services/am.rs
```diff
@@ -56,6 +68,55 @@ impl<'a> Title<'a> {
             Ok(info.assume_init())
         }
     }
+
+    pub fn get_smdh(&self) -> crate::Result<Smdh> {
+        // i have no idea how to make this look better
+        let archive_path_data: [u32; 4] =
+            [self.low_u32(), self.high_u32(), self.mediatype as u32, 0x0];
```
ian-h-chamberlain, line 75: Do any of the helpers in `src/services/fs.rs` work here? At minimum it looks like there are some helpers already for `Archive`, 

If not, I'd say maybe we should add on there to include SMDH / `"icon"` support , although it might take a bit of finagling to get everything working.

ctru-rs/src/services/am.rs
```diff
@@ -56,6 +68,55 @@ impl<'a> Title<'a> {
...
+        let smdh_path_data: [u32; 5] = [0x0, 0x0, 0x2, u32::from_le_bytes(*b"icon"), 0x0];
+
+        let archive_path = ctru_sys::FS_Path {
+            type_: ctru_sys::PATH_BINARY,
+            size: size_of::<[u32; 4]>() as u32,
+            data: archive_path_data.as_ptr() as *const libc::c_void,
```
ian-h-chamberlain, line 81: Maybe not as relevant with the other changes I mentioned, but in general:

```suggestion
            data: archive_path_data.as_ptr().cast(),
```

ctru-rs/src/services/am.rs
```diff
@@ -56,6 +68,55 @@ impl<'a> Title<'a> {
...
+                ret.as_mut_ptr() as *mut libc::c_void,
+                size_of::<Smdh>() as u32,
+            ))?;
+
+            ResultCode(ctru_sys::FSFILE_Close(smdh_handle))?;
+
+            ret.assume_init()
+        };
+
+        assert_eq!(&smdh.magic(), b"SMDH");
```
ian-h-chamberlain, line 116: Instead of assert could this return `Err` if the magic is wrong ("corrupted SMDH" or something)

ctru-rs/src/services/am.rs
```diff
@@ -56,6 +68,55 @@ impl<'a> Title<'a> {
...
+            size: size_of::<[u32; 4]>() as u32,
+            data: archive_path_data.as_ptr() as *const libc::c_void,
+        };
+        let smdh_path = ctru_sys::FS_Path {
+            type_: ctru_sys::PATH_BINARY,
+            size: size_of::<[u32; 5]>() as u32,
+            data: smdh_path_data.as_ptr() as *const libc::c_void,
+        };
+
+        let _fs = Fs::init();
```
ian-h-chamberlain, line 89: ideally I would suggest we return an error that `Fs` wasn't initialized yet here, if possible, rather than initializing when this function is called.

ian-h-chamberlain (changes requested):
Phew, finally got around to reviewing this, sorry! I think the new structure looks fine, but it would be preferable if we can take advantage of / augment the existing `Fs` functionality instead of writing the `ctru_sys` calls directly in the implementation there.

Hopefully this isn't too tricky to figure out -- requesting changes for now.

ctru-rs/src/services/am.rs
Meziu: Passing a reference to an existing `Fs` instance might be preferable.

ian-h-chamberlain: That's true, I didn't consider that but it would enforce that `Fs` is initialized beforehand and lives long enough to be used for this purpose. Seems like a better option

Maccraft123: I'm thinking of adding sdmh loading code in a type-safe and generic way, into `src/services/fs.rs`, with `Archive<ArchiveType>` struct storing inner `ArchiveType` if the archive requires a binary path, otherwise it would store a generic variant that would represent archives which use only text paths. Also looking at the `smdh_path_data`, it sorta looks like a struct with 2 `repr(u32)` enums, and either `repr(u32)` enum that looks *very much* like a C string without null termination with a null bit put at the end of it, or just a C string

ctru-rs/src/smdh.rs
AzureMarker: +1 on removing Copy

ctru-rs/src/smdh.rs
```diff
@@ -0,0 +1,99 @@
+use crate::services::cfgu::Language;
+use bitflags::bitflags;
+
+#[derive(Copy, Clone, Debug)]
+#[repr(C)]
+pub struct Smdh {
+    magic: [u8; 4],
+    version: u16,
```
AzureMarker, line 8: There's supposed to be 2 bytes of reserved space after the version.

ctru-rs/src/smdh.rs
```diff
@@ -0,0 +1,99 @@
+use crate::services::cfgu::Language;
+use bitflags::bitflags;
+
+#[derive(Copy, Clone, Debug)]
+#[repr(C)]
```
AzureMarker, line 5: I wonder if using `repr(C)` will cause extra padding that throws off the byte offsets. You probably need to pass a byte buffer to libctru and then parse the bytes into a struct.

ctru-rs/src/smdh.rs
```diff
@@ -0,0 +1,99 @@
...
+    magic: [u8; 4],
+    version: u16,
+    titles: [SmdhTitle; 16],
+    age_ratings: [u8; 16],
+    region_lock: RegionLock,
+    matchmaker_id: u32,
+    matchmaker_bit_id: u64,
+    flags: SmdhFlags,
+    eula_version_major: u8,
+    eula_version_minor: u8,
```
AzureMarker, line 16: I think the minor version comes before major

ctru-rs/src/services/am.rs
```diff
@@ -29,6 +33,14 @@ impl<'a> Title<'a> {
         self.id
     }
 
+    pub fn low_u32(&self) -> u32 {
+        (self.id & 0x0000_0000_ffff_ffff) as u32
+    }
+
+    pub fn high_u32(&self) -> u32 {
+        ((self.id & 0xffff_ffff_0000_0000) >> 32) as u32
+    }
```
AzureMarker, line 42: It isn't clear what these functions are doing from the name & signature. Maybe they can be inlined. If these functions are really needed, they should probably be private and well documented.

Meziu (changes requested):
All raw calls to `FS_` functions should be replaced by safe `ctru::fs` calls.

PR #101, opened 2023-03-17
Initial pass of `libctru` version checking in `ctru-sys`

### LINKED ISSUE #100: Versioning of `ctru-sys` crate
> > Pick initial numbers for each crate and coordinate publishing a release!

> In my opinion, `ctru-rs`can release with a generic version number (like `1.0`or `0.5`), and the same can be said for `pthread-3ds`and `linker-fix-3ds`. However, I would like to underline the importance of `ctru-sys`'s version number, since we could release it with the same version number of `libctru` (which would double down as a "supported version" number). We'd need to be careful about doc generation though, since we wouldn't be able to easily make a patch release.

_Originally posted by @Meziu in https://github.com/rust3ds/ctru-rs/discussions/97#discussioncomment-5295698_

----

This is an interesting idea I hadn't considered before! For inspiration I thought I'd look at some other `sys` crates, and I found this which looks kinda interesting: https://gitlab.com/taricorp/llvm-sys.rs#llvm-compatibility

Basically, we could have a `ctru-sys` version `21.2.x` for the current release, and bump the minor version whenever we regenerate bindings for a new patch of `libctru`. This would also allow us to make docgen improvements, etc. as a patch version on the crate instead.

In the case of `llvm-sys`, they have some checks in place to verify library versions at build time. It might be harder to do that with `libctru`, since it's just a static library, but I was thinking there might be some ways to check it with `dkp-pacman` / `pacman`, and spit out a [build warning](https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargowarningmessage) if the versions mismatch, which would be displayed if the build failed (link errors etc).

Opening this issue to track + discuss the possibilities.

ian-h-chamberlain:
Partially addresses #100, but we'd probably want to do something as part of the CI release process as well.

This isn't set in stone, I'm just experimenting with some different things to see what makes sense. So far this seems to work pretty well, but I'm not sure if we want to encode the "duplicate" info with the `+` metadata on the build version as well as with the major/minor scheme mentioned in #100 

Let me know what you think, I'll leave as a draft for now and we can discuss it.

### DIFF
--- a/ctru-sys/build.rs
+++ b/ctru-sys/build.rs
@@ -27,9 +27,6 @@ fn main() {
             println!("cargo:rustc-env=LIBCTRU_MAJOR={maj}");
             println!("cargo:rustc-env=LIBCTRU_MINOR={min}");
             println!("cargo:rustc-env=LIBCTRU_PATCH={patch}");
-
-            // It would be nice if we could rerun-if-changed on libctru itself,
-            // maybe we can write a file to OUT_DIR and check that file?
         }
         Err(err) => println!("cargo:warning=failed to check libctru version: {err}"),
     }
@@ -50,7 +47,7 @@ fn parse_version(version: &str) -> Result<(String, String, String), &str> {
 fn check_libctru_version() -> Result<(String, String, String), Box<dyn Error>> {
     let pacman = which::which("dkp-pacman").or_else(|_| which::which("pacman"))?;
 
-    let Output { stdout, .. } = Command::new(pacman)
+    let Output { stdout, .. } = Command::new(&pacman)
         .args(["--query", "libctru"])
         .stderr(Stdio::inherit())
         .output()?;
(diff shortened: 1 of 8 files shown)


ctru-sys/build.rs
```diff
@@ -14,4 +16,69 @@ fn main() {
...
+
+            // These are accessible by the crate during build with `env!()`.
+            // We might consider exporting some public constants or something.
+            println!("cargo:rustc-env=LIBCTRU_VERSION={maj}.{min}.{patch}");
+            println!("cargo:rustc-env=LIBCTRU_MAJOR={maj}");
+            println!("cargo:rustc-env=LIBCTRU_MINOR={min}");
+            println!("cargo:rustc-env=LIBCTRU_PATCH={patch}");
+
+            // It would be nice if we could rerun-if-changed on libctru itself,
+            // maybe we can write a file to OUT_DIR and check that file?
```
AzureMarker: I don't think that will work since we're the ones changing the file, so it won't change if libctru changes. Instead, we could do a `rerun-if-changed` on the `libctru` static library file. We'd just have to resolve that path.

ian-h-chamberlain: Good point, we already know the library name / search dirs we're linking against so it should be possible to figure out the filename of the actual library being used, which would change if the libs got updated. Ideally we could detect header changes too, but maybe just library is enough...

I suppose in theory we could also take the output of `[dkp-]pacman -Ql libctru` to list _all_ files owned by the package, but I'm not sure if that provides any significant advantage over just using the library. I guess it would include docstring changes ?

Meziu (approved):

Meziu:
We should let people the have the choice to use an external `libctru` binary rather than requiring `dkp-pacman`. It would be helpful to ensure compatibility in the future.

ian-h-chamberlain:
> We should let people the have the choice to use an external `libctru` binary rather than requiring `dkp-pacman`. It would be helpful to ensure compatibility in the future.

That's a fair point. This change doesn't prevent anything like that, it would just display a warning that we couldn't determine the `libctru` version (and even then, that would only be displayed if using `-vv` or the compilation/linking failed). Hopefully that warning would give the user a clue that they might need to upgrade or use a supported version instead.

Meziu (approved):
LGTM 👍

Issue #102, opened 2023-03-22
Streamline the API


Meziu:
Before release, the API should be cleaned up to be in line with modern Rust standards. For example, `get_` methods in struct implementations aren't the standard, and we should check for name repetitions within modules and enums (i just noticed how annoying [`KeyPad`](https://github.com/rust3ds/ctru-rs/blob/0b54ffd0b87e3848ce7a3af3f98eca865b6f3f11/ctru-rs/src/services/hid.rs#L12) is with its `KEY` prefix).

In general, the API should be modern and, most importantly, cohesive.

Meziu:
I will close this issue on the premise that a lot of work has been done in #107 and #116. However, as most “generic” issues, the work is ongoing and there might be new waves of big changes + general improvement over time.

PR #104, opened 2023-03-26
Improve Ndsp API

Meziu:
This PR comes as an after effect of my experience with #103 and needs for #102 and #97.

The focus points are: 
- `const fn` for `AudioFormat::size`
- creation of `AudioMix`
- rename of the methods (usually to avoid `get_` in getters)
- every "size" type changed into `usize` in the public interface, regardless of the underlying format.

The last two should be applied to all modules, but this PR only covers `Ndsp`.

### DIFF
--- a/ctru-rs/examples/audio-filters.rs
+++ b/ctru-rs/examples/audio-filters.rs
@@ -5,13 +5,13 @@ use std::f32::consts::PI;
 use ctru::linear::LinearAllocator;
 use ctru::prelude::*;
 use ctru::services::ndsp::{
-    wave::{WaveInfo, WaveStatus},
-    AudioFormat, InterpolationType, Ndsp, OutputMode,
+    wave::{Wave, WaveStatus},
+    AudioFormat, AudioMix, InterpolationType, Ndsp, OutputMode,
 };
 
 const SAMPLE_RATE: usize = 22050;
 const SAMPLES_PER_BUF: usize = SAMPLE_RATE / 10; // 2205
-const BYTES_PER_SAMPLE: usize = 4;
+const BYTES_PER_SAMPLE: usize = AudioFormat::PCM16Stereo.size();
 const AUDIO_WAVE_LENGTH: usize = SAMPLES_PER_BUF * BYTES_PER_SAMPLE;
 
 // Note Frequencies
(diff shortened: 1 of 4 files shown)


ctru-rs/src/services/ndsp/mod.rs
```diff
@@ -30,6 +31,18 @@ pub enum AudioFormat {
     PCM16Stereo = ctru_sys::NDSP_FORMAT_STEREO_PCM16,
 }
 
+/// Representation of volume mix for a channel.
+/// Each member is made up of 2 values, the first is for the "left" channel, while the second is for the "right" channel.
+#[derive(Copy, Clone, Debug)]
```
Techie-Pi, line 35: Is there a reason to why this struct derives `Copy`?

Techie-Pi (commented):
This looks great, but I left a question in the review

ctru-rs/examples/audio-filters.rs
```diff
@@ -79,10 +79,7 @@ fn main() {
     channel_zero.set_format(AudioFormat::PCM16Stereo);
 
     // Output at 100% on the first pair of left and right channels.
```
ian-h-chamberlain, line 81: This comment might be good to move to the `Default` impl itself since the example is no longer specifying it explicitly.

ctru-rs/src/services/ndsp/mod.rs
```diff
@@ -30,6 +31,18 @@ pub enum AudioFormat {
     PCM16Stereo = ctru_sys::NDSP_FORMAT_STEREO_PCM16,
 }
 
+/// Representation of volume mix for a channel.
+/// Each member is made up of 2 values, the first is for the "left" channel, while the second is for the "right" channel.
```
ian-h-chamberlain: If you wanted you could have a `struct { pub left: f32, pub right: f32 }` for each of those just to be explicit. I think it's mostly self-explanatory as-is though

ctru-rs/src/services/ndsp/mod.rs
```diff
@@ -302,14 +306,70 @@ impl Channel<'_> {
...
+            front: (data[0], data[1]),
+            back: (data[2], data[3]),
+            aux1_front: (data[4], data[5]),
+            aux1_back: (data[6], data[7]),
+            aux2_front: (data[8], data[9]),
+            aux2_back: (data[10], data[11]),
+        }
+    }
+
+    pub fn to_raw(&self) -> [f32; 12] {
```
ian-h-chamberlain: Is this different than e.g. `bytemuck::cast(&self)` ?

Also I think usual convention would be to call this `into_raw` if it take `self`, or `as_raw` if it takes `&self`. If you decide to keep the `Copy` impl for this struct, it could probably be `into_raw`.

ian-h-chamberlain (approved):

ctru-rs/src/services/ndsp/mod.rs
Meziu: It's an informational struct. It holds no service data or anything, but simply wraps an array. There is no reason why it shouldn't be `Copy`, you can just get a reference regardless.

ctru-rs/src/services/ndsp/mod.rs
Meziu: This has been implemented with `From`.

ctru-rs/src/services/ndsp/mod.rs
Meziu: This changed, review the new version.

Techie-Pi (approved):

PR #105, opened 2023-03-27
Remove extern "C" process stubs from test runner

ian-h-chamberlain:
One piece of https://github.com/rust3ds/rust-linker-fix-3ds/issues/18 

See also https://github.com/rust3ds/rust-linker-fix-3ds/pull/19


Meziu (approved):
We can merge them immediately.

PR #107, opened 2023-03-31
Improve general API

Meziu:
This PR comes with a lot of changes to the API for all modules (related to #102), loosely following the [standards](https://rust-lang.github.io/api-guidelines/checklist.html). The changes include:
- Moved `gfx` and `romfs` to the `services` module
- All `ctru-sys` derivative enums have members set to the `ctru-sys` equivalent (and a blanket `From` implementation)
- `get_*` methods have been renamed
- Massive renames for enums and structs (especially in `cam`)
- Fixed name repetitions in enum variants and module members
- fixed `cam` enums being wrongly used as `bitflags!`
- `usize` -> `u32` convertions have been solved with `.into()` where possible
- Small documentation improvements (a full-blown doc PR will come)

This is a massive PR when counting changed lines, but in reality it's mostly just cleanup and refactoring.


Techie-Pi (approved):
Everything seems great 👍

ctru-rs/src/applets/mii_selector.rs
```diff
@@ -135,10 +135,10 @@ impl MiiSelector {
 
     /// Set where the cursor will be.
     /// If there's no Mii at that index, the cursor will start at the Mii with the index 0
-    pub fn set_initial_index(&mut self, index: u32) {
+    pub fn set_initial_index(&mut self, index: usize) {
         // This function is static inline in libctru
         // https://github.com/devkitPro/libctru/blob/af5321c78ee5c72a55b526fd2ed0d95ca1c05af9/libctru/include/3ds/applets/miiselector.h#L155
-        self.config.initial_index = index
+        self.config.initial_index = index as u32;
```
ian-h-chamberlain, line 141: man it would be nice if we could opt in to allowing `.into()` or implicit conversion since we only target a single platform, but I guess `as` casting like this is fine in general (we already use it a lot I think).

ctru-rs/src/applets/swkbd.rs
```diff
@@ -99,32 +101,32 @@ impl Swkbd {
     ///
     /// The text received from the keyboard will be truncated if it is greater than 2048 bytes
     /// in length.
-    pub fn get_utf8(&mut self, buf: &mut String) -> Result<Button, Error> {
+    pub fn write_to_string(&mut self) -> Result<(String, Button), Error> {
```
ian-h-chamberlain: hmm, this name seems to me like it would be more appropriate for the old signature (write to some mutable string, like [`std::io::Read::read_to_string`](https://doc.rust-lang.org/std/io/trait.Read.html#method.read_to_string)).

I think `write_bytes` makes sense for the other function, but maybe this one could be something like this?
```rs
pub fn get_string(&mut self, max_bytes: usize) -> Result<(String, Button), Error>
```

ctru-rs/src/applets/swkbd.rs
```diff
@@ -99,32 +101,32 @@ impl Swkbd {
...
     }
 
     /// Fills the provided buffer with a UTF-8 encoded, NUL-terminated sequence of bytes from
     /// this software keyboard.
     ///
     /// If the buffer is too small to contain the entire sequence received from the keyboard,
-    /// the output will be truncated but should still be well-formed UTF-8
-    pub fn get_bytes(&mut self, buf: &mut [u8]) -> Result<Button, Error> {
+    /// the output will be truncated but should still be well-formed UTF-8.
+    pub fn write_bytes(&mut self, buf: &mut [u8]) -> Result<Button, Error> {
```
ian-h-chamberlain: actually per my other comment maybe [`write_exact`](https://doc.rust-lang.org/std/io/trait.Read.html#method.read_exact) or something like that would a good name.

... now I'm wondering if we should just `impl Read` for this struct, but I guess the `Button` part of the result makes it a little trickier to do that

ctru-rs/src/lib.rs
```diff
@@ -15,10 +15,50 @@ extern crate pthread_3ds;
...
+
+macro_rules! from_type_to_i32 {
+    ($from_type:ty) => {
+        impl From<$from_type> for i32 {
+            fn from(v: $from_type) -> Self {
+                v as i32
+            }
+        }
+    };
+}
```
ian-h-chamberlain, line 26: these could probably go into a `mod macros` or something. Also might be simpler to have a single macro, `from_impl!(Ty, i32` and just call it like
```rs
from_impl!(Kind, u32);
```

I'm sure there's some upstream crate that can do this for us too but seems like overkill for this simple case.

ctru-rs/src/services/fs.rs
```diff
@@ -1016,45 +1018,6 @@ impl Drop for Dir {
...
-            NandTwlFS => ctru_sys::ARCHIVE_NAND_TWL_FS,
-            GameCardSavedata => ctru_sys::ARCHIVE_GAMECARD_SAVEDATA,
-            UserSavedata => ctru_sys::ARCHIVE_USER_SAVEDATA,
-            DemoSavedata => ctru_sys::ARCHIVE_DEMO_SAVEDATA,
-        }
-    }
-}
+from_type_to_u32!(FsMediaType);
+from_type_to_u32!(PathType);
+from_type_to_u32!(ArchiveID);
```
ian-h-chamberlain: do these replace the previous `From` impls because `ctru_sys::FS_PathType` is just an alias or something?

ctru-rs/src/services/gspgpu.rs
```diff
@@ -1,31 +1,31 @@
 //! GSPGPU service
 
-use std::convert::From;
-
 #[derive(Copy, Clone, Debug)]
+#[repr(u32)]
 pub enum Event {
```
ian-h-chamberlain, line 5: I wonder if we should also add `#[non_exhaustive]` on any of these enums? It might be hard to say at a glance which ones might have stuff added or not

ctru-rs/src/services/mod.rs
```diff
@@ -18,6 +19,25 @@ mod reference;
...
+            //! The RomFS folder has not been detected and/or the `romfs` feature has not been enabled.
+            //!
+            //! Configure the path in Cargo.toml (the default path is "romfs"). Paths are relative to the
+            //! `CARGO_MANIFEST_DIR` environment variable, which is the directory containing the manifest of
+            //! your package.
+            //!
+            //! ```toml
+            //! [package.metadata.cargo-3ds]
+            //! romfs_dir = "romfs"
+            //! ```
```
ian-h-chamberlain, line 36: at some point we might consider adding something like
```rs
} else if #[cfg(all(feature = "romfs", not(romfs_exists)))] {
    compile_error!("romfs feature is enabled but no romfs found!");
}
```

haven't tested to see if that does what I think though

ctru-rs/src/services/cam.rs
```diff
@@ -21,191 +20,165 @@ pub struct Cam {
...
-bitflags! {
-    /// A set of flags to be passed to [Camera::flip_image]
-    #[derive(Default)]
-    pub struct CamFlip: u32 {
-        const NONE       = ctru_sys::FLIP_NONE;
-        const HORIZONTAL = ctru_sys::FLIP_HORIZONTAL;
-        const VERTICAL   = ctru_sys::FLIP_VERTICAL;
-        const REVERSE    = ctru_sys::FLIP_REVERSE;
-    }
+/// Flag to pass to [Camera::flip_image]
```
ian-h-chamberlain, line 23: tiny nit, I usually prefer ``[`backticks`]`` inside the doc links for stuff that refers to real APIs like this. I guess you didn't change this but just something to think about while cleaning up docstrings.

ctru-rs/src/services/cam.rs
```diff
@@ -21,191 +20,165 @@ pub struct Cam {
...
-    }
+/// Flag to pass to [Camera::set_white_balance] or
+/// [Camera::set_white_balance_without_base_up]
+#[derive(Copy, Clone, Debug)]
+#[repr(u32)]
+pub enum WhiteBalance {
+    /// Normal
+    Auto = ctru_sys::WHITE_BALANCE_AUTO,
+    /// Tungsten
+    Type3200K = ctru_sys::WHITE_BALANCE_3200K,
```
ian-h-chamberlain: `Type` prefix here seems redundant to me, but I guess you can't just use numbered values. Maybe `Temp3200K` instead?

ctru-rs/src/services/cam.rs
```diff
@@ -21,191 +20,165 @@ pub struct Cam {
...
-        const LETTER    = ctru_sys::PHOTO_MODE_LETTER;
-    }
+/// Flag to pass to [Camera::set_photo_mode]
+#[derive(Copy, Clone, Debug)]
+#[repr(u32)]
+pub enum PhotoMode {
+    Normal = ctru_sys::PHOTO_MODE_NORMAL,
+    Portrait = ctru_sys::PHOTO_MODE_PORTRAIT,
+    Landscape = ctru_sys::PHOTO_MODE_LANDSCAPE,
+    Nightview = ctru_sys::PHOTO_MODE_NIGHTVIEW,
```
ian-h-chamberlain: nit: maybe `NightView` ?

ctru-rs/src/services/cam.rs
```diff
@@ -21,191 +20,165 @@ pub struct Cam {
...
-        const OFF   = ctru_sys::LENS_CORRECTION_OFF;
-        const ON_70 = ctru_sys::LENS_CORRECTION_ON_70;
-        const ON_90 = ctru_sys::LENS_CORRECTION_ON_90;
+/// Flag to pass to [Camera::set_lens_correction]
+#[derive(Copy, Clone, Debug)]
+#[repr(u32)]
+pub enum LensCorrection {
+    Off = ctru_sys::LENS_CORRECTION_DARK,
+    Normal = ctru_sys::LENS_CORRECTION_NORMAL,
+    Bright = ctru_sys::LENS_CORRECTION_BRIGHT,
```
ian-h-chamberlain, line 129: I assume these are synonyms for `OFF`, `70`, `90` ? Maybe comments indicating that would be helpful

ctru-rs/src/services/cam.rs
```diff
@@ -307,7 +280,7 @@ pub trait Camera {
 
     /// Returns the maximum amount of transfer bytes based on the view size, trimming, and other
     /// modifications set to the camera
-    fn get_transfer_bytes(&self) -> crate::Result<u32> {
+    fn transfer_bytes(&self) -> crate::Result<u32> {
```
ian-h-chamberlain: minor nit, this sounds like it could be a verb ("transfer the bytes") but I guess the signature makes it more obvious. Maybe `transfer_byte_count()` is clearer?

ian-h-chamberlain (approved):
Looks good! :pray: thank you for doing all this cleanup work.

I had some smaller comments but nothing major

ctru-rs/src/applets/mii_selector.rs
Meziu: I tried removing as much casting as possible, but when it comes to simple types we can't really do much 😅

ctru-rs/src/applets/swkbd.rs
Meziu: Setting the amount of bytes manually is quite interesting, I'll implement it. 👍

ctru-rs/src/applets/swkbd.rs
Meziu: I thought about implementing `Read` already, but the nature of the applet is a bit more complicated than a simple byte read, so I tried to change the naming conventions instead.

ctru-rs/src/services/fs.rs
Meziu: Yes. I could still try to get the "official" signature by following the other suggestion you [made](https://github.com/rust3ds/ctru-rs/pull/107#discussion_r1154893427).

ctru-rs/src/services/cam.rs
Meziu: I will look into this when making the other PR 👍

ctru-rs/src/services/cam.rs
Meziu: I searched far and wide for a good prefix, let's go with `Temp`.

ctru-rs/src/services/cam.rs
Meziu: Oh, true. I hadn't seen it that way, thanks for the suggestion 👍

ctru-rs/src/services/mod.rs
Meziu: That's not really in the scope of this PR, but yes, ROMFS does have improvement potential.

ctru-rs/src/services/gspgpu.rs
Meziu: Unless there are specific changes in `libctru` (which we cannot predict) there shouldn't be any major changes in the enums already implemented. `#[non_exhaustive]` would be a substantial limit to the possibilities enabled by the public enums.

ctru-rs/src/services/fs.rs
ian-h-chamberlain: yeah, I'm not sure if it makes a difference as far as rustdoc is concerned (maybe it would show the alias name instead of the underlying type?)

ctru-rs/src/services/mod.rs
ian-h-chamberlain: Yeah you're right, sorry. Opened #108

ian-h-chamberlain (approved):

ctru-rs/src/services/fs.rs
Meziu: Yes, it shows the `ctru_sys` name instead.

ctru-rs/src/applets/swkbd.rs
```diff
@@ -90,41 +92,39 @@ impl Swkbd {
...
     }
 
     /// Fills the provided buffer with a UTF-8 encoded, NUL-terminated sequence of bytes from
     /// this software keyboard.
     ///
     /// If the buffer is too small to contain the entire sequence received from the keyboard,
-    /// the output will be truncated but should still be well-formed UTF-8
-    pub fn get_bytes(&mut self, buf: &mut [u8]) -> Result<Button, Error> {
+    /// the output will be truncated but should still be well-formed UTF-8.
+    pub fn write_exact(&mut self, buf: &mut [u8]) -> Result<Button, Error> {
```
AzureMarker, line 127: Not sure that `write_exact` is a good name for this, since it's reading not writing

ian-h-chamberlain: I guess there's two ways of looking at it - it's reading from the software keyboard, but writing to the buffer that's passed in. I guess the `std::io::Read` names are based on "reading from self" so perhaps `read_exact` would be more in line with that...

Meziu: Honestly, I believe it should just implement `std::io::Read`. We can document the blocking behaviour in the trait impl.

Issue #106, opened 2023-03-27
Improve Testing and CI


Meziu:
We should mass produce tests now that the API is getting stabilized, and integrate with the CI to make sure the examples build after changes.

ian-h-chamberlain:
Definitely agree, unfortunately after doing a bit of research, I found out some things:

* Our current CI does not appear to actually check any doctests at all, so we will want to fix that. related https://github.com/rust-lang/cargo/issues/6669
* Doc tests are... well, kind of an unfinished feature it seems like. A few gaps we will most likely run into:
	* We need an unstable flag to even compile doctests: https://github.com/rust-lang/cargo/issues/7040
	* `--no-run` is not supported by `rustdoc` / `cargo`:
		```console
		$ cargo 3ds test --doc -Zdoctest-xcompile
		No pre-build std found, using build-std
		error: Can't skip running doc tests with --no-run
		```
		We can probably work around this in `cargo-3ds` somehow since this still seems to work: `cargo test -vvv -Zdoctest-xcompile -Zbuild-std --target armv6k-nintendo-3ds --doc`. 
		See also https://github.com/rust-lang/rust/issues/87022
 * For CI testing purposes, we can try using `target.armv6k-nintendo-3ds.runner` Cargo config but we don't really have a good way of verifying the tests actually pass. I have some ideas involving `citra` but it might take a bit of work

For now, we should probably keep all the examples as ` ```no_run `, and try adding something to CI to make sure they at least compile. I might open a separate issue for the runner itself, since I think that will be a bit more involved and might even require changes to upstream `citra` to be effective.

ian-h-chamberlain:
With lots of doctests in #134 I've marked this "In Progress", at least. I haven't had a chance to revisit the `citra` runner to really test everything though, so I'd say let's leave it open for now.

Meziu:
I've noticed the CI still needs a little boost, and unit tests are very much lacking, so this issue needs to stay open for a while longer.

ian-h-chamberlain:
I've made some progress on a proof of concept for running tests using Citra: 
https://github.com/ian-h-chamberlain/test-runner-3ds

So far I can get it to work with integration + unit tests, but it's not running doctests yet.

Example of passing: https://github.com/ian-h-chamberlain/test-runner-3ds/actions/runs/5781441190/job/15666535932
and failing: https://github.com/ian-h-chamberlain/test-runner-3ds/actions/runs/5781347357/job/15666313158

[...]

PR #109, opened 2023-04-03
Set up reusable CI workflow and add building doctests

ian-h-chamberlain:
Part of #106 

This does a couple things:
* Makes sure we are actually building doctests now
* Switch to using https://github.com/actions-rust-lang/setup-rust-toolchain so we get nice inline warnings and caching
* Break out some common logic into a reusable composite action. I hope this can also be used for other crates going forward.

- [x] TODO update the git hash once we merge https://github.com/rust3ds/cargo-3ds/pull/32


Meziu (approved):
This is great! Thanks for setting up the CI this way. The more automated checks we have, the easier it is to maintain the repo 👍

.github/workflows/ci.yml
```diff
@@ -68,5 +53,28 @@ jobs:
...
+    steps:
+      - name: Checkout branch
+        uses: actions/checkout@v2
+
+      - uses: ./.github/actions/setup
+        with:
+          toolchain: ${{ matrix.toolchain }}
+
+      - name: Hide duplicated warnings from lint job
+        run: echo "::remove-matcher owner=clippy::"
```
AzureMarker, line 74: Is this needed for the doc test pipeline?

AzureMarker (approved):

.github/workflows/ci.yml
ian-h-chamberlain: yeah, the matcher is added during setup by `actions-rust-lang/setup-rust-toolchain` and it's called `clippy` even though it also encompasses normal `rustc` diagnostics. In practice that means any compiler warnings would get duplicated here when the library is built by `cargo 3ds test --doc`, without this `::remove-matcher` in place.

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,48 @@
...
+      run: sudo apt-get update && sudo apt-get install -y build-essential
+
+    - name: Install cargo-3ds
+      uses: actions-rs/cargo@v1
+      with:
+        command: install
+        # TODO: this should probably just be a released version from crates.io
+        # once cargo-3ds gets published somewhere...
+        #
+        # also switch to master hash once feature/doctests gets merged
```
Meziu:
```suggestion
```

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,48 @@
...
+        args: >-
+          --git https://github.com/rust3ds/cargo-3ds
+          --rev 5a06e64307499ffe681453e7d88d7ee986054eb7
```
Meziu:
```suggestion
          --rev 78a652fdfb01e2614a792d1a56b10c980ee1dae9
```

Meziu (approved):
I’ve made the suggestions to update this PR and merge it.

PR #110, opened 2023-04-04
Add `bindgen-ctru-sys` package and update bindings

Techie-Pi:
I've used the [citro3d-rs](https://github.com/rust3ds/citro3d-rs) repo as _inspiration_ for creating the [script](https://github.com/rust3ds/citro3d-rs/tree/main/bindgen-citro3d). This doesn't bring any improvements _per se_, but I think this is a cleaner approach :)

This also updates `doxygen-rs` to 0.4, which has a new engine that will allow me to more easily start supporting the things that are left. Right now there are some regressions, but the documentations is still completely understandable.

Also, thanks to the use of a package to create the bindings, we're able to use `ParseCallbacks` (which is great for the Doxygen conversions, because the previous approach was quite _hacky_ to say the least...)

Related #32 #75 https://github.com/rust3ds/citro3d-rs/pull/10


ctru-sys/bindgen-ctru-sys/src/main.rs
```diff
@@ -0,0 +1,61 @@
...
+            sysroot.to_str().unwrap(),
+            "-isystem",
+            system_include.to_str().unwrap(),
+            "-I",
+            include_path.to_str().unwrap(),
+            "-mfloat-abi=hard",
+            "-march=armv6k",
+            "-mtune=mpcore",
+            "-mfpu=vfp",
+            "-DARM11 ",
```
ian-h-chamberlain: are the trailing spaces here intentional? I'm not sure how clang behaves here.

Also, I think as of [libctru 2.1.0](https://github.com/devkitPro/libctru/releases/tag/v2.1.0) only `__3DS__` should be needed, not `_3DS`. But probably no harm in having both

Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-members = ["ctru-rs", "ctru-sys", "ctru-sys/docstring-to-rustdoc"]
+members = ["ctru-rs", "ctru-sys", "ctru-sys/bindgen-ctru-sys"]
```
ian-h-chamberlain, line 2: you might need to set `default-members = ["ctru-rs", "ctru-sys"]` to avoid trying to build the bindgen package for 3DS as part of CI, I think I had to do that in `citro3d` so that `cargo 3ds check --workspace` still worked.

ctru-sys/bindgen.sh
ian-h-chamberlain, line 1: As more of this is converted to Rust now I almost wonder if we should consider writing the other logic that I just added for `libctru` version numbering into `bindgen-ctru-sys` as well... Doesn't need to be part of this PR just something to think about

ian-h-chamberlain (approved):
Looks good to me! Thanks, the new docs seem nicer for sure!

ctru-sys/bindgen-ctru-sys/src/main.rs
Techie-Pi: I took a look at the code from citro-3d-bindgen, the shell script and the main.rs had the same params so I just put the same 😄

So, yeah, I just took it from citro-3d

ctru-sys/bindgen.sh
Techie-Pi: Yeah, that would be great. That would also allow support for multiple platforms (right now I'm using Windows and I had to run the commands manually because the script couldn't determine my version due to the need of pacman, which of course I don't have)

ian-h-chamberlain: Ah man, I had been wondering about Windows support with that script, but don't have a Windows environment handy for testing it (and couldn't get Wine to work either).
I opened #113 to track this.

Meziu:
A new release of libctru has been released: https://github.com/devkitPro/libctru/releases/tag/v2.2.0

You should update the bindings and change the version of `ctru-sys` to correspond to libctru.

Meziu:
As of #101 we also need to bump the `ctru-sys` version when pulling a new `libctru`. As such, you should change the version of `ctru-sys` to `22.0.0+2.2.0` in its `Cargo.toml` file.

ctru-sys/bindgen-ctru-sys/src/main.rs
Meziu: > I think as of [libctru 2.1.0](https://github.com/devkitPro/libctru/releases/tag/v2.1.0) only `__3DS__` should be needed, not `_3DS`. But probably no harm in having both

We should only ask for `__3DS__`, in my opinion. The other definition isn't used anymore, so we won't lose anything, it's only more confusing. Let's just stick to the standard.

Also, the trailing spaces should be removed, I don't think they break anything, but it's not pretty.

ctru-sys/src/bindings.rs
```diff
@@ -4736,27 +4772,119 @@ impl Default for pthread_attr_t {
...
+#[derive(Debug, Default, Copy, Clone)]
+pub struct pthread_barrierattr_t {}
+#[repr(C)]
+#[derive(Debug, Default, Copy, Clone)]
+pub struct pthread_rwlock_t {
+    pub lock: _LOCK_T,
+    pub cond_r: _COND_T,
+    pub cond_w: _COND_T,
+    pub _bitfield_align_1: [u32; 0],
+    pub _bitfield_1: __BindgenBitfieldUnit<[u8; 4usize]>,
```
ian-h-chamberlain: huh, lots of new `pthread` related changes in these bindings... I wonder if this will break / can replace some of our `pthread-3ds` implementation...

ian-h-chamberlain: Looking into it more... we might be better off not upgrading `libctru`/`newlib` just yet: https://github.com/rust3ds/pthread-3ds/issues/26

Meziu (commented):
@ian-h-chamberlain `__getreent` has been moved to `newlib`. As such, it doesn’t appear in the bindings anymore. Since it’s not available on `libc` either, maybe we should add it to ctru-sys’ `lib.rs`?

ian-h-chamberlain:
> @ian-h-chamberlain `__getreent` has been moved to `newlib`. As such, it doesn’t appear in the bindings anymore. Since it’s not available on `libc` either, maybe we should add it to ctru-sys’ `lib.rs`?

Hmmm, we could, I suppose... It looks like the only reason we use it today is for `errno`, which we might be able to work around by fixing https://github.com/rust-lang/libc/issues/1995 upstream, or maybe just by providing our own binding like this:
```rs
extern "C" {
	unsafe fn __errno() -> *mut libc::c_int;
}
```

Or we could even just include `errno.h` when generating bindings, from a quick test it looks like it would add a bunch of constants like `ENOSYS` etc., plus this small diff:
```diff
@@ -22929,3 +23016,13 @@ extern "C" {
     #[doc = ""]
     pub fn link3dsConnectToHost(redirStdout: bool, redirStderr: bool) -> ::libc::c_int;
 }
+pub type error_t = ::libc::c_int;
+extern "C" {
+    pub fn __errno() -> *mut ::libc::c_int;
+}
+extern "C" {
+    pub static _sys_errlist: [*const ::libc::c_char; 0usize];
+}
+extern "C" {
+    pub static mut _sys_nerr: ::libc::c_int;
+}
```

PR #112, opened 2023-04-05
Fix licensing and credits

Meziu:
With these changes the original authors are credited directly within the README and the license has been moved to a dedicated LICENSE file. Also, the Rust3DS Org is directly credited as author of the package, together with the original names.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,33 +24,10 @@ This repository is organized as follows:
   (displayed when built with `-vv`) will be issued if the build script detects
   a mismatch or is unable to check the installed `libctru` version.
 
-## License
+## Original version
 
-Copyright (C) Rust 3DS Project authors, 2015-2016
+This project is based on the efforts the original authors:
+ * [Eidolon](https://github.com/HybridEidolon)
+ * [FenrirWolf](https://github.com/FenrirWolf)
 
-See AUTHORS.md.
-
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -24,33 +24,10 @@ This repository is organized as follows:
   (displayed when built with `-vv`) will be issued if the build script detects
   a mismatch or is unable to check the installed `libctru` version.
 
-## License
+## Original version
 
-Copyright (C) Rust 3DS Project authors, 2015-2016
```
ian-h-chamberlain, line 29: Do we need to retain this copyright header in the LICENSE file? I am not a lawyer but in most cases I've seen you are meant to retain original copyright headers and just add onto them?

Maybe it doesn't apply here since it's a derivative of `ctru-rs-old` rather than a direct copy.

ian-h-chamberlain (approved):

README.md
Meziu: Wait, does anyone actually hold a copyright for this name? Don’t you need to file a copyright ©️ for it to have effect?

Techie-Pi: You don't need to file anything to Have the copyright of a creation. The moment you create any work, the copyright is yours. And yes, you should retain the original headers

PR #115, opened 2023-04-05
Fix mutability requirements

Meziu:
Closes #111

Many modules didn't properly require internal mutability checks. This should fix all of the cases.

PR #116, opened 2023-04-06
Second wave of API changes

Meziu:
Related to #102 

This is a second wave of breaking API changes based on the official [guidelines](https://rust-lang.github.io/api-guidelines/checklist.html) (I can hear you cry from here).

Most notable changes:
- All service `init` functions have been renamed to `new` (it's standard for any constructor).
- Types derive as many traits as possible (each was thought out for the type's specific purpose, of course)
- Small renames
- `Gfx::swap_buffers` and `Gfx::flush_buffers` have been moved to the `Screen` trait. Originally, those functions modified a global state. Now each `Screen` can flush its own buffers. Also, `Console` handles flushing and swapping by itself, so most examples don't even mention those functions anymore.

This is one of the last PRs in preparation to the docs PR.

### DIFF
--- a/ctru-rs/examples/gfx-3d-mode.rs
+++ b/ctru-rs/examples/gfx-3d-mode.rs
@@ -61,6 +61,12 @@ fn main() {
             buf.copy_from(IMAGE.as_ptr(), IMAGE.len());
         }
 
+        left.flush_buffer();
+        left.swap_buffers();
+
+        right.flush_buffer();
+        right.swap_buffers();
+
         //Wait for VBlank
         gfx.wait_for_vblank();
     }
--- a/ctru-rs/examples/gfx-3d-mode.rs
+++ b/ctru-rs/examples/gfx-3d-mode.rs
@@ -61,10 +61,6 @@ fn main() {
             buf.copy_from(IMAGE.as_ptr(), IMAGE.len());
         }
 
-        // Flush and swap framebuffers
-        gfx.flush_buffers();
-        gfx.swap_buffers();
-
         //Wait for VBlank
         gfx.wait_for_vblank();
     }
--- a/ctru-rs/examples/gfx-3d-mode.rs
+++ b/ctru-rs/examples/gfx-3d-mode.rs
@@ -62,8 +62,7 @@ fn main() {
             buf.copy_from(IMAGE.as_ptr(), IMAGE.len());
         }
 
-        drop(left);
-        drop(right);
+        drop((left, right));
 
         top_screen.flush_buffers();
         top_screen.swap_buffers();
--- a/ctru-rs/examples/gfx-3d-mode.rs
+++ b/ctru-rs/examples/gfx-3d-mode.rs
@@ -12,10 +12,10 @@ static ZERO: &[u8] = &[0; IMAGE.len()];
 fn main() {
     ctru::use_panic_handler();
 
-    let gfx = Gfx::init().expect("Couldn't obtain GFX controller");
-    let mut hid = Hid::init().expect("Couldn't obtain HID controller");
-    let apt = Apt::init().expect("Couldn't obtain APT controller");
-    let _console = Console::init(gfx.bottom_screen.borrow_mut());
+    let gfx = Gfx::new().expect("Couldn't obtain GFX controller");
+    let mut hid = Hid::new().expect("Couldn't obtain HID controller");
+    let apt = Apt::new().expect("Couldn't obtain APT controller");
+    let _console = Console::new(gfx.bottom_screen.borrow_mut());
 
     println!("Press Start to exit.\nPress A to switch sides (be sure to have 3D mode enabled).");
 
--- a/ctru-rs/examples/gfx-3d-mode.rs
+++ b/ctru-rs/examples/gfx-3d-mode.rs
@@ -64,9 +64,6 @@ fn main() {
         left.flush_buffer();
         left.swap_buffers();
 
-        right.flush_buffer();
-        right.swap_buffers();
-
         //Wait for VBlank
         gfx.wait_for_vblank();
     }
--- a/ctru-rs/examples/gfx-3d-mode.rs
+++ b/ctru-rs/examples/gfx-3d-mode.rs
@@ -1,5 +1,5 @@
 use ctru::prelude::*;
-use ctru::services::gfx::{Screen, Side, TopScreen3D};
+use ctru::services::gfx::{Screen, Side, Swap, TopScreen3D};
 
 /// See `graphics-bitmap.rs` for details on how the image is generated.
 ///
@@ -21,8 +21,7 @@ fn main() {
 
     gfx.top_screen.borrow_mut().set_double_buffering(true);
 
-    let top_screen = TopScreen3D::from(&gfx.top_screen);
-    let (mut left, mut right) = top_screen.split_mut();
+    let mut top_screen = TopScreen3D::from(&gfx.top_screen);
 
     let mut current_side = Side::Left;
 
@@ -34,6 +33,8 @@ fn main() {
         if hid.keys_down().contains(KeyPad::START) {
             break;
         }
+
+        let (mut left, mut right) = top_screen.split_mut();
 
         let left_buf = left.raw_framebuffer();
         let right_buf = right.raw_framebuffer();
@@ -61,8 +62,11 @@ fn main() {
             buf.copy_from(IMAGE.as_ptr(), IMAGE.len());
         }
 
-        left.flush_buffer();
-        left.swap_buffers();
+        drop(left);
+        drop(right);
+
+        top_screen.flush_buffers();
+        top_screen.swap_buffers();
 
         //Wait for VBlank
         gfx.wait_for_vblank();
--- a/ctru-rs/examples/hello-world.rs
+++ b/ctru-rs/examples/hello-world.rs
@@ -5,10 +5,10 @@ use std::io::BufWriter;
 fn main() {
     ctru::use_panic_handler();
 
-    let gfx = Gfx::init().expect("Couldn't obtain GFX controller");
-    let mut hid = Hid::init().expect("Couldn't obtain HID controller");
-    let apt = Apt::init().expect("Couldn't obtain APT controller");
-    let _console = Console::init(gfx.top_screen.borrow_mut());
+    let gfx = Gfx::new().expect("Couldn't obtain GFX controller");
+    let mut hid = Hid::new().expect("Couldn't obtain HID controller");
+    let apt = Apt::new().expect("Couldn't obtain APT controller");
+    let _console = Console::new(gfx.top_screen.borrow_mut());
 
     let out = b"Hello fellow Rustaceans, I'm on the Nintendo 3DS!";
     let width = 24;
--- a/ctru-rs/examples/hello-world.rs
+++ b/ctru-rs/examples/hello-world.rs
@@ -29,9 +29,6 @@ fn main() {
         if hid.keys_down().contains(KeyPad::START) {
             break;
         }
-        // Flush and swap framebuffers
-        gfx.flush_buffers();
-        gfx.swap_buffers();
 
         //Wait for VBlank
         gfx.wait_for_vblank();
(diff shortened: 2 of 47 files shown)


ctru-rs/src/services/gfx.rs
```diff
@@ -55,6 +55,28 @@ pub trait Screen: private::Sealed {
...
+                (framebuffer.height * framebuffer.width) as u32,
+            )
+        };
+    }
+
+    /// Swaps the video buffers.
+    ///
+    /// This should be used even if double buffering is disabled.
+    fn swap_buffers(&mut self) {
+        unsafe { ctru_sys::gfxScreenSwapBuffers(self.side().into(), true) };
```
ian-h-chamberlain: ah nice, I was wondering how these would work since the `gfxFlushBudders` etc were global, I didn't realize there are APIs for doing it per screen.

Should we set the `hasStereo` parameter to `false` here by default, and set to `true` only if the screen in question is the "combined" top screen?

ctru-rs/src/services/gfx.rs
```diff
@@ -96,14 +118,14 @@ pub struct RawFrameBuffer<'screen> {
     /// Pointer to graphics data to be rendered.
     pub ptr: *mut u8,
     /// The width of the framebuffer in pixels.
-    pub width: u16,
+    pub width: usize,
     /// The height of the framebuffer in pixels.
-    pub height: u16,
+    pub height: usize,
```
ian-h-chamberlain, line 186: IIRC these were `u16` since the maximum value is much smaller than `usize::MAX` but it's probably fine to have like this as long as there is some check for max values somewhere

ctru-rs/src/services/gfx.rs
```diff
@@ -55,6 +55,28 @@ pub trait Screen: private::Sealed {
         unsafe { ctru_sys::gfxSetDoubleBuffering(self.as_raw(), enabled) }
     }
 
+    /// Swaps the video buffers.
```
ian-h-chamberlain:
```suggestion
    /// Flushes the video buffer for this screen.
```

ctru-rs/src/services/cfgu.rs
```diff
@@ -33,15 +33,15 @@ pub enum Language {
...
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
 #[repr(u32)]
 pub enum SystemModel {
-    Model3DS = ctru_sys::CFG_MODEL_3DS,
-    Model3DSXL = ctru_sys::CFG_MODEL_3DSXL,
-    ModelNew3DS = ctru_sys::CFG_MODEL_N3DS,
-    Model2DS = ctru_sys::CFG_MODEL_2DS,
-    ModelNew3DSXL = ctru_sys::CFG_MODEL_N3DSXL,
-    ModelNew2DSXL = ctru_sys::CFG_MODEL_N2DSXL,
+    N3DS = ctru_sys::CFG_MODEL_3DS,
```
ian-h-chamberlain: This might be slightly confusing since sometimes I have seen `N3DS` used to mean `New 3DS`, but I suppose the fact that there is a separate `NewN3DS` should make that clear. Don't have a better suggestion really

ctru-rs/src/applets/swkbd.rs
```diff
@@ -7,6 +7,7 @@ use std::iter::once;
 use std::str;
 
 /// An instance of the software keyboard.
+#[derive(Clone)]
```
ian-h-chamberlain, line 10: does cloning the inner Box here end up cloning all the global state too? I would guess normally we don't want `Clone` for these "global-ish" kind of structs...

ctru-rs/src/applets/mii_selector.rs
```diff
@@ -8,14 +8,14 @@ use std::ffi::CString;
 
 /// Index of a Mii used to configure some parameters of the Mii Selector
 /// Can be either a single index, or _all_ Miis
-#[derive(Debug, Clone)]
-pub enum MiiConfigIndex {
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum MiiIndex {
```
ian-h-chamberlain: this could probably even just be called `mi_selector::Index`, same for other structs in this module.

Or actually, we could call the module `mii` and then you'd have `mii::Type`, `mii::Selector`, etc.

ctru-rs/examples/hello-world.rs
```diff
@@ -29,9 +29,6 @@ fn main() {
         if hid.keys_down().contains(KeyPad::START) {
             break;
         }
-        // Flush and swap framebuffers
-        gfx.flush_buffers();
-        gfx.swap_buffers();
```
ian-h-chamberlain, line 34: you mentioned that the console handles flushing the screen automatically... does it do that whenever the stdout is flushed (e.g. on `\n` or explicit `flush()` calls)? 

I'd imagine that for most applications (non-console) the user would still want to flush + swap buffers every frame instead of letting underlying helpers do it at random times.

Edit: also, from looking at https://github.com/devkitPro/libctru/blob/master/libctru/source/console.c it seems the buffers are flushed by default but not normally swapped, which I guess doesn't matter since `consoleInit` also sets double-buffering off...

ctru-rs/examples/gfx-3d-mode.rs
```diff
@@ -61,9 +61,8 @@ fn main() {
             buf.copy_from(IMAGE.as_ptr(), IMAGE.len());
         }
 
-        // Flush and swap framebuffers
-        gfx.flush_buffers();
-        gfx.swap_buffers();
+        left.flush_buffer();
+        left.swap_buffers();
```
ian-h-chamberlain: this probably should be flushing whichever side is being drawn to, not just the left screen?

ian-h-chamberlain (commented):
Mostly looks good! I had some questions about the Gfx changes but I think everything else looks good.

Also you may want to merge `master` back into this branch to make the diff a bit simpler (without displaying the changes from #115 )

ctru-rs/src/services/gfx.rs
Meziu: That would require quite a bit of work on the `TopScreen3D` struct, since it doesn't implement `Screen`. Since you built it, do you have any idea how we could do so?

ctru-rs/src/services/gfx.rs
Meziu: Yes, it's always smaller that `u16::MAX`. However, since this is a "read-only" member (there is no function that asks for these sizes, right?) there is no problem using it this way. The user will convert it back to `usize` 99% of times since that needed to work with pointers. It's just easier to have it like this.

ctru-rs/src/services/gfx.rs
Meziu: Copying and pasting is always a bad habit... thanks for noticing (though I will look into all docs later on) 👍

ctru-rs/examples/gfx-3d-mode.rs
Meziu: The issue is with the "stereo" parameter. Since the screen is in 3D mode, and the `TopRight` screen returns "top" as its raw value, the swap/flush would be done twice on the top screen, once when "Flushing left" and once when "flushing right" (it flushes both of them each time). I guess that means this function should live in `TopScreen3D`, but that doesn't implement `Screen`, so it's a bit of an issue.

ctru-rs/examples/hello-world.rs
Meziu: We could specify that behaviour in a comment, since there is no need for manually flushing. Console does it on `print()`. I'll put this off for the docs PR. 

Edit: Overall this goes nicely with our "screen ownership" model: `Console` has it, `Console` manages it. If the user has a different screen (like in `graphics-bitmap`) they can manage that.

ctru-rs/src/applets/mii_selector.rs
Meziu: The mii module already exists, lol. I can rename it to settle for a better `module::struct` nomenclature.

ctru-rs/src/applets/swkbd.rs
Meziu: Today in "Wacky facts about libctru" we will learn: Software Keyboards!

Did you know? The "SwkbdState" struct only holds config data, since the Software Keyboard is "spawned" on request, and is not a global service.

Next episode, we will find out why internal atomic counters aren't accessible via the API (small spoiler: there is no reason! 🥳 )

ctru-rs/src/applets/mii_selector.rs
ian-h-chamberlain: ahh I didn't realize that, maybe it should just be `mii::selector::Index` etc then

ctru-rs/examples/gfx-3d-mode.rs
ian-h-chamberlain: Hmm, after rereading the documentation for the `libctru` functions I see what you mean. I guess in most cases people will want to swap both sides' buffers at the same time, so it's probably fine.

I guess we could awkwardly make the `TopLeft` and `TopRight` impls do nothing and force the user to call a special `flush_both` type method on the `TopScreen3D`, but I think we will run into mutability issues etc, plus having a method do nothing feels kinda wrong to me.

Another option is to define a totally new trait, maybe something like this:
```rs
pub trait Flush: private::Sealed {
    fn swap_buffers(&mut self);
    fn flush_buffer(&mut self);
}
```

I think we'd have to manually impl it for `TopScreen`, `BottomScreen` and `TopScreen3D` (which would have to flush both left + right sides). I'm not sure if there is a use case for flushing left/right separately, but if needed we could also split `Flush` and `Swap` to separate traits as well.

ctru-rs/src/services/gfx.rs
ian-h-chamberlain: See my other comment, I think we might be able to do it with a separate trait instead of having these methods on `Screen`, but it might take a bit of experimenting, so maybe better off in a separate PR.

ian-h-chamberlain (approved):

ctru-rs/examples/gfx-3d-mode.rs
ian-h-chamberlain: Update: I opened #118 to show what I am thinking of. We can discuss further on that PR

ctru-rs/src/applets/mii_selector.rs
Techie-Pi: Yeah, `mii:selector` seems like a good option tbh

ctru-rs/src/services/sslc.rs
```diff
@@ -8,23 +8,12 @@ pub struct SslC(());
...
-    pub fn init() -> crate::Result<Self> {
+    pub fn new() -> crate::Result<Self> {
         unsafe {
             ResultCode(ctru_sys::sslcInit(0))?;
             Ok(SslC(()))
         }
     }
-
-    /// Fill `buf` with `buf.len()` random bytes
-    pub fn generate_random_data(&self, buf: &mut [u8]) -> crate::Result<()> {
```
AzureMarker, line 19: Is this supposed to get removed?

ctru-rs/src/applets/mii_selector.rs
Meziu: That would move the module out of the `applets` folder. Do you think that would be a good idea? There are many more applets left to be implemented.

ctru-rs/src/services/sslc.rs
Meziu: This is a duplicate of `Ps::generate_random_data`. The user should use that one or the `rand` crate, which works via `getrandom`.

ctru-rs/src/services/cfgu.rs
Meziu: I went with Old3DS. It's not official but people recognize it and it's much less confusing.

ctru-rs/src/applets/mii_selector.rs
Meziu: I simply renamed it to `Index` for now. Tell me if you think more changes are needed.

ian-h-chamberlain (approved):

AzureMarker (approved):

PR #117, opened 2023-04-08
Compilation error when romfs can't be found

### LINKED ISSUE #108: Compile error if romfs is enabled but missing
at some point we might consider adding something like
```rs
} else if #[cfg(all(feature = "romfs", not(romfs_exists)))] {
    compile_error!("romfs feature is enabled but no romfs found!");
}
```

haven't tested to see if that does what I think though

_Originally posted by @ian-h-chamberlain in https://github.com/rust3ds/ctru-rs/pull/107#discussion_r1154897602_

Meziu:
Closes #108 

This simply adds a compilation error if the feature is enabled, but the dir isn't properly set. Technically the end user would see compilation issues regardless (no `Romfs` struct found), but this way it's a bit more clear.

### DIFF
--- a/ctru-rs/src/services/mod.rs
+++ b/ctru-rs/src/services/mod.rs
@@ -34,7 +34,7 @@ cfg_if::cfg_if! {
             //! [package.metadata.cargo-3ds]
             //! romfs_dir = "romfs"
             //! ```
-            
+
             compile_error!("romfs feature is enabled but no romfs found!");
         }
     }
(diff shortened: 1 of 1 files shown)


Techie-Pi (approved):

ctru-rs/src/services/mod.rs
```diff
@@ -34,6 +34,8 @@ cfg_if::cfg_if! {
             //! [package.metadata.cargo-3ds]
             //! romfs_dir = "romfs"
             //! ```
+
+            compile_error!("romfs feature is enabled but no romfs found!");
```
AzureMarker, line 40: Won't this fail compilation for any program not using ROMFS?

Meziu: Yes, it would have. It's been fixed.

AzureMarker (approved):

PR #118, opened 2023-04-12
Create some traits for swapping/flushing screen buffers

ian-h-chamberlain:
Coming from https://github.com/rust3ds/ctru-rs/pull/116#discussion_r1164231506

I tried to implement this, it seems to work mostly for compile time checks at least, but I haven't tested that it actually works as expected when running. Let me know what you think, maybe this is a slightly better API to make it clearer that both left + right top screen buffers have to be swapped at once?

### DIFF
--- a/ctru-rs/examples/gfx-3d-mode.rs
+++ b/ctru-rs/examples/gfx-3d-mode.rs
@@ -61,6 +61,12 @@ fn main() {
             buf.copy_from(IMAGE.as_ptr(), IMAGE.len());
         }
 
+        left.flush_buffer();
+        left.swap_buffers();
+
+        right.flush_buffer();
+        right.swap_buffers();
+
         //Wait for VBlank
         gfx.wait_for_vblank();
     }
--- a/ctru-rs/examples/gfx-3d-mode.rs
+++ b/ctru-rs/examples/gfx-3d-mode.rs
@@ -1,5 +1,5 @@
 use ctru::prelude::*;
-use ctru::services::gfx::{Screen, Side, Swap, TopScreen3D};
+use ctru::services::gfx::{Flush, Screen, Side, Swap, TopScreen3D};
 
 /// See `graphics-bitmap.rs` for details on how the image is generated.
 ///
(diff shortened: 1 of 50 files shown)


ctru-rs/src/services/gfx.rs
```diff
@@ -247,7 +296,7 @@ impl TopScreen {
 
 impl Screen for TopScreen {
     fn as_raw(&self) -> ctru_sys::gfxScreen_t {
-        self.left.as_raw()
+        ctru_sys::GFX_TOP
```
ian-h-chamberlain: oops, this was probably actually fine before, but I thought it was another mistake use of implicit conversion from `gfx3dSide_t` to `gfxScreen_t`.

AzureMarker (approved):

ctru-rs/examples/gfx-3d-mode.rs
```diff
@@ -61,8 +62,10 @@ fn main() {
             buf.copy_from(IMAGE.as_ptr(), IMAGE.len());
         }
 
-        left.flush_buffer();
-        left.swap_buffers();
+        drop((left, right));
```
Meziu, line 65: Hmm, dropping the sides isn’t really a great thing. I understand that (especially when the double buffer swaps) the written framebuffer would be different, however, this will create some issues when more complex structs like those we could write in `citro3d-rs` take the sides as arguments.

Meziu:
I don’t like this solution. My issues mainly come down to 2 things:
1. It’s unclear why `TopScreen3D` wouldn’t implement `Screen` (it’s in the name).
2. The new traits are pretty much useless by themselves. There is no situation in which the user (or even us, for that matter) could use them anywhere. (E.g. why would you want to only “flush” a screen, but not control its framebuffer?).

Maybe we should take a different approach, and I believe that should come from redefining the meaning of the `Screen` trait. As the name suggests, the `Screen` should hold information and methods about controlling the physical LCD. This includes: changing framebuffer format, flushing, swapping and setting the double buffering.

Then, we would have a different trait, called something like `Side` (though I find this name confusing for the bottom screen) or `RenderSurface` which would hold information about the “virtual” rendering, like: getting the raw framebuffer data, the raw `Side` data (which would be “left” most of the time).

Each screen will return the respective `RenderSurface`s,  like `TopScreen3D` already does.

Tell me if you find this proposal acceptable.

PR #119, opened 2023-05-02
add touch screen example

adryzz:
I realized i made a project using the touchscreen but there wasn't any example, so i made it.

The console on this example updates very quickly when moving the stylus or finger, but it's probably fine.


Meziu:
You made me notice there still where some remnants of the old version which behaved in an unpredictable way. To accomodate the new changes, I've implemented your example in #116. Thanks a lot for the help!

adryzz:
yeah it did seem kinda off-brand for touch input to work this way tbh...

anyway that's great! thank you for all your work, this project is insane.

PR #122, opened 2023-05-08
Handle empty Mii selector return

### LINKED ISSUE #121: Mii Selector returns garbage result in Ok
On my device, running examples/mii-selector.rs produces a popup window that says "There are no selectable miis" and "Press (A) to go back". Pressing A closes the popup but the program continues without panicking at `let result = mii_selector.launch().unwrap();`, ~~and instead it appears that an empty Box is interpreted as valid MiiData instead~~. is_mii_selected is false, but the program still uses the MiiData object. ~~I believe this is UB in safe code (pointer to garbage data)~~. For example it says the Mii is of type Guest and the "name" field of the Guest variant is shown as a bunch of `\0` null chars. The checksum test doesn't fix this because the CRC of all zeroes is also 0.

## Suggestion
One of the following changes:
1. `MiiSelector::launch` changed to return `Result<Option<SelectionResult>, LaunchError>`
2. Add a variant `Cancelled` to `LaunchError`

In either case,  `is_mii_selected` can be removed/deprecated (as it should always be `true` if a SelectionResult has been created!).

DeltaF1:
In some cases the Mii selector can fail even if `MII_SELECTOR_CANCEL` is not set. This PR accounts for this possibility and makes it so the user is never passed fake Mii data.

Closes #121

### DIFF
--- a/ctru-rs/src/applets/mii_selector.rs
+++ b/ctru-rs/src/applets/mii_selector.rs
@@ -65,7 +65,10 @@ pub struct SelectionResult {
 /// Error type for the Mii selector
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 pub enum LaunchError {
+    /// The selected Mii's data is corrupt in some way
     InvalidChecksum,
+
+    /// Either the user cancelled the selection (see [Options::MII_SELECTOR_CANCEL]), or no valid Miis were available to select
     NoMiiSelected,
 }
 
--- a/ctru-rs/src/applets/mii_selector.rs
+++ b/ctru-rs/src/applets/mii_selector.rs
@@ -56,9 +56,6 @@ pub struct MiiSelector {
 #[derive(Clone, Debug)]
 pub struct SelectionResult {
     pub mii_data: MiiData,
-
-    #[deprecated]
-    pub is_mii_selected: bool,
     pub mii_type: MiiType,
 }
 
@@ -67,7 +64,6 @@ pub struct SelectionResult {
 pub enum LaunchError {
     /// The selected Mii's data is corrupt in some way
     InvalidChecksum,
-
     /// Either the user cancelled the selection (see [Options::MII_SELECTOR_CANCEL]), or no valid Miis were available to select
     NoMiiSelected,
 }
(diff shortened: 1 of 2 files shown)


ctru-rs/src/applets/mii_selector.rs
```diff
@@ -56,14 +56,20 @@ pub struct MiiSelector {
...
     pub is_mii_selected: bool,
     pub mii_type: MiiType,
 }
 
 /// Error type for the Mii selector
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 pub enum LaunchError {
+    /// The selected Mii's data is corrupt in some way
     InvalidChecksum,
+
```
Meziu: Remove the empty space between variants. Even with doc comments we don't put newlines in between.

ctru-rs/src/applets/mii_selector.rs
```diff
@@ -56,14 +56,20 @@ pub struct MiiSelector {
 #[derive(Clone, Debug)]
 pub struct SelectionResult {
     pub mii_data: MiiData,
+
+    #[deprecated]
     pub is_mii_selected: bool,
```
Meziu: I suggest to simply remove the `is_mii_selected` attribute altogether. As @Techie-Pi said, it's a remnant of a 1 to 1 translation of the C API, but we don't need it here.

Also, we never released the library anyways, so there is no reason to use `#[deprecated]`. We are undergoing big API changes and it's currently very unstable either way.

Meziu (changes requested):
Looks great! Better error handling is why we made this Rust library in the first place 😄

ctru-rs/src/applets/mii_selector.rs
DeltaF1: Sure, will do. Do I need to do something to update a version number or anything?

Meziu (approved):

Techie-Pi (approved):

ctru-rs/src/applets/mii_selector.rs
```diff
@@ -56,14 +56,16 @@ pub struct MiiSelector {
 #[derive(Clone, Debug)]
 pub struct SelectionResult {
     pub mii_data: MiiData,
-    pub is_mii_selected: bool,
     pub mii_type: MiiType,
 }
 
 /// Error type for the Mii selector
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
```
ian-h-chamberlain, line 63: doesn't need to be in this PR but we probably should make this `#[non_exhaustive]` as well

ian-h-chamberlain (approved):

PR #126, opened 2023-05-24
Remove feature gate for stabilized feature (stabilized in 1.70.0)

DeltaF1:
Remove feature gate for `nonnull_slice_from_raw_parts`

```rs
warning: the feature `nonnull_slice_from_raw_parts` has been stable since 1.70.0 and no longer requires an attribute to enable
 --> ctru-rs/src/lib.rs:7:12
  |
7 | #![feature(nonnull_slice_from_raw_parts)]
  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(stable_features)]` on by default
```

### DIFF
--- a/ctru-rs/src/lib.rs
+++ b/ctru-rs/src/lib.rs
@@ -4,7 +4,6 @@
 #![feature(custom_test_frameworks)]
 #![feature(try_trait_v2)]
 #![feature(allocator_api)]
-#![feature(nonnull_slice_from_raw_parts)]
 #![test_runner(test_runner::run)]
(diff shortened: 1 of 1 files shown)


DeltaF1:
Ah I see now that the project targets a specific older nightly version as well. Good to know.

Meziu:
Yeah it’s a bit annoying when using the newer toolchains, but it’s fine. We will update the minimum version once a good chunk of the unstable features we use get stabilised.

PR #133, opened 2023-07-26
Support static inline functions in ctru-sys

### LINKED ISSUE #123: Expose static inline functions to bindings
libctru has some functions that are `static inline` (I ran into [`getThreadCommandBuffer`](https://libctru.devkitpro.org/svc_8h.html#a2e0c230a1bb2d5400edc08bd7c687099) specifically) and so are not exposed by bindgen by default.

https://github.com/rust-lang/rust-bindgen/pull/2335 adds a feature to generate bindings for such functions. I got as far as adding these arguments to bindgen.sh:
```sh
    --experimental \
    --wrap-static-fns \

```
which adds the missing functions to bindings.rs, and generates a C file with wrapper functions in `/tmp/bindgen/extern.c`. I don't understand build processes enough to proceed from this point, but `extern.c` needs to be compiled and then linked with the library. [Here's a link](https://github.com/rust-lang/rust-bindgen/discussions/2405) discussing how to do this and also about how to enable LTO so that ctru-rs can benefit from inlining those functions. I believe this also depends upon #110 fixing the `__getreent` issue.

FenrirWolf:
A first attempt at generating bindings for static inline functions in libctru. I'm pretty sure this functions properly, but I haven't played around with LTO settings yet and I'm sure there are some hand-written wrappers that will need to be replaced with calls to the generated fns instead.

Also now that we have to compile, build, and link a whole new C static lib as part of the process, it might be cleaner to move everything to a proper build script instead of pre-generating the ctru-sys bindings like we've been doing up until now. Honestly the only reason I set things up like that back in the day is because I didn't like having to wait for `bindgen` to build and run each time I did a `cargo clean` so perhaps it's finally time to change things. 😄

closes #123


Meziu:
I'm very happy to see you working on the project again! I'm sorry for not responding sooner, but I was busy with the documentation PR that is now finally over #134.

These new changes seem very promising, especially since they are pretty much required to officially support many `libctru` functions.

There are just 2 things I'd like to say before actually working towards merging this PR:
1. As previously mentioned by @Techie-Pi, not using a shell script to generate bindings would greatly improve general compatibility and integration with `doxygen-rs`, which is currently run as a standalone program. I'm very supportive of the idea to completely port over the shell script to the `build.rs`, even in this PR.
2. Is releasing the crate/documentation impacted by the changes? Especially the extern library blob. Should we upload the binary to `crates.io` or should we let the user build it themselves (automatically). How would the versioning system get impacted by the changes? Currently we assign a version manually so that the user can be sure they are using the correct one for their `libctru` version, but it might get harder to do that once the crate becomes more dynamic.

Tell me what you think about the second issue in particular.

FenrirWolf:
Both the bindings and the extern library can be created via `build.rs` so I can go ahead and make that change. I should be able to move the version-checking code there too so I don't think that will be a problem.

FenrirWolf:
So the CI pipeline is failing because `clang` can't find `stdbool.h` for some reason. Not sure what's up with that since everything works on my machine™. I'll have to dig around and see if I can fix that.

FenrirWolf:
Okay, so `stdbool.h` lives in a directory whose name depends on `gcc`'s version. Fun. So I just have to add some code to parse for that info and that should do the trick.

FenrirWolf:
Looks like everything is working now. Other than the build script being kind of a mess to read, we should be good to go.

Meziu (approved):
Looks great to me! I've had to do some cleanups before using it (most likely because the new build script completely broke all earlier incremental artifacts) but it works very nicely now.

Meziu:
@ian-h-chamberlain @FenrirWolf I was wondering about a way to solve the versioning system conundrum.

With the changes proposed by this PR (which I’m very supportive of) the current versioning system makes very little sense. It’s good if we are the ones providing the bindings when developers pull the `ctru-sys` package, but now that bindings are always generated locally, making `ctru-sys` have a restrictive version is more or less meaningless, since there is no guarantee the used version will even have the required bindings.

Instead, maybe the dependant package should be the one checking for the version, by pulling a generic `ctru-sys` version (which would change based on our work on the binding generator) and a different requirement if a specific `libctru` version is needed.

# Example

```toml
[dependencies]
ctru-sys = “0.1.0”

[package.metadata.ctru-sys]
required-libctru = "2.2.0"
``` 

This requirement would still not be restrictive, instead just printing the “wrong version” warning that is already in place.

Using the `metadata` field was my first idea, but tell me if you think other options might be better.

AzureMarker:
What's the benefit of running bindgen during the build instead of via script? Will docs.rs be able to show accurate info? (I don't think it will have libctru available when building docs)

FenrirWolf:
@AzureMarker Not much really changes from an end-user perspective. I felt like it was a good change to make though since we're not only generating bindings now, but we also need to build and compile the extra static fns wrapper and it seems better to handle both those steps in one place instead of having logic split between a shell script and a dedicated bindings generator crate.

Docs are still generated properly too, so there won't be any changes or issues when it comes to that. I would say the main benefit from the maintenance side is that everything is handled by cargo and we don't have to manually run any scripts or programs when updating `ctru-sys`. Speaking of which...

@Meziu I would say it's still worth enforcing versions for the `ctru-sys` crate in some way even if the build script itself can technically handle arbitrary `libctru` versions. It's always been best practice for 3ds homebrew devs to move to the latest versions of `devkitArm` and `libctru` as they get released, and it's also useful for any given version of `ctru-rs` to know what version of libctru it's actually being linked against. So I feel like it would be fine for `ctru-sys` to outright refuse to build if it detects that the wrong version of libctru is present. Or we could stick with a soft warning setup instead if we feel like it's useful to leave a bit of wiggle room still.

ctru-sys/build.rs
```diff
@@ -30,9 +45,97 @@ fn main() {
...
+        .arg("--version")
+        .stderr(Stdio::inherit())
+        .output()
+        .unwrap();
+
+    let stdout_str = String::from_utf8_lossy(&stdout);
+
+    stdout_str
+        .split(|c: char| c.is_whitespace())
+        .nth(4)
```
ian-h-chamberlain, line 133: Looks like gcc also has `-dumpversion`/`-dumpfullversion` flags to make this even simpler if you want

ctru-sys/build.rs
```diff
@@ -30,9 +45,97 @@ fn main() {
...
+    let gcc_version = get_gcc_version(PathBuf::from(&devkitarm).join("bin/arm-none-eabi-gcc"));
+
+    let include_path = PathBuf::from_iter([devkitpro.as_str(), "libctru", "include"]);
+    let ctru_header = include_path.join("3ds.h");
+
+    let sysroot = Path::new(&devkitarm).join("arm-none-eabi");
+    let system_include = sysroot.join("include");
+    let gcc_include = PathBuf::from(format!(
+        "{devkitarm}/lib/gcc/arm-none-eabi/{gcc_version}/include"
+    ));
```
ian-h-chamberlain, line 58: It would be nice if we could get `gcc` to list these flags for us, but it doesn't seem like there's a standard way to to do it so this seems fine.

ctru-sys/build.rs
```diff
@@ -30,9 +45,97 @@ fn main() {
...
+    let errno_header = system_include.join("errno.h");
+
+    // Build libctru bindings
+    let bindings = Builder::default()
+        .header(ctru_header.to_str().unwrap())
+        .header(errno_header.to_str().unwrap())
```
ian-h-chamberlain, line 64: Surprised that this header needs to be included explicitly, but if it's not working without then :shrug:

ctru-sys/build.rs
```diff
@@ -30,9 +45,97 @@ fn main() {
...
+    let ar = Path::new(devkitarm.as_str()).join("bin/arm-none-eabi-ar");
+
+    cc::Build::new()
+        .compiler(cc)
+        .archiver(ar)
+        .include(&include_path)
+        .file(out_dir.join("libctru_statics_wrapper.c"))
+        .flag("-march=armv6k")
+        .flag("-mtune=mpcore")
+        .flag("-mfloat-abi=hard")
```
ian-h-chamberlain, line 115: It would be nice to have a common (function or array or something) definition for these flags, at least where they overlap between `clang_args` and `cc::Build::flags`

ctru-sys/src/lib.rs
```diff
@@ -5,12 +5,10 @@
 #![allow(clippy::all)]
 
 pub mod result;
-
-mod bindings;
-
-pub use bindings::*;
 pub use result::*;
 
+include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
```
ian-h-chamberlain, line 10: One reason I preferred a module previously was for `rust-analyzer`'s "go to definition", autocomplete, and doc comments to work a lot nicer (not sure if other IDEs / language servers have the same issue). It looks like this should be working now with https://github.com/rust-lang/rust-analyzer/issues/3767 closed, but when I check out your branch I can't seem to see any documentation, nor jump to these definitions.

Does your IDE work with the `include!()` like this? I wonder if I just have a misconfiguration — in any case, this probably isn't a big enough deal to block this change — but the developer experience is a lot nicer when IDE features work with it IMO.

ian-h-chamberlain:
> With the changes proposed by this PR (which I’m very supportive of) the current versioning system makes very little sense. It’s good if we are the ones providing the bindings when developers pull the `ctru-sys` package, but now that bindings are always generated locally, making `ctru-sys` have a restrictive version is more or less meaningless, since there is no guarantee the used version will even have the required bindings.

Yeah, I think an approach like this makes sense, but we probably could just hardcode a minimum version right in the `ctru-sys` build script and check it at build time. So something like this:
```
ctru-rs v0.7.1
└── ctru-sys v0.7.1 (probably just 1-to-1 version match for this dependency?)
    └── libctru v2.2.2-1 (could be enforced >= by build script)
```

If we want to get fancy with it we could even do an actual semver check or something, although I'm not sure upstream really follows that convention -- so a minimum version seems like it would work okay IMO.

----

> Docs are still generated properly too, so there won't be any changes or issues when it comes to that. I would say the main benefit from the maintenance side is that everything is handled by cargo and we don't have to manually run any scripts or programs when updating `ctru-sys`. Speaking of which...

That's true for our CI environment, but probably not for `docs.rs` as @AzureMarker alluded to... Come to think of it, I don't know if this would have ever worked. There are some notes on https://docs.rs/about/builds#missing-dependencies about how the `docs.rs` build environment works and adding dependencies. I'm not sure if a full build (i.e. linking) is required, or if the checked-in bindings would have worked before. Either way I think this is a separate problem we'll have to resolve so I wouldn't say this PR should be blocked by it. In the meantime we could always host docs on `rust3ds.github.io` using Github Pages.

ian-h-chamberlain (approved):

ctru-sys/build.rs
Meziu: As the [comment](https://github.com/rust3ds/ctru-rs/pull/110#issuecomment-1507139787) you made to support this change, the reason why we need to explicitly include it is because it's not included neither by the C standard library nor by `libctru`.

Meziu:
> Yeah, I think an approach like this makes sense, but we probably could just hardcode a minimum version right in the `ctru-sys` build script and check it at build time. So something like this:
> 
> ```
> ctru-rs v0.7.1
> └── ctru-sys v0.7.1 (probably just 1-to-1 version match for this dependency?)
>     └── libctru v2.2.2-1 (could be enforced >= by build script)
> ```

At that point `ctru-sys` could just hold a generic stable version number (like `0.1.0`) and simply make the checks at compile time (build script) based on the package that depends on it. `ctru-rs` won't be the only crate using `ctru-sys` forever, so I wouldn't like to bind them together.

> If we want to get fancy with it we could even do an actual semver check or something, although I'm not sure upstream really follows that convention -- so a minimum version seems like it would work okay IMO.

It's fine, `>=` is probably the best way for any package. Usually there aren't any breaking changes between `libctru` versions.

> That's true for our CI environment, but probably not for `docs.rs` as @AzureMarker alluded to...

Uff, I'd never thought about it. It's true that the `armv6k-nintendo-3ds` target *itself* requires `devkitPRO`, so this probably affects `ctru-rs` as well...

Regardless, this PR can go on without all these discussions.

ctru-sys/src/lib.rs
FenrirWolf: I just checked and I'm not getting code completion for `ctru-sys` on the branch either, and the reason for that is because rust-analyzer is failing to build and run ctru-sys with these changes: https://gist.github.com/FenrirWolf/2944fb0378d2724f5ce3d6785a59276f

Looks like the culprit is that `TARGET` is set `x86_64-unknown-linux-gnu` instead of `armv6k-nintendo-3ds`, and so r-a passes `-m64` as a parameter and devkitARM doesn't like that.

ctru-sys/build.rs
ian-h-chamberlain: Lol — forgot about that, thanks for the reminder.

ctru-sys/src/lib.rs
ian-h-chamberlain: Ah, nice find! That gave me the idea to set `"rust-analyzer.cargo.target": "armv6k-nintendo-3ds"` in my config, and now it looks like everything is working the same as how it was when we were using a module!

I suppose this would be a good hint to add to our development wiki for people getting setup with `rust-analyzer` for general 3DS development.

ian-h-chamberlain:
> At that point `ctru-sys` could just hold a generic stable version number (like `0.1.0`) and simply make the checks at compile time (build script) based on the package that depends on it. `ctru-rs` won't be the only crate using `ctru-sys` forever, so I wouldn't like to bind them together.

Ok, let's move discussion about versioning back to #100 and this PR can move along (for now whatever version it already has should be fine since nothing's published yet).

ctru-sys/src/lib.rs
Meziu: > Ah, nice find! That gave me the idea to set `"rust-analyzer.cargo.target": "armv6k-nintendo-3ds"` in my config, and now it looks like everything is working the same as how it was when we were using a module!
> 
> I suppose this would be a good hint to add to our development wiki for people getting setup with `rust-analyzer` for general 3DS development.

True, this is wiki material. Still, I prefer overriding the check command with `cargo-3ds` so that I can use incremental builds (damn you, env variables!).

```json
"rust-analyzer.check.overrideCommand": [
        "cargo",
        "3ds",
        "clippy",
        "--message-format=json"
    ],
```

ian-h-chamberlain: Yeah, I actually already had something like that configured, but it seems like for the build script to work correctly you _also_ need the `cargo.target` configured correctly. I'll see if I can add a section to the `Guides` page about this.

Edit: created https://github.com/rust3ds/ctru-rs/wiki/Guides#setting-up-rust-analyzer

Meziu:
Will merge this one since it holds some pretty valuable changes for the repository, and it hasn't received any blocking reviews by anyone. The discussion about "generating vs including" the bindings can be made elsewhere.

PR #134, opened 2023-07-28
Improve documentation and examples

### LINKED ISSUE #32: General documentation improvements

Meziu:
Closes #32 

Finally, the long awaited "documentation PR" I have so much anticipated has come.

This PR's changes boil down to:
- Newly written READMEs for each package (with the idea in mind that they will eventually be read from [crates.io](https://crates.io/)).
- Changed `Cargo.toml` manifests to include all metadata required for a future release on [crates.io](https://crates.io/).
- Improved examples with the addition of a better documentation that explains their behavior (except for the "threads" and "futures" examples that still need to be settled).
- A pretty substantial list of new TODOs that I discovered while thoroughly testing the crate (a lot of testing has been done...)
- `#[doc(alias)]` for ALL functions and structs that benefit a 1:1 relation with `ctru-sys`, so that users used to the C library can search for names they already know and find out how that functionality translates in `ctru-rs`.
- Slight bump to the minimum rust versions (the same as `cargo-3ds`).

And most importantly: 
- COMPLETE member-by-member documentation that includes multiple documentation layers, doclinks, and small doctests/examples for most functions.

I suggest you review this PR mostly by looking at the result of the `cargo 3ds doc` compilation, since the git diff is huge.

This PR took me weeks of work and it marks one big milestone for this project (in some cases, the new docs are actually better than `libctru`'s official ones :P ).

Thanks for your attention. I may finally go puke in a nearby bucket. Writing docs is nauseous 🤮 .


FenrirWolf:
When I check out the branch and run  `cargo 3ds doc --open`, I get the following error from `shim_3ds` that prevents the docs from fully building:

```
error[E0432]: unresolved import `ctru_sys::__errno`
 --> /home/fenrir/.cargo/git/checkouts/shim-3ds-0fd8aff15f8f05f2/2d3ebe6/src/lib.rs:3:5
  |
3 | use ctru_sys::__errno;
  |     ^^^^^^^^^^-------
  |     |         |
  |     |         help: a similar name exists in the module: `errno`
  |     no `__errno` in the root

For more information about this error, try `rustc --explain E0432`.
error: could not compile `shim-3ds` (lib) due to previous error
```

Meziu:
> When I check out the branch and run `cargo 3ds doc --open`, I get the following error from `shim_3ds` that prevents the docs from fully building:
> 
> ```
> error[E0432]: unresolved import `ctru_sys::__errno`
>  --> /home/fenrir/.cargo/git/checkouts/shim-3ds-0fd8aff15f8f05f2/2d3ebe6/src/lib.rs:3:5
>   |
> 3 | use ctru_sys::__errno;
>   |     ^^^^^^^^^^-------
>   |     |         |
>   |     |         help: a similar name exists in the module: `errno`
>   |     no `__errno` in the root
> 
> For more information about this error, try `rustc --explain E0432`.
> error: could not compile `shim-3ds` (lib) due to previous error
> ```

You probably need to update your git cache. It’s been updated a while ago. `cargo update` should do the trick.

FenrirWolf:
That's the thing though, I can build and read the docs just fine on the latest `master` but not on this branch. Same thing after running `cargo update` too.

Meziu:
> That's the thing though, I can build and read the docs just fine on the latest `master` but not on this branch. Same thing after running `cargo update` too.

I don’t really know what to tell you except `cargo clean` and `cargo update`. The function is very much there and even the CI (which uses a fresh environment) can build the documentation just fine.

FenrirWolf:
Got them building after nuking my local project and then checking the branch out again. Not sure what that was about but at least the problem's gone now 😆

ctru-rs/src/services/cam.rs
```diff
@@ -783,7 +974,30 @@ impl Cam {
         }
     }
 
-    /// Plays the specified sound based on the [ShutterSound] argument
+    /// Play the specified sound based on the [`ShutterSound`] argument
+    ///
+    /// # Notes
+    ///
+    /// Playing the shutter sound does not require a liviving handle to the [`Ndsp`](crate::services::ndsp::Ndsp) service.
```
FenrirWolf:
```suggestion
    /// Playing the shutter sound does not require a living handle to the [`Ndsp`](crate::services::ndsp::Ndsp) service.
```

FenrirWolf:
Other than typos and some suggestions from Github Actions that might be worth implementing, I'd say that this looks good to me

FenrirWolf (approved):
Looks like the things I saw from Github Actions weren't actually caused by this PR so they don't need to block it. I think this looks good to merge

adryzz (approved):
great job! LGTM

Techie-Pi (approved):

ctru-rs/Cargo.toml
```diff
@@ -1,11 +1,15 @@
 [package]
-authors = ["Rust3DS Org", "Ronald Kinard <furyhunter600@gmail.com>"]
-description = "A safe wrapper around smealum's ctrulib."
-license = "Zlib"
 name = "ctru-rs"
 version = "0.7.1"
+authors = ["Rust3DS Org", "Ronald Kinard <furyhunter600@gmail.com>"]
+description = "A safe wrapper around libctru"
+repository = "https://github.com/rust3ds/ctru-rs"
+keywords = ["3ds", "libctru"]
```
ian-h-chamberlain, line 7: Do you think it's worth including a mention of `devkitPro` or `devkitARM` somewhere in the metadata? Not sure if people would be likely to search for something like that on https://lib.rs or https://crates.io

ctru-rs/Cargo.toml
```diff
@@ -1,11 +1,15 @@
...
+categories = ["os", "api-bindings"]
```
ian-h-chamberlain: Possible additional categories we could have (I forget how many are allowed):
- `game-development`
- `hardware-support`

Not sure if they would also apply to `ctru-sys` or if they even make sense but just thinking about visibility

ctru-rs/Cargo.toml
```diff
@@ -1,11 +1,15 @@
...
+exclude = ["examples"]
```
ian-h-chamberlain, line 9: Just curious, why exclude examples? Just for crate size or something?

ctru-rs/Cargo.toml
```diff
@@ -45,6 +49,10 @@ std-threads = []
 [package.metadata.cargo-3ds]
 romfs_dir = "examples/romfs"
 
+[package.metadata.docs.rs]
+default-target = "armv6k-nintendo-3ds"
+cargo-args = ["-Z", "build-std"]
```
ian-h-chamberlain, line 55: As mentioned in https://github.com/rust3ds/ctru-rs/pull/133#issuecomment-1666027623 this may not be enough if the crate has to link with libctru etc., but we can cross that bridge when we get there.

ctru-sys/Cargo.toml
```diff
@@ -11,3 +16,7 @@ libc = { version = "0.2.121", default-features = false }
 
 [build-dependencies]
 which = "4.4.0"
+
+[package.metadata.docs.rs]
+default-target = "armv6k-nintendo-3ds"
```
ian-h-chamberlain, line 21: Should we also have `targets = []` to skip building other targets? From https://docs.rs/about/metadata:
> ```toml
> # Set this to `[]` to only build the default target.
> ```

ctru-rs/examples/file-explorer.rs
```diff
@@ -137,7 +143,7 @@ impl<'a> FileExplorer<'a> {
         }
     }
 
-    /// Paginate output
+    /// Paginate output.'
```
ian-h-chamberlain: IIRC doc comments on non-`pub` functions are the same as regular comments, and I think it actually might emit a clippy warning in some cases to have three slashes like this

```suggestion
    // Paginate output.
```

ctru-rs/examples/gfx-3d-mode.rs
```diff
@@ -1,11 +1,15 @@
+//! 3D Graphics example.
+//!
+//! This example showcases 3D mode rendering (using the CPU).
+//! In a normal application, all rendering should be hanlded via the GPU.
```
ian-h-chamberlain:
```suggestion
//! In a normal application, all rendering should be handled via the GPU.
```

ctru-rs/examples/gfx-3d-mode.rs
```diff
@@ -1,11 +1,15 @@
...
+
 use ctru::prelude::*;
 use ctru::services::gfx::{Flush, Screen, Side, Swap, TopScreen3D};
 
-/// See `graphics-bitmap.rs` for details on how the image is generated.
-///
-/// WARNING: this example uses 3D mode in a rather unnatural way, and should
-/// probably not be viewed for too long or at all if you are photosensitive.
-
+// See `graphics-bitmap.rs` for details on how the image is generated.
```
ian-h-chamberlain: Maybe this comment should be moved up to the top-level doc comment, actually...

Edit: also should be updated to point to the new name `gfx-bitmap`.

ctru-rs/examples/gfx-3d-mode.rs
```diff
@@ -17,47 +21,50 @@ fn main() {
     let apt = Apt::new().expect("Couldn't obtain APT controller");
     let _console = Console::new(gfx.bottom_screen.borrow_mut());
 
-    println!("Press Start to exit.\nPress A to switch sides (be sure to have 3D mode enabled).");
+    println!("Press A to switch sides.");
+    println!("Make sure to have set the 3D slider correctly");
+    println!("\x1b[29;12HPress Start to exit");
```
ian-h-chamberlain, line 29: 👍

ctru-rs/src/applets/mii_selector.rs
```diff
@@ -90,12 +115,47 @@ impl MiiSelector {
...
+        unsafe { ctru_sys::miiSelectorSetOptions(self.config.as_mut(), options.bits()) }
     }
 
-    /// Whitelist a guest Mii
+    /// Whitelist a guest Mii based on its index.
+    ///
+    /// # Notes
+    ///
+    /// Guest Mii's won't be available regardless of their whitelist/blacklist state if the [`MiiSelector`] is run without setting [`Options::ENABLE_GUESTS`].
+    /// Look into [`MiiSelector::set_options()`] to see how to work with options.
```
ian-h-chamberlain, line 144: We don't have to (and if we do, we should definitely still have doc aliases) but lots of software projects have moved to using `allowlist`/`blocklist` wording for APIs like these. Maybe something to consider before publishing but doesn't have to be in this PR.

ctru-rs/src/applets/mii_selector.rs
```diff
@@ -135,28 +242,54 @@ impl MiiSelector {
...
         self.config.initial_index = index as u32;
     }
 
     /// Launch the Mii Selector.
-    /// Returns an error when the checksum of the Mii is invalid.
-    pub fn launch(&mut self) -> Result<SelectionResult, LaunchError> {
+    ///
+    /// Depending on the configuration, the Mii Selector window will appear either on the bottom screen (default behaviour) or the top screen (see [`Options::USE_TOP_SCREEN`]).
+    ///
+    /// TODO: UNSAFE OPERATION, LAUNCHING APPLETS REQUIRES GRAPHICS, WITHOUT AN ACTIVE GFX THIS WILL CAUSE A SEGMENTATION FAULT.
```
ian-h-chamberlain, line 259: You could just put this in a `# Safety` section for now, even though the function isn't marked as `unsafe` I think that's more conventional vs a note like this.

ctru-rs/src/applets/swkbd.rs
```diff
@@ -1,106 +1,208 @@
...
     }
 
-    /// Gets input from this keyboard and appends it to the provided string.
+    /// Launches the applet based on the given configuration and returns a string containing the text input.
+    ///
+    /// # Notes
     ///
     /// The text received from the keyboard will be truncated if it is longer than `max_bytes`.
+    ///
+    /// TODO: UNSAFE OPERATION, LAUNCHING APPLETS REQUIRES GRAPHICS, WITHOUT AN ACTIVE GFX THIS WILL CAUSE A SEGMENTATION FAULT.
```
ian-h-chamberlain, line 190: Same comment here re: `# Safety` heading for the doc comment.

ctru-rs/src/applets/swkbd.rs
```diff
@@ -181,19 +390,34 @@ impl Swkbd {
         }
     }
 
-    /// Configures the maximum number of UTF-16 code units that can be entered into the software
-    /// keyboard. By default the limit is 0xFDE8 code units.
+    /// Configure the maximum number of UTF-16 code units that can be entered into the software
+    /// keyboard. By default the limit is `0xFDE8` code units.
```
ian-h-chamberlain: Btw `0xFDE8` == 65000 which although still arbitrary might be more readable for documentation purposes

ctru-rs/src/services/ndsp/mod.rs
```diff
@@ -1,7 +1,11 @@
-//! NDSP (Audio) service
+//! NDSP (Audio) service.
```
ian-h-chamberlain, line 1: Is there such thing as `#![doc(alias)]` for entire modules? It might be useful to alias this whole module as `"audio"` or something like that, since the name is not obvious at first glance.

Maybe having the word `Audio` in the title is enough to find it, but if not, an alias would be helpful for this and a bunch of other modules that have weird acronyms or unusual names.

Edit: looking through all the `services`, we could probably stand to do this for nearly _all_ of them, and a quick test confirms you can alias a module with `#![doc(alias = "...")]`. Offhand I can think of some simple ones:
- `ndsp` -> `audio`
- `am` -> `application` (maybe `application manager`? not sure how spaces are treated)
- `apt` -> `applet`, although possible confusing with the `applets` module...
- `cam` -> `camera`
- `cfgu` -> `localization`, `region`, idk
- `fs` -> `filesystem`
- `gfx` -> `graphics`
- `gspgpu` -> `gpu`? This one's less generally useful I guess so maybe not
- `hid` -> `input`, `controller`, `gamepad`, `touch[screen]`?
- `ps` -> `random`, `crypto`? maybe not so important
- `romfs` -> `embed`, maybe also `filesystem`?
- `soc` -> `socket`, `network`
- `sslc` -> `crypto`? maybe ignore for now

Feel free to take/discard any of these as you see fit but I definitely think adding some aliases like these would make the docs a lot more searchable / usable (I know I've struggled many times to find things in `libctru` docs just because of the cryptic / meaningless names).

ctru-rs/src/services/cam.rs
```diff
@@ -783,7 +974,30 @@ impl Cam {
         }
     }
 
-    /// Plays the specified sound based on the [ShutterSound] argument
+    /// Play the specified sound based on the [`ShutterSound`] argument
+    ///
+    /// # Notes
+    ///
```
ian-h-chamberlain, line 980:
```suggestion
    /// Volume will always be maxed out to ensure everyone within photo range can hear the picture being taken (as by Japanese law).
```

ctru-rs/src/services/romfs.rs
```diff
@@ -1,30 +1,57 @@
-//! Read-Only Memory FileSystem
+//! Read-Only Memory FileSystem service.
+//!
+//! This service lets the application access a virtual mounted device created using a folder included within the application bundle.
+//! After mounting the RomFS file system, the included files and folders will be accessible exactly like any other file, just by using the drive prefix `romfs:/<file-path>`.
```
ian-h-chamberlain, line 4: I wonder if it's worth mentioning the weirdness with `std::path` (i.e. https://github.com/rust-lang/rust/issues/52331) somewhere in here, or maybe in the `fs` module. I guess we at least have some examples and _most_ of the time people shouldn't have to worry about it (just pass their paths in plainly).

ctru-rs/src/services/romfs.rs
```diff
@@ -36,14 +63,19 @@ impl RomFS {
...
             },
         )?;
 
         Ok(Self { _service_handler })
     }
 }
 
+impl Drop for RomFS {
+    #[doc(alias = "romfsUnmount")]
+    fn drop(&mut self) {}
```
ian-h-chamberlain: By the way, implementing Drop like this actually does change the semantics of the `RomFS` struct. I don't think it should matter much in this case (mostly has implications [around lifetimes](https://doc.rust-lang.org/std/ops/trait.Drop.html#drop-check), I think), but just something to be aware of.

Edit: I guess since it contains a `ServiceReference` there's probably already some drop glue in place, so maybe it actually doesn't matter in this case.

ctru-rs/src/services/hid.rs
```diff
@@ -1,75 +1,154 @@
...
         const CPAD_LEFT     = ctru_sys::KEY_CPAD_LEFT;
+        /// CirclePad Up.
         const CPAD_UP       = ctru_sys::KEY_CPAD_UP;
+        /// CirclePad Down.
         const CPAD_DOWN     = ctru_sys::KEY_CPAD_DOWN;
-        // Convenience catch-all for the dpad and cpad
+
+        // Convenience catch-all for the D-Pad and the C-Pad
+
+        /// Direction Up (either D-Pad or C-Pad).
```
ian-h-chamberlain: Tiny nit: this is called `CirclePad` in the comments above, but `C-Pad` here.

ctru-rs/src/services/hid.rs
```diff
@@ -1,75 +1,154 @@
...
         const UP    = KeyPad::DPAD_UP.bits()    | KeyPad::CPAD_UP.bits();
+        /// Direction Down (either D-Pad or C-Pad).
         const DOWN  = KeyPad::DPAD_DOWN.bits()  | KeyPad::CPAD_DOWN.bits();
+        /// Direction Left (either D-Pad or C-Pad).
         const LEFT  = KeyPad::DPAD_LEFT.bits()  | KeyPad::CPAD_LEFT.bits();
+        /// Direction Right (either D-Pad or C-Pad)..
```
ian-h-chamberlain:
```suggestion
        /// Direction Right (either D-Pad or C-Pad).
```

ctru-rs/src/services/gfx.rs
```diff
@@ -65,40 +65,53 @@ pub trait Screen: private::Sealed {
...
+
+    /// Set whether to use double buffering.
+    ///
+    /// # Notes
+    ///
+    /// Double buffering is enabled by default.
+    /// [`Swap::swap_buffers`] must be called after this function for the configuration
+    /// change to take effect.
+    #[doc(alias = "gfxSetDoubleBuffering")]
+    fn set_double_buffering(&mut self, enabled: bool);
```
ian-h-chamberlain, line 115: Yeah, this makes more sense as a part of `Swap` 👍

ctru-rs/src/services/gfx.rs
```diff
@@ -138,7 +166,7 @@ impl<S: Screen> Flush for S {
         let framebuffer = self.raw_framebuffer();
 
         // Flush the data array. `self.raw_framebuffer` should get the correct parameters for all kinds of screens
-        unsafe {
+        let _ = unsafe {
```
ian-h-chamberlain, line 170: huh. Should this function be returning a `Result`? I wonder what scenarios flushing could fail like this...

ctru-rs/src/services/cfgu.rs
```diff
@@ -1,104 +1,224 @@
...
     USA = ctru_sys::CFG_REGION_USA,
+    /// Europe.
     Europe = ctru_sys::CFG_REGION_EUR,
+    /// Australia.
     Australia = ctru_sys::CFG_REGION_AUS,
+    /// China.
     China = ctru_sys::CFG_REGION_CHN,
+    /// Korea.
     Korea = ctru_sys::CFG_REGION_KOR,
+    /// Taiwan.
```
ian-h-chamberlain, line 25: Do these enum variants really need doc comments? It's nice to have everything documented but in this case it just seems redundant.

ctru-rs/src/console.rs
```diff
@@ -7,19 +14,61 @@ use crate::services::gfx::Screen;
 
 static mut EMPTY_CONSOLE: PrintConsole = unsafe { const_zero::const_zero!(PrintConsole) };
 
+/// Virtual text console.
+///
+/// [`Console`] lets the application redirect `stdout` to a simple text displayer on the 3DS screen.
+/// This means that any text written to `stdout` (e.g. using `println!` or `dbg!`) will become visible in the area taken by the console.
```
ian-h-chamberlain: Stderr is redirected by default too, right? Maybe worth mentioning

ctru-rs/src/console.rs
```diff
@@ -7,19 +14,61 @@ use crate::services::gfx::Screen;
...
+///
+/// # Notes
+///
+/// The [`Console`] will take full possession of the screen handed to it as long as it stays alive. It also supports ANSI codes.
```
ian-h-chamberlain:
```suggestion
/// The [`Console`] will take full possession of the screen handed to it as long as it stays alive. It also supports some ANSI codes, such as text color and cursor positioning.
```

ctru-rs/src/console.rs
```diff
@@ -7,19 +14,61 @@ use crate::services::gfx::Screen;
...
+/// The [`Console`]'s window will have a size of 40x30 on the bottom screen, 50x30 on the normal top screen and
+/// 100x30 on the top screen when wide mode is enabled.
+///
+/// # Alternatives
+///
+/// If you'd like to see live `stdout` output while running the application but cannnot/do not want to show the text on the 3DS itself,
```
ian-h-chamberlain:
```suggestion
/// If you'd like to see live `stdout` output while running the application but cannot or do not want to show the text on the 3DS itself,
```

ctru-rs/src/console.rs
```diff
@@ -7,19 +14,61 @@ use crate::services::gfx::Screen;
...
+/// you can try using [`Soc::redirect_to_3dslink`](crate::services::soc::Soc::redirect_to_3dslink) while activating the `--server` flag for `3dslink` (also supported by `cargo-3ds`).
+/// More info in the `cargo-3ds` docs.
```
ian-h-chamberlain:
```suggestion
/// More info in the [`cargo-3ds` docs](https://github.com/rust3ds/cargo-3ds#running-executables).
```

ctru-rs/src/console.rs
```diff
@@ -7,19 +14,61 @@ use crate::services::gfx::Screen;
...
 impl<'screen> Console<'screen> {
-    /// Initialize a console on the chosen screen, overwriting whatever was on the screen
-    /// previously (including other consoles). The new console is automatically selected for
-    /// printing.
+    /// Initialize a console on the chosen screen.
     ///
     /// # Notes
     ///
-    /// [Console] automatically takes care of flushing and swapping buffers for its screen when printing.
+    /// This operation overwrites whatever was on the screen before the inizialization (including other [`Console`]s)
```
ian-h-chamberlain:
```suggestion
    /// This operation overwrites whatever was on the screen before the initialization (including other [`Console`]s)
```

ctru-rs/src/console.rs
```diff
@@ -44,26 +121,65 @@ impl<'screen> Console<'screen> {
...
+    /// #
+    /// use ctru::console::Console;
+    ///
+    /// // Create a `Console` that takes control of the upper LCD screen.
+    /// let top_console = Console::new(gfx.top_screen.borrow_mut());
+    ///
+    /// // Create a `Console` that takes control of the lower LCD screen.
+    /// let bottom_console = Console::new(gfx.bottom_screen.borrow_mut());
+    ///
+    /// // Remember that `Console::new` automatically selects the new `Console` for ouput.
```
ian-h-chamberlain:
```suggestion
    /// // Remember that `Console::new` automatically selects the new `Console` for output.
```

ctru-rs/src/console.rs
```diff
@@ -44,26 +121,65 @@ impl<'screen> Console<'screen> {
         }
     }
 
-    /// Select this console as the current target for stdout
+    /// Select this console as the current target for `stdout`.
```
ian-h-chamberlain: Same question as before - is stderr also selected with this function?

ctru-rs/src/error.rs
```diff
@@ -6,8 +10,29 @@ use std::ops::{ControlFlow, FromResidual, Try};
 
 use ctru_sys::result::{R_DESCRIPTION, R_LEVEL, R_MODULE, R_SUMMARY};
 
+/// Custom type alias for generic [`ctru-rs`](crate) operations.
+///
+/// This type is compatible with `ctru-sys::Result` codes.
```
ian-h-chamberlain: Can we do intra-doc linking with `` [`ctru_sys::Result`] `` for references like this?

ctru-rs/src/error.rs
```diff
@@ -48,13 +73,20 @@ impl<T> FromResidual<Error> for Result<T> {
...
 
-/// The error type returned by all libctru functions.
+/// The generic error enum returned by [`ctru-rs`](crate) functions.
+///
+/// This error enum supports parsing and displaying [`ctru_sys::Result`] codes.
 #[non_exhaustive]
 pub enum Error {
+    /// Raw [`ctru_sys::Result`] codes.
     Os(ctru_sys::Result),
+    /// Generic `libc` error codes.
```
ian-h-chamberlain: Same question here, could this be `` [`libc`] `` ? Also since it's a string probably would just say `errors` instead of `error codes`.

ctru-rs/src/lib.rs
```diff
@@ -1,16 +1,44 @@
...
+//!
+//! # Usage
+//!
+//! Thoroughly read the official [`ctru-rs` wiki](https://github.com/rust3ds/ctru-rs/wiki) which guides you through the setup needed to install the required toolchain and helpful tools.
+//! After following the guide and understanding the many quirks of the Nintendo 3DS homebrew development environment, you can create a new project by including this crate as a dependency
+//! in your `Cargo.toml` manifest and build your binaries either manually (for the `armv6k-nintendo-3ds` target) or via [`cargo-3ds`](https://github.com/rust3ds/cargo-3ds).
+
 #![crate_type = "rlib"]
 #![crate_name = "ctru"]
+#![warn(missing_docs)]
```
ian-h-chamberlain, line 20: Very nice!

ctru-rs/src/lib.rs
```diff
@@ -1,16 +1,44 @@
...
 
 // Nothing is imported from these crates but their inclusion here assures correct linking of the missing implementations.
 extern crate pthread_3ds;
 extern crate shim_3ds;
 
+/// Expanded stack size used to spawn the main thread by `libctru`.
+///
+/// It takes effect only if the `big-stack` feature is active. Otherwise, the default stack size should be ~32kB.
+///
+/// This value was chosen to support crate dependencies which expected more stack than provided. It's suggested to use less stack if possible.
```
ian-h-chamberlain, line 41: This isn't `pub` so I think this can just be a normal `//` comment. Still useful to have though.

ctru-rs/src/lib.rs
```diff
@@ -25,19 +53,24 @@ macro_rules! from_impl {
...
 pub fn use_panic_handler() {
     #[cfg(not(test))]
     panic_hook_setup();
 }
 
+/// Internal protocol to activate the custom panic handler hook.
+///
+/// # Notes
+///
+/// When `test` is enabled, this function will be ignored.
```
ian-h-chamberlain, line 73: same here, don't think this needs to be a doc comment

ctru-rs/src/mii.rs
```diff
@@ -1,213 +1,301 @@
...
 
-/// Represents all the data of a Mii
+/// Full Mii data representation.
 ///
-/// Some values are not ordered _like_ the Mii Editor UI. The mapped values can be seen here:
-/// <https://www.3dbrew.org/wiki/Mii#Mapped_Editor_.3C-.3E_Hex_values>
+/// Some values are not ordered *like* the Mii Editor UI. The mapped values can be seen [here](https://www.3dbrew.org/wiki/Mii#Mapped_Editor_.3C-.3E_Hex_values).
 ///
-/// This struct is returned by the [``MiiSelector``](crate::applets::mii_selector::MiiSelector)
+/// This struct can be retrieved by [`MiiSelector::lauch()`](crate::applets::mii_selector::MiiSelector::launch).
```
ian-h-chamberlain:
```suggestion
/// This struct can be retrieved by [`MiiSelector::launch()`](crate::applets::mii_selector::MiiSelector::launch).
```

ian-h-chamberlain (approved):
Wow, awesome work!!! This is really a huge step forward for the library and looks like it took a lot of effort, so thank you!

Many of my comments are just little typos or minor wording change suggestions, so feel free to ignore them as you want. I probably missed some stuff but I tried to look at everything I could, so hopefully some of it is helpful.

This is really exciting!

ian-h-chamberlain (approved):
Wow, awesome work!!! This is really a huge step forward for the library and looks like it took a lot of effort, so thank you!

Many of my comments are just little typos or minor wording change suggestions, so feel free to ignore them as you want. I probably missed some stuff but I tried to look at everything I could, so hopefully some of it is helpful.

This is really exciting!

ctru-rs/Cargo.toml
Meziu: I don't feel like that's something pertinent to the `keywords` field. `3ds` was obvious for anyone searching crates related to the platform, and `libctru` is the name of the original library. `devkitPro` seems a bit too generic/mostly superfluous (basically any crate that uses `3ds` would also use `devkitPRO`) and `devkitARM` isn't any better.

ctru-rs/Cargo.toml
Meziu: `hardware-support` may be worth adding, though `game-development` is a bit too specific for what's a general purpose crate.

ctru-rs/Cargo.toml
Meziu: Yes, people suggest to do so since the crates pulled from `crates.io` end up in the `cargo` cache and aren't easily accessible. It's just meant to keep the archive size small.

ctru-rs/Cargo.toml
Meziu: Correct, I hadn't thought that out when I made this change.

ctru-rs/examples/gfx-3d-mode.rs
Meziu: ❓

ctru-rs/src/applets/mii_selector.rs
Meziu: One more chore on the massive todo list!

Thank you for noting this out.

ctru-rs/examples/gfx-3d-mode.rs
ian-h-chamberlain: Just thought this looks more readable with separate lines. Nice change

ctru-rs/src/applets/mii_selector.rs
Meziu: This is meant to be a very temporary message meant to be fixed as soon as this PR merges. It's not meant to be actually seen by users.

ctru-rs/src/services/gfx.rs
Meziu: Indeed. I've made this change as part of the "documentation" PR since it was a bit dumb to link it in a different struct (also, I'm pretty sure it didn't work correctly with the 3D left-right sides before, but it should fine now).

ctru-rs/src/services/gfx.rs
Meziu: I guess it could fail if the pointers are invalid, but given the guarantees made by the lifetime system, it shouldn't be a problem.

ctru-rs/src/services/cfgu.rs
Meziu: I activated the lint for all docs since in most cases the enum variants have pretty useful documentation. This is just a small side effect that for me is worth having.

Tell me if you think there are better ways to handle these cases.

ctru-rs/src/error.rs
Meziu: They do locally but I was unsure whether it would work in the `docs.rs` environment (though that may be a problem for another time...).

ctru-rs/src/lib.rs
Meziu: When I said "complete" i meant complete! 😄

ctru-rs/src/lib.rs
Meziu: I know, it's just that I had actually forgot how it worked originally and I wanted to annotate it in a somewhat official way.

ctru-rs/src/services/cam.rs
```diff
@@ -783,7 +974,30 @@ impl Cam {
...
+    /// Playing the shutter sound does not require a living handle to the [`Ndsp`](crate::services::ndsp::Ndsp) service.
+    /// Volume will always be maxed out to ensure everyone within photo range can hear the picture being taken (as by japanese law).
```
ian-h-chamberlain: Oops, this one ended up on the wrong line somehow, just a quick typo fix:
```suggestion
    /// Volume will always be maxed out to ensure everyone within photo range can hear the picture being taken (as by Japanese law).
```

ctru-rs/src/services/cfgu.rs
ian-h-chamberlain: I'd probably just add `#[allow(missing_docs)]` on enums like this, but up to you if you think that's better or not.

ctru-rs/src/services/gfx.rs
ian-h-chamberlain: Ah yeah that makes sense, The error here looks like it would normally only fail if the address is not in linear memory (which for these frame buffers should never happen as far as I know): https://www.3dbrew.org/wiki/GSPGPU:FlushDataCache

ctru-rs/src/lib.rs
Meziu: Doc comments work differently when building internal documentation (the docs that aren't publicly visible). As such, even though private members without docs don't trigger the `#[warning(missing_docs)]`, there is a way to build and read them (`cargo 3ds doc --document-private-items`).

Normal comments wouldn't be able to do that.

ian-h-chamberlain (approved):

PR #135, opened 2023-08-10
Use external actions / test runner crate so we can actually run our tests!

ian-h-chamberlain:
There are three components here:
1. Replace the `setup` action with an external action (which basically does exactly the same thing)
2. Replace the `test_runner` module with an external crate which supports GDB output for running tests
3. Add an external action to run tests using Citra + GDB. This relies on the test runner implementation mentioned above.

I'm leaving this as a draft for now, since we probably want to move the external repo under the `rust3ds` organization first, but I wanted to show what I've been working on and see if the CI can actually run tests as expected!


ian-h-chamberlain:
It worked! See https://github.com/rust3ds/ctru-rs/actions/runs/5821500494/job/15784060428#step:9:16 for the test output

Meziu:
I suggest you transfer the runner etc. to the org and continue development from there. This PR is already taking shape!

ctru-rs/examples/thread-locals.rs
```diff
@@ -1,10 +1,10 @@
 #![feature(horizon_thread_ext)]
 
-use ctru::prelude::*;
-
 use std::cell::RefCell;
 use std::os::horizon::thread::BuilderExt;
 
+use ctru::prelude::*;
+
```
ian-h-chamberlain: Oops, I didn't mean to commit all these changes but forgot I configured my rustfmt to `group_imports = "StdExternalCrate"` by default... hopefully it's not too distracting from the main contents of this PR

Meziu: The number of files changed did seem a bit too large 😅

ian-h-chamberlain: I'll see if I can undo it to make it more clear what changed. I did end up touching a lot of files in order to add the test runner setup in doctests, which is how so many of them got reformatted (plus maybe I did a `cargo fmt` at one point).

ian-h-chamberlain:
> I suggest you transfer the runner etc. to the org and continue development from there. This PR is already taking shape!

@Meziu sounds good, I might ask you to make me admin on that repo too in case I need to adjust Actions settings or something. It should be transferring now...

ian-h-chamberlain:
Okay, the repository was transferred and this PR updated to use the new location! As far as I can tell everything looks pretty good for now, and this should be ready to merge whenever others have a chance to review

.github/workflows/ci.yml
```diff
@@ -33,11 +27,15 @@ jobs:
       - name: Checkout branch
         uses: actions/checkout@v2
 
-      - uses: ./.github/actions/setup
+      # TODO point to to rust3ds/test-runner once migrated
```
Meziu: Should remove this line now.

Meziu (approved):
LGTM! The new CI seems very powerful 😄.

PR #136, opened 2023-10-01
Add `os` module for querying OS and hardware state

ian-h-chamberlain:
Most of these are dead simple and don't require any service initialization etc., so all we need is a simple wrapper. I didn't implement everything in `<3ds/os.h>` yet, but got most of the basic ones which seemed like likely use cases to me.

I based this PR off of #135 for now so that the tests could run, but I can retarget if that one has more stuff to address before merging.

### DIFF
--- a/ctru-rs/src/os.rs
+++ b/ctru-rs/src/os.rs
@@ -1,6 +1,16 @@
 //! Utilities to get information about the operating system and hardware state.
 
-/// System version information.
+/// System version information. This struct is used for both kernel and firmware versions.
+///
+/// # Example
+/// ```
+/// # let _runner = test_runner::GdbRunner::default();
+/// let firm_version = ctru::os::firm_version();
+/// assert_ne!(firm_version.major(), 0);
+///
+/// let kernel_version = ctru::os::kernel_version();
+/// assert_ne!(kernel_version.major(), 0);
+/// ```
 #[derive(Clone, Copy)]
 pub struct Version(u32);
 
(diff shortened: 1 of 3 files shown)


ctru-rs/src/os.rs
```diff
@@ -0,0 +1,154 @@
...
+            1 => Self::Bad,
+            2 => Self::Decent,
+            3 => Self::Good,
+            other => panic!("Got unexpected WiFi strength value {other}"),
+        }
+    }
+}
+
+/// Get the current value of the stereoscopic 3D slider on a scale from 0.0­–­1.0.
+pub fn current_3d_slider_state() -> f32 {
```
Meziu, line 147: I intended to add this in tandem with some other changes to the `hid` service I'm working with. Sure, it's not required to use `Hid` in this circumstance, but it makes much more sense (since `hid` handles hardware input by definition) and since I'm also working on using the volume slider (which instead *requires* the `hid` service).

I don't like the redundancy usually found within `libctru` (usually generated by some OS-level redundancy...), so I'd like to avoid having this function in two places. Do you think we should keep it here or move it to hid?

Meziu:
I find this already good enough to merge, however I'd wait for the test-runner PR to merge first, which I'll need a bit of time to review. 😅

ctru-rs/src/os.rs
ian-h-chamberlain: Well, from what I can tell reading `libctru` code, the main difference is:

* functions defined in `os.h` are a simple pointer read of a well-known address `((osSharedConfig_s*)OS_SHAREDCFG_VADDR)->slider_3d`

* HID functions use `svcSendSyncRequest` IPC, which is also why it requires the HID service to be initialized first

I put this in here to match `libctru`'s structure, and after looking more carefully at the implementation I think it probably makes sense to keep it that way. Since these functions don't require an initialized HID service to be used, I'm not sure it makes sense to add them to our `Hid` implementation either.

ian-h-chamberlain:
> I find this already good enough to merge, however I'd wait for the test-runner PR to merge first, which I'll need a bit of time to review. 😅

@Meziu rebased to `master` since this PR didn't actually have any changes that required the `ci/reusable-actions` branch. It should be mergeable before that one goes in and there was no reason to block it that I could think of.

Meziu (approved):
Simple and straigthforward changes. Looks good 👍.

PR #137, opened 2023-10-15
Camera rework and small fixes

Meziu:
Big (and messy) PR incoming!

Changes related to the `cam` service:
- Rework of many `cam` functions to fix weird behaviour or crashes and to streamline the service. Still WIP (Trimming still needs some research and the image calibration configs don't seem to work).
- Removed the need for ALL redundant "image size" requests in function arguments.
- Support for `BothOutwardCam` using a specific `take_picture` implementation.
- The `camera-image` example now takes 3D images (and is abnormally slow thanks to CPU rendering).
- The `camera-image` example doesn't need all the specific `const` values to work properly.

Some other minor changes that were made in the mean time:
- Renamed white/blacklist functions to allow/blocklist.
- Safely wrapped `Console::set_window`.
- Properly documented GSPGPU events.
- Added missing functionality for the `Hid` module, such as volume slider and accelerometer/gyroscope. After some more research (done for the changes) I noticed the service requires to be exclusively usable from one instance, using `ServiceReference`, as it modifies global states which are normally un-trackable.
- Changed how `ServiceReference` checks for service usage. Poison handling is now a feature!
- Fixed some `panic!` cases that are now handled as new `Error` types. (also renamed `NdspError` to just `Error` following the module namespaces changes)
- Added warnings and changed functionality to avoid dangerous unsafe cases when using `Gfx` on the VRAM.
- Translated a couple of calls to static inlines after the merge of #133.

P.S.
I'm sorry for mixing up the changes to the cam service and those related to TODO or other small fixes. I had initially envisioned this PR as a "todo fix" PR but as I worked on it I noticed that `cam` was going to need a lot of work.


Meziu:
It seems like nobody on github ever used stereo calibration before (not even devkitPro's own examples include it) nor image quality calibration. In my tests it seems like `CAMU_SetStereoCameraCalibrationData` has no effect, so I'm wondering how these functions are supposed to work. The same goes for trimming, of which I have understood only that it accepts images of the "standard" supported resolutions (like VGA and TOP_LCD) but that I still can't grasp in its entirety (does it accept all resolutions with an area multiple of 256?).

@steveice10 Sorry for the tag, but since you've written the `cam:u` implementation for `libctru`, and have a lot of experience with FBI and cameras in general, would you mind sharing some of your knowledge with us? I've spent a couple of weeks reverse-engineering the `cam` service, but your insight would be greatly appreciated.

Steveice10:
> Sorry for the tag, but since you've written the `cam:u` implementation for `libctru`, and have a lot of experience with FBI and cameras in general, would you mind sharing some of your knowledge with us? I've spent a couple of weeks reverse-engineering the `cam` service, but your insight would be greatly appreciated.

It's been a *long* time since I've looked at the cam stuff but I can try to answer your questions the best I can:

1. For `CAMU_SetStereoCameraCalibrationData`:
  * You need to use `cam:s` (I think `cam:c` also works?) for privileged access. All 3 cam services share the same handler but have state checks on what service needs to be used to access them.
  * If you want an example of its usage, it's used by the System Settings app to perform camera calibration.
  * Camera calibration data is saved by cam to the first 0x40 of config block 0x00060000 if you want to look at that data.

2. For trimming I don't think I ever played with it thoroughly enough beyond standard dimensions to know what exactly it will and won't accept.

Meziu:
Thank you so much for the information! I didn't expect your quick response 😄.

>     * You need to use `cam:s` (I think `cam:c` also works?) for privileged access. All 3 cam services share the same handler but have state checks on what service needs to be used to access certain commands.

Interesting. This effectively means I'd need to re-implement all IPC calls to use my own `cam:s` handle, correct? In `libctru` everything uses `cam:u`, so that probably won't work. Also, would that change bring any trade-offs that you know of, or is the handle to use only up to the specific needs?

>     2. For trimming I don't think I ever played with it thoroughly enough beyond standard dimensions to know what exactly it will and won't accept.

No problem, I've already been experimenting with it enough. I should be able to find some conclusive evidence in no time, i hope...

I guess I should try to merge this PR more or less as-is and work on re-implementing everything over `cam:s` if possible later, since it would be way too disruptive right now.

Steveice10:
> Interesting. This effectively means I'd need to re-implement all IPC calls to use my own `cam:s` handle, correct? In `libctru` everything uses `cam:u`, so that probably won't work. Also, would that change bring any trade-offs that you know of, or is the handle to use only up to the specific needs?

I haven’t looked at how you have things set up here, but you basically can use all the same command code, you just need to have a code path to open “cam:s” instead of “cam:u” where you open the service handle.

I don’t think there’s any trade-offs for us. The way it’s set up is intended so Nintendo can prevent standard applications from accessing certain things by only allowing games to use lower-privilege service variants, but for homebrew we can just have access to all services.

ctru-rs/examples/camera-image.rs
```diff
@@ -75,28 +76,42 @@ fn main() {
...
-                .take_picture(
-                    &mut buf,
-                    WIDTH.try_into().unwrap(),
-                    HEIGHT.try_into().unwrap(),
-                    WAIT_TIMEOUT,
-                )
+                .take_picture(&mut buf, WAIT_TIMEOUT)
                 .expect("Failed to take picture");
 
+            let image_size = camera.final_view_size();
```
ian-h-chamberlain: Could do `let (width, height) = ` if you want to make it more clear

ctru-rs/src/console.rs
```diff
@@ -174,16 +201,126 @@ impl<'screen> Console<'screen> {
...
+    /// of the new window based on the row/column coordinates of a full-screen console.
+    /// The second pair is the new width and height.
     ///
-    /// # Safety
+    /// # Example
     ///
-    /// This function is unsafe because it does not validate whether the input will produce
-    /// a console that actually fits on the screen.
-    // TODO: Wrap this safely.
+    /// ```no_run
```
ian-h-chamberlain: Do these need to be `no_run`? With #135 merged they would hopefully work in CI?

ctru-rs/src/console.rs
```diff
@@ -174,16 +201,126 @@ impl<'screen> Console<'screen> {
...
+    ///
+    /// println!("Phew, finally a breath of fresh air.");
+    /// #
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn reset_window(&mut self) {
+        let width = self.max_width();
+
+        unsafe { consoleSetWindow(self.context.as_mut(), 0, 0, width.into(), 30) };
```
ian-h-chamberlain: nit: could just call `self.set_window(0, 0, width.into(), 30)`

ctru-rs/src/console.rs
```diff
@@ -174,16 +201,126 @@ impl<'screen> Console<'screen> {
...
+    /// // The maximum width for the top screen (without any alterations) is 50 characters.
+    /// assert_eq!(top_console.max_width(), 50);
+    /// #
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn max_width(&self) -> u8 {
+        match self.screen.as_raw() {
+            ctru_sys::GFX_TOP => {
+                if unsafe { ctru_sys::gfxIsWide() } {
```
ian-h-chamberlain, line 315: It would be nice to use our safe `is_wide()` but would require a downcast to `TopScreen` so probably not worth it

ctru-rs/src/lib.rs
```diff
@@ -87,15 +87,24 @@ fn panic_hook_setup() {
...
-                    hid.scan_input();
-                    let keys = hid.keys_down();
-                    if keys.contains(KeyPad::SELECT) {
+            // Due to how the Hid service operates, we can't safely use 2 handles to it at the same time.
+            // Furthermore, the panic hook runs before the panic cleanup is done, which means that any other handles
+            // to the service will still be alive during this process.
+            // Regardless, we can "unsafely" spin up a new instance, since the module won't be used any further from the main process,
+            // which is going to get cleaned up right after this loop.
+            unsafe {
+                let _ = ctru_sys::hidInit();
```
ian-h-chamberlain: If initialization fails here it might be best to avoid calling `hidScanInput` etc. I'm not sure if it makes sense to just `loop {}` or immediately exit in that case though. Or maybe even just `while apt.main_loop() {}` ?

ctru-rs/src/services/gfx.rs
```diff
@@ -298,22 +307,71 @@ impl Gfx {
...
+                Ok(())
+            },
+            || unsafe { ctru_sys::gfxExit() },
+        )?;
+
+        Ok(Self {
+            top_screen: RefCell::new(TopScreen::new()),
+            bottom_screen: RefCell::new(BottomScreen),
+            _service_handler: handler,
+        })
```
ian-h-chamberlain: nit: this logic could be shared between the two `with_formats` functions similar to what we had previously

ctru-rs/src/services/hid.rs
```diff
@@ -282,11 +317,195 @@ impl Hid {
...
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[doc(alias = "HIDUSER_EnableGyroscope")]
+    #[doc(alias = "HIDUSER_DisableGyroscope")]
+    pub fn set_gyroscope(&mut self, enabled: bool) {
+        if enabled {
+            let _ = unsafe { ctru_sys::HIDUSER_EnableGyroscope() };
+        } else {
+            let _ = unsafe { ctru_sys::HIDUSER_DisableGyroscope() };
```
ian-h-chamberlain: Should we return a `Result` for these instead of discarding the return value? I'm not sure in what scenarios these might fail...

ctru-rs/src/services/hid.rs
```diff
@@ -282,11 +317,195 @@ impl Hid {
...
+    /// hid.scan_input();
+    ///
+    /// // This call fails if the accelerometer was not previously enabled.
+    /// let acceleration = hid.accelerometer_vector()?;
+    /// #
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[doc(alias = "hidAccelRead")]
+    pub fn accelerometer_vector(&self) -> Result<(i16, i16, i16), Error> {
```
ian-h-chamberlain: personally I'd probably go for a public type alias for `accelVector` / `angularRate`, or a wrapper type or something instead of a 3-tuple. Not a big deal though

ctru-rs/src/services/hid.rs
```diff
@@ -74,8 +80,21 @@ bitflags! {
     }
 }
 
+/// Error enum for generic errors within the [`Hid`] service.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum Error {
```
ian-h-chamberlain, line 86: this should probably be `#[non_exhaustive]` like most of our enums

ctru-rs/src/services/ndsp/mod.rs
```diff
@@ -67,7 +77,7 @@ pub enum InterpolationType {
 
 /// Errors returned by [`ndsp`](self) functions.
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
-pub enum NdspError {
+pub enum Error {
```
ian-h-chamberlain, line 90: Probably should also be `#[non_exhaustive]`

ctru-rs/src/services/cam.rs
```diff
@@ -381,115 +604,111 @@ pub trait Camera {
...
+        // Run checks for all trimming possibilities.
+        match trimming {
+            Trimming::Centered(trim_size) => unsafe {
+                let view_size: (i16, i16) = self.view_size().into();
+                let trim_size: (i16, i16) = trim_size.into();
+
+                // Trim sizes are within the view.
+                assert!(
+                    trim_size.0 <= view_size.0 && trim_size.1 <= view_size.1,
+                    "trimmed view is bigger than the camera view"
```
ian-h-chamberlain: Is this something that is possible for the user to configure? Does it make sense to return an `Err` for them to handle it in that case?

ian-h-chamberlain (approved):
I didn't have a chance to deeply read the new camera code yet but mostly LGTM! Couple minor comments but otherwise all seems good

ctru-rs/src/services/hid.rs
```diff
@@ -282,11 +317,195 @@ impl Hid {
...
+    ///
+    /// hid.scan_input();
+    ///
+    /// let volume = hid.volume_slider();
+    /// #
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[doc(alias = "HIDUSER_GetSoundVolume")]
+    pub fn volume_slider(&self) -> f32 {
```
adryzz, line 377: I feel like it should be clear that this is not the actual volume slider reading but the volume reading.

For the volume slider, 3D slider, etc there's `mcuhwc` to wrap still

ctru-rs/src/lib.rs
Meziu: `apt.main_loop()` won't be able to read for user input (like the home button) without the use of the HID service. At this point, we either exit immediately (and suggest 3dslink as an alternative debugger) or do something like `std::thread::sleep(Duration::from_secs(5))`, which I'm not a big fan of.

ctru-rs/src/services/hid.rs
Meziu: I believe it is the actual volume slider reading. Where did you get this information?

ctru-rs/src/services/cam.rs
Meziu: It wouldn't make sense for the user to pass a bigger trim size than the original view. The trimming is meant to cut *down* the view so to get a particular detail of the image.

Other than panicking, we could either return an `Error` or just "clamp" the trim view to the original size (e.g. if the original width is 400 and the user trims to 500, we just return the image with the original width of  400).

Maybe the second approach would be preferable?

ctru-rs/src/services/hid.rs
adryzz: well... the MCU is the device handling these peripherals, and `mcu:hwc` communicates directly with it.

If you read from the MCU and move the slider, the data it reads is pretty "coarse".

https://github.com/devkitPro/libctru/blob/c579a65655900f6fd1bcbad42b0d68d80caba303/libctru/source/services/mcuhwc.c#L117C1-L129C1

Meziu: That’s true, but it also wouldn’t be the only time that HorizonOS leaves multiple ways to handle the same task. 😅

It’s possible they are equally as correct, or maybe one is a more sophisticated (and possibly differently working) version of the other. In my opinion it makes sense for the slider to be in HID, together with the other faceplate buttons.

ctru-rs/src/services/cam.rs
Meziu: I've come to the conclusion that this situation cannot be handled with clamping or other measures, as the final image resolution isn't guaranteed to adhere to the imposed standards. It's best to place the responsibility on the developer by making this issue known.

Meziu:
This PR is finally ready to merge.

ctru-rs/examples/movement.rs
```diff
@@ -34,13 +36,17 @@ fn main() {
         // Be careful: reading without activating the sensors (as done before this loop) will result in a panic.
         println!(
             "\x1b[3;0HAcceleration: {:?}              ",
-            hid.accelerometer_vector()
-                .expect("could not retrieve acceleration vector")
+            Into::<(i16, i16, i16)>::into(
```
ian-h-chamberlain: Can this be `From::from` ?

```suggestion
            <(i16, i16, i16)>::from(
```

ian-h-chamberlain (approved):

ctru-rs/src/lib.rs
```diff
@@ -88,18 +87,9 @@ fn panic_hook_setup() {
...
-                Ok(mut hid) => loop {
-                    hid.scan_input();
-                    let keys = hid.keys_down();
-                    if keys.contains(KeyPad::SELECT) {
-                        break;
-                    }
-                },
-                Err(e) => println!("Error while intializing Hid controller during panic: {e}"),
-            }
+            std::thread::sleep(std::time::Duration::from_secs(5));
```
ian-h-chamberlain: I could see this making panics a little harder to debug sometimes, but it does seem like it's tricky to make sure the Hid / console are initialized to do the other one... I suppose apps can always install their own hook to record panic info somewhere, if they need.

ctru-rs/examples/movement.rs
Meziu: Yeah it should be, sorry for the inconvenience.

ctru-rs/src/lib.rs
Meziu: Losing this functionality would be quite disastrous, it's one of the things I enjoy the most about our toolchain. I'll look a bit into other possibilities, otherwise I'll push a version without any panic hook.

Meziu: I've decided to remove the panic handler entirely. I will miss it very much :cry:.

PR #139, opened 2023-10-25
add `set_initial_text` swkbd method

adryzz:
Added the missing method for setting the initial text in the keyboard as documented [here](https://libctru.devkitpro.org/swkbd_8h.html#a3188c3c8f62da1f740e195976dcabf1f).

Some software keyboard methods are still missing, such as predictive input and custom dictionary words, but i didn't wanna implement them yet, as they require more data structures to wrap.

### DIFF
--- a/ctru-rs/src/applets/swkbd.rs
+++ b/ctru-rs/src/applets/swkbd.rs
@@ -7,7 +7,8 @@
 
 use bitflags::bitflags;
 use ctru_sys::{
-    self, swkbdInit, swkbdInputText, swkbdSetButton, swkbdSetFeatures, swkbdSetHintText, SwkbdState,
+    self, swkbdInit, swkbdInputText, swkbdSetButton, swkbdSetFeatures, swkbdSetHintText,
+    swkbdSetInitialText, SwkbdState,
 };
 use libc;
 use std::fmt::Display;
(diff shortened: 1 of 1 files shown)


adryzz:
oops, forgot to run `cargo fmt`

ctru-rs/src/applets/swkbd.rs
```diff
@@ -336,6 +337,30 @@ impl SoftwareKeyboard {
...
+    /// use ctru::applets::swkbd::SoftwareKeyboard;
+    /// let mut keyboard = SoftwareKeyboard::default();
+    ///
+    /// keyboard.set_initial_text("Write here what you like!");
+    /// #
+    /// # }
+    #[doc(alias = "swkbdSetInitialText")]
+    pub fn set_initial_text(&mut self, text: &str) {
+        unsafe {
+            let nul_terminated: String = text.chars().chain(once('\0')).collect();
```
Meziu, line 359: You should probably use CString for a more idiomatic conversion.

Meziu (approved):
I intend on changing these services already because there are still some flaws (mainly stuff that came from their original design) but this is a small and quick addition. lgtm

ctru-rs/src/applets/swkbd.rs
ian-h-chamberlain: Agreed, I think this could just say `CString::new(text).unwrap()`

ian-h-chamberlain (approved):

ctru-rs/src/applets/swkbd.rs
adryzz: i just copied what `set_hint_text` does, as the signature is the same

PR #140, opened 2023-10-27
Implement `ac:u` and `ac:i` services (Wi-Fi info) + example

adryzz:
Implements the `ac` service, both `ac:u` (user) and `ac:i` methods (intended for mset).

Did this because i needed some of these methods for an app

Methods implemented:
- [x] `wait_internet_connection` (returns immediately for some reason independently of network connection, have to investigate)
- [x] `get_wifi_status`
- [x] `get_wifi_security`
- [x] `get_wifi_ssid`
- [x] `get_proxy_enabled`
- [x] `get_proxy_port`
- [x] `get_proxy_username`
- [x] `get_proxy_password`
- [x] `load_network_slot` (`ac:i`)
- [ ] `ACU_SetAllowApType`
- [ ] `ACU_SetNetworkArea`
- [ ] `ACU_SetRequestEulaVersion`
- [ ] `ACU_CreateDefaultConfig`
- [ ] `ACU_ConnectAsync`, `ACU_GetLastErrorCode` and `ACU_GetLastDetailErrorCode` (all at once)
- [ ] `ACI_GetNetworkWirelessEssidSecurity`

(screenshot didnt work)
PXL_20231028_105524873

### DIFF
--- a/ctru-rs/examples/wifi-info.rs
+++ b/ctru-rs/examples/wifi-info.rs
@@ -1,0 +1,43 @@
+//! Wi-Fi info example
+//!
+//! This example prints out all the info about the console's network, like SSID, security, proxy info...
+
+use ctru::{prelude::*, services::ac::Ac};
+
+fn main() {
+    ctru::use_panic_handler();
+
+    let gfx = Gfx::new().expect("Couldn't obtain GFX controller");
+    let mut hid = Hid::new().expect("Couldn't obtain HID controller");
+    let apt = Apt::new().expect("Couldn't obtain APT controller");
+
+    let _console = Console::new(gfx.top_screen.borrow_mut());
+
+    let mut ac = Ac::new().expect("Couldn't get an AC handle");
+
+    print_network_info(&ac).expect("Error while gathering network info");
+    println!("Press START to exit.");
+
+    while apt.main_loop() {
+        hid.scan_input();
+
+        if hid.keys_down().contains(KeyPad::START) {
+            break;
+        }
+
+        gfx.wait_for_vblank();
+    }
+}
+
+fn print_network_info(ac: &Ac) -> ctru::Result<()> {
+    let connected = ac.get_wifi_status()?;
+    println!("Wi-Fi connected: {}", connected);
+
+    // Some methods error out if the console isn't connected
+    if connected {
+        println!("Wi-Fi SSID: {}", ac.get_wifi_ssid()?);
+        println!("Wi-Fi security: {:?}", ac.get_wifi_security()?);
+    }
+
+    Ok(())
+}
--- a/ctru-rs/examples/wifi-info.rs
+++ b/ctru-rs/examples/wifi-info.rs
@@ -29,7 +29,7 @@ fn main() {
 
 fn print_network_info(ac: &Ac) -> ctru::Result<()> {
     let connected = ac.get_wifi_status()?;
-    println!("Wi-Fi connected: {}", connected);
+    println!("Wi-Fi status: {:?}", connected);
 
     // Some methods error out if the console isn't connected
     if connected {
--- a/ctru-rs/examples/wifi-info.rs
+++ b/ctru-rs/examples/wifi-info.rs
@@ -37,6 +37,13 @@ fn print_network_info(ac: &Ac) -> ctru::Result<()> {
     if connected {
         println!("Wi-Fi SSID: {}", ac.get_wifi_ssid()?);
         println!("Wi-Fi security: {:?}", ac.get_wifi_security()?);
+        let proxied = ac.get_proxy_enabled()?;
+        println!("Proxy enabled: {}", proxied);
+        if proxied {
+            println!("Proxy port: {}", ac.get_proxy_port()?);
+            println!("Proxy username: {}", ac.get_proxy_username()?);
+            println!("Proxy password: {}", ac.get_proxy_password()?);
+        }
     }
 
     Ok(())
--- a/ctru-rs/examples/wifi-info.rs
+++ b/ctru-rs/examples/wifi-info.rs
@@ -5,8 +5,6 @@
 use ctru::{prelude::*, services::ac::Ac};
 
 fn main() {
-    ctru::use_panic_handler();
-
     let gfx = Gfx::new().expect("Couldn't obtain GFX controller");
     let mut hid = Hid::new().expect("Couldn't obtain HID controller");
     let apt = Apt::new().expect("Couldn't obtain APT controller");
--- a/ctru-rs/src/error.rs
+++ b/ctru-rs/src/error.rs
@@ -96,8 +96,6 @@ pub enum Error {
     },
     /// An error that doesn't fit into the other categories.
     Other(String),
-    /// UTF-8 buffer to [String] conversion error
-    Utf8(std::string::FromUtf8Error),
 }
 
 impl Error {
@@ -139,12 +137,6 @@ impl From<ResultCode> for Error {
     }
 }
 
-impl From<std::string::FromUtf8Error> for Error {
-    fn from(err: std::string::FromUtf8Error) -> Self {
-        Error::Utf8(err)
-    }
-}
-
 impl fmt::Debug for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
@@ -165,7 +157,6 @@ impl fmt::Debug for Error {
                 .field("wanted", wanted)
                 .finish(),
             Self::Other(err) => f.debug_tuple("Other").field(err).finish(),
-            Self::Utf8(e) => f.debug_tuple("Utf8").field(e).finish(),
         }
     }
 }
@@ -190,7 +181,6 @@ impl fmt::Display for Error {
             }
             Self::BufferTooShort{provided, wanted} => write!(f, "the provided buffer's length is too short (length = {provided}) to hold the wanted data (size = {wanted})"),
             Self::Other(err) => write!(f, "{err}"),
-            Self::Utf8(e) => write!(f, "UTF-8 conversion error: {e}")
         }
     }
 }
--- a/ctru-rs/src/error.rs
+++ b/ctru-rs/src/error.rs
@@ -96,6 +96,8 @@ pub enum Error {
     },
     /// An error that doesn't fit into the other categories.
     Other(String),
+    /// UTF-8 buffer to [String] conversion error
+    Utf8(std::string::FromUtf8Error),
 }
 
 impl Error {
@@ -137,6 +139,12 @@ impl From<ResultCode> for Error {
     }
 }
 
+impl From<std::string::FromUtf8Error> for Error {
+    fn from(err: std::string::FromUtf8Error) -> Self {
+        Error::Utf8(err)
+    }
+}
+
 impl fmt::Debug for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
@@ -157,6 +165,7 @@ impl fmt::Debug for Error {
                 .field("wanted", wanted)
                 .finish(),
             Self::Other(err) => f.debug_tuple("Other").field(err).finish(),
+            Self::Utf8(e) => f.debug_tuple("Utf8").field(e).finish(),
         }
     }
 }
@@ -181,6 +190,7 @@ impl fmt::Display for Error {
             }
             Self::BufferTooShort{provided, wanted} => write!(f, "the provided buffer's length is too short (length = {provided}) to hold the wanted data (size = {wanted})"),
             Self::Other(err) => write!(f, "{err}"),
+            Self::Utf8(e) => write!(f, "UTF-8 conversion error: {e}")
         }
     }
 }
--- a/ctru-rs/src/services/ac.rs
+++ b/ctru-rs/src/services/ac.rs
@@ -32,7 +32,7 @@ impl Ac {
         }
     }
 
-    /// Waits for an internet connection
+    /// Waits for an internet connection.
     ///
     /// # Example
     ///
@@ -61,7 +61,7 @@ impl Ac {
         }
     }
 
-    /// Returns whether the console is connected to Wi-Fi
+    /// Returns the current Wi-Fi connection status.
     ///
     /// # Example
     ///
@@ -74,23 +74,32 @@ impl Ac {
     ///
     /// let ac = Ac::new()?;
     ///
-    /// println!("Wi-Fi connected: {}", ac.get_wifi_status()?);
+    /// println!("Wi-Fi status: {:?}", ac.get_wifi_status()?);
     /// #
     /// # Ok(())
     /// # }
     /// ```
     #[doc(alias = "ACU_GetWifiStatus")]
-    pub fn get_wifi_status(&self) -> crate::Result<bool> {
+    pub fn get_wifi_status(&self) -> crate::Result<NetworkStatus> {
         unsafe {
             let mut ret = 0u32;
             ResultCode(ctru_sys::ACU_GetStatus(&mut ret))?;
 
-            Ok(ret == 3)
+            Ok(
+                match ret {
+                    0 => NetworkStatus::None,
+                    1 => NetworkStatus::Idle,
+                    2 => NetworkStatus::LANConnected,
+                    3 => NetworkStatus::WANConnected,
+                    _ => Err(crate::Error::Other(format!("Unknown value {}", ret)))
+                }
+            )
         }
     }
 
-    /// Returns whether the console is connected to Wi-Fi
+    /// Returns the [`SecurityMode`] of the currently connected network, or error if the console isn't connected to any network.
     ///
+    /// You can check if the console is connected to a network using [`Ac::get_wifi_status()`].
     /// # Example
     ///
     /// ```
@@ -101,6 +110,10 @@ impl Ac {
     /// use ctru::services::ac::Ac;
     ///
     /// let ac = Ac::new()?;
+    ///
+    /// if ac.get_wifi_status()? == NetworkStatus::WANConnected {
+    ///     println!("Network security: {:?}", ac.get_wifi_security()?);
+    /// }
     ///
     /// #
     /// # Ok(())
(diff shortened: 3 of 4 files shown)


ctru-rs/examples/wifi-info.rs
```diff
@@ -0,0 +1,52 @@
+//! Wi-Fi info example
+//!
+//! This example prints out all the info about the console's network, like SSID, security, proxy info...
+//!
+//! ```
```
ian-h-chamberlain: stray ` ``` ` comment?

ctru-rs/src/services/ac.rs
```diff
@@ -0,0 +1,331 @@
+use crate::error::ResultCode;
+use crate::services::fs::FsMediaType;
```
ian-h-chamberlain: Couple of clippy warnings flagged here if you don't mind addressing those

ctru-rs/src/services/ac.rs
```diff
@@ -0,0 +1,331 @@
...
+        unsafe { ctru_sys::acExit() };
+    }
+}
+
+#[doc(alias = "acSecurityMode")]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(u32)]
+pub enum SecurityMode {
+    Open = 0,
+    WEP40Bit = 1,
```
ian-h-chamberlain: To make sure the constants stay in sync you should be able to write things like `WEP40Bit = ctru_sys::AC_WEP_40BIT`

ctru-rs/src/services/ac.rs
```diff
@@ -0,0 +1,331 @@
...
+    #[doc(alias = "acExit")]
+    fn drop(&mut self) {
+        unsafe { ctru_sys::acExit() };
+    }
+}
+
+#[doc(alias = "acSecurityMode")]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(u32)]
+pub enum SecurityMode {
```
ian-h-chamberlain, line 345: This should probably be marked `#[non_exhaustive]` just in case other types get added in the future

ctru-rs/src/services/ac.rs
```diff
@@ -0,0 +1,331 @@
...
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[doc(alias = "ACU_GetProxyPassword")]
+    pub fn get_proxy_password(&self) -> crate::Result<String> {
+        unsafe {
+            let mut vec = vec![0u8; 0x20];
+            ResultCode(ctru_sys::ACU_GetProxyPassword(vec.as_mut_ptr()))?;
+
+            // how do i handle this error?
```
ian-h-chamberlain: Usually we'd have a `From` implementation that converts from these error types to `ctru::Error`. In this case it would probably make sense to add a new `Error` variant that wraps `std::string::FromUtf8Error`.

ctru-rs/src/services/ac.rs
```diff
@@ -0,0 +1,331 @@
...
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[doc(alias = "ACI_LoadNetworkSetting")]
+    pub fn load_network_slot(&self, slot: NetworkSlot) -> crate::Result<()> {
+        unsafe {
+            ResultCode(ctru_sys::ACI_LoadNetworkSetting(slot as u32))?;
+            Ok(())
+        }
+    }*/
```
ian-h-chamberlain: leftover TODO?

ctru-rs/src/services/ac.rs
```diff
@@ -0,0 +1,331 @@
...
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[doc(alias = "ACU_GetWifiSecurityMode")]
+    pub fn get_wifi_security(&self) -> crate::Result<SecurityMode> {
+        unsafe {
+            let mut ret = 0u32;
+            ResultCode(ctru_sys::ACU_GetSecurityMode(&mut ret))?;
+            // fix this, for some reason the bindings have the type as a struct and not enum
+            // and so i can't impl TryFrom automatically
```
ian-h-chamberlain, line 140: Hmm, strange, on my system it looks like it's just a `libc::c_uint` alias? What does the struct definition look like? I wonder if we might just need to adjust some flags in `ctru-sys/build.rs` to change how the bindings are generated...

ctru-rs/src/services/ac.rs
```diff
@@ -0,0 +1,331 @@
...
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[doc(alias = "ACU_GetWifiStatus")]
+    pub fn get_wifi_status(&self) -> crate::Result<bool> {
+        unsafe {
+            let mut ret = 0u32;
+            ResultCode(ctru_sys::ACU_GetStatus(&mut ret))?;
+
+            Ok(ret == 3)
```
ian-h-chamberlain: Could we return an enum here instead of a boolean, to handle the different numerical values?

ian-h-chamberlain (approved):
Mostly looks reasonable to me! Couple clippy lints that could use fixing and a few comments about error handling / enum types but mostly fine I think

ctru-rs/src/services/ac.rs
adryzz: that's the problem, i usually would do that, but the bindings have it as a struct and not an enum, and so the variants aren't available as consts.

moreover, I can't auto derive `From<T>` because of this.

ctru-rs/src/services/ac.rs
adryzz: here is the enum definition

https://github.com/devkitPro/libctru/blob/bf4a24a4e7e3e9fa1642845818d3d59ccb8c506a/libctru/include/3ds/services/ac.h#L17C6-L8C6

i really don't know why it's just a `c_uint` alias.

ctru-rs/src/services/ac.rs
adryzz: no, this function works, it just requires a libctru version bump, as it got recently added by TuxSH.

ctru-rs/src/services/ac.rs
adryzz: the docs here say that it returns zero when not connected and 3 when connected, so i just implemented it this way.

adryzz:
also, will add that 2 days ago [libctru 2.3.0](https://github.com/devkitPro/libctru/releases/tag/v2.3.0) released, adding the 2 methods needed for this.

adryzz:
fixed a few of the issues, will now investigate the bindings, as that's the main issue apart from the `libctru` version bump

ctru-rs/src/services/ac.rs
adryzz: oh i was wrong

adryzz:
never mind, i got the enum issue sorted out.

All that remains now is the libctru version bump for the `load_network_slot` setting.

also cleaned up the history a bit, now all commits build

ian-h-chamberlain (approved):
Sorry for the long wait on a review! The new changes look good to me, but I think you have a few leftover comments that aren't relevant anymore, and you probably need to merge in the latest `master` since a couple things (in particular `use_panic_handler()`) have changed recently. If you have a chance to clean up some of those bits then I think this should be good to merge!

adryzz:
will do right now!

ctru-rs/src/services/ac.rs
```diff
@@ -0,0 +1,347 @@
...
+    ///
+    /// ac.load_network_slot(NetworkSlot::Second)?;
+    /// #
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[doc(alias = "ACI_LoadNetworkSetting")]
+    pub fn load_network_slot(&self, slot: NetworkSlot) -> crate::Result<()> {
+        unsafe {
+            ResultCode(ctru_sys::ACI_LoadNetworkSetting(slot as u32))?;
```
ian-h-chamberlain, line 333: Hmm, I guess this API is not available yet in the base `devkitpro/devkitarm` Docker image we use for CI, which apparently only has `libctru 2.2.2-2` installed.

I think this could be resolved a couple different ways, offhand
- We could always update to latest devkitPro in CI. Might not work since IIRC the devkitPro maintainers block downloads from Github Actions runner IPs...
- We could do a little work to add a `#[cfg]` for this based on the `libctru` 
version (easier once #141 is merged)
- Wait for a new docker image to be published... These come every couple months normally but not sure how long it would have to wait. It could be commented out or `#[cfg(ignore)]` or something in the meantime just to get this PR merged
- Fork and maintain our own docker image 🤮 not my preference this one

@rust3ds/active any thoughts?

adryzz: the `#[cfg]` based on version is the cleanest way, yes, but I don't see any meaningful reason to support multiple `libctru` versions at the same time (especially on something like this, if anything)

so yeah

adryzz:
time to rebase this

adryzz:
perfect! the unit tests failing have nothing to do with this PR afaik, so it all works now!

Jhynjhiruu:
Seems like, once this and #156 get merged, all 3DS networking functionality besides Download Play is done - that'll be neat to see!

ctru-rs/src/services/ac.rs
ian-h-chamberlain: There is a new devkitARM docker image as of `20240202` so I think this is good (and linter passed)

ian-h-chamberlain (approved):
LGTM, @rust3ds/active if anyone else would like to have a look I'll wait a little but I think this is ready to merge!

CI may be completely broken due to https://github.com/rust3ds/test-runner/issues/16 but the linter passed at least so I think we can merge and fix up any test issues that might show up once that's resolved

adryzz:
> Seems like, once this and #156 get merged, all 3DS networking functionality besides Download Play is done - that'll be neat to see!

what other parts of libctru/3ds services are missing? i may just try implementing other stuff too

Jhynjhiruu:
> > Seems like, once this and #156 get merged, all 3DS networking functionality besides Download Play is done - that'll be neat to see!
> 
> what other parts of libctru/3ds services are missing? i may just try implementing other stuff too

I think at this point it's mostly less important stuff like `boss`, `news`, `qtm` etc. - though `ir` still needs some work, since the Circle Pad Pro code doesn't work properly on New 3DS. I don't think `irrst` is done, either, though that should probably be merged into `hid` in my opinion.
Just go to the libctru repo and compare the source files with what's done on the ctru-rs side.

ctru-rs/src/services/ac.rs
```diff
@@ -0,0 +1,349 @@
...
+    /// ```
+    #[doc(alias = "acInit")]
+    pub fn new() -> crate::Result<Ac> {
+        unsafe {
+            ResultCode(ctru_sys::acInit())?;
+            Ok(Ac(()))
+        }
+    }
+
+    /// Waits for an internet connection
```
Meziu: Nit: put full stops at the end of phrases used for documentation.

ctru-rs/src/services/ac.rs
Meziu: If that's really all this function does, it might be better to rename it to something like `is_connected` and keep the boolean return value. Otherwise, we could keep the current name but return a more descriptive enum with named variants, such as `Connected` and `NotConnected`.

ctru-rs/src/services/ac.rs
```diff
@@ -0,0 +1,349 @@
...
+    pub fn get_wifi_status(&self) -> crate::Result<bool> {
+        unsafe {
+            let mut ret = 0u32;
+            ResultCode(ctru_sys::ACU_GetStatus(&mut ret))?;
+
+            Ok(ret == 3)
+        }
+    }
+
+    /// Returns the [SecurityMode] of the currently connected network, or error if the console isn't connected to any network.
```
Meziu: Nit: we usually use code blocks to specify doc links.
```suggestion
    /// Returns the [`SecurityMode`] of the currently connected network, or error if the console isn't connected to any network.
```

ctru-rs/src/services/ac.rs
```diff
@@ -0,0 +1,349 @@
...
+            let mut vec = vec![0u8; 0x20];
+            ResultCode(ctru_sys::ACU_GetProxyPassword(vec.as_mut_ptr()))?;
+
+            Ok(String::from_utf8(vec)?)
+        }
+    }
+
+    /// Load the selected network slot, if present.
+    ///
+    /// Note: this method requires `ac:i` access
```
Meziu, line 313: Is this access given by the normal `Ac::new()` implementation, or should the user do something more to gain it? Reimplementing IPS calls isn't exactly a straightforward process.

ctru-rs/src/services/ac.rs
```diff
@@ -0,0 +1,349 @@
...
+    /// ```
+    #[doc(alias = "ACU_GetSSID")]
+    pub fn get_wifi_ssid(&self) -> crate::Result<String> {
+        unsafe {
+            let mut len = 0u32;
+            ResultCode(ctru_sys::ACU_GetSSIDLength(&mut len))?;
+            // we don't really need space for the terminator
+            let mut vec = vec![0u8; len as usize];
+            ResultCode(ctru_sys::ACU_GetSSID(vec.as_mut_ptr()))?;
+            Ok(String::from_utf8(vec)?)
```
Meziu, line 187: Rather than returning the `String`'s error, it might be wiser to check for `Ac::get_wifi_status()` and return an error based on *that*. We could try to do that in every case needed, such as proxy getters. Otherwise, we could write a custom `ac::Error` to hold those errors.

I'd like to avoid implementing `Error::Utf8`.

ctru-rs/src/error.rs
```diff
@@ -96,6 +96,8 @@ pub enum Error {
     },
     /// An error that doesn't fit into the other categories.
     Other(String),
+    /// UTF-8 buffer to [String] conversion error
+    Utf8(std::string::FromUtf8Error),
```
Meziu: Refer to my other comment about this.

ctru-rs/src/services/ac.rs
```diff
@@ -0,0 +1,349 @@
...
+    ///
+    /// let ac = Ac::new()?;
+    /// #
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[doc(alias = "acInit")]
+    pub fn new() -> crate::Result<Ac> {
+        unsafe {
+            ResultCode(ctru_sys::acInit())?;
```
Meziu: Note: if we ever plan to implement methods such as setting up wifi connections, we would need a `&mut` reference and the `ServiceReference` structure to keep the thing thread-safe. For now it's not needed.

ctru-rs/src/services/ac.rs
```diff
@@ -0,0 +1,349 @@
...
+    ///
+    /// println!("Waiting for an internet connection...");
+    /// ac.wait_internet_connection()?;
+    /// println!("Connected.");
+    /// #
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[doc(alias = "acWaitInternetConnection")]
+    pub fn wait_internet_connection(&self) -> crate::Result<()> {
```
Meziu, line 70: This function completely locks the thread until a connection is established, without a timeout. It's a bit... radical. We should probably suggest users to implement their own connection checks.

Meziu (changes requested):

ctru-rs/src/services/ac.rs
adryzz: the rosalina loader (e.g. the default) already gives all the arm11 permissions, but if using the hax2 loader, these permissions are missing unless ran from mset. can probably remove this, only thing is it may fail in the Citra CI because of permissions, although i haven't checked.

adryzz:
uhh good news. i tested around the `get_wifi_status` function, and there are definitely more values.

3dbrew has very little information on this, and i was only able to confirm these:
- `0`: Hardware Wi-Fi switch is disabled
- `1`: Enabled but not connected
- `2`: LAN connected
- `3`: WAN connected

I also was able to get in a few occasions the values `102` and `300` while turning on/off the Wi-Fi network, although i have no clue what they actually mean, and this suggests there may be even more values.

`#[non_exhaustive]` works i guess

ctru-rs/src/services/ac.rs
ian-h-chamberlain: I think normally Citra grants basically _all_ permissions since it's just emulated, although it's possible some methods are just unimplemented / stubs, I don't think permissions are usually denied.

ctru-rs/src/services/ac.rs
```diff
@@ -74,24 +74,32 @@ impl Ac {
...
+    /// println!("Wi-Fi status: {:?}", ac.get_wifi_status()?);
     /// #
     /// # Ok(())
     /// # }
     /// ```
     #[doc(alias = "ACU_GetWifiStatus")]
-    pub fn get_wifi_status(&self) -> crate::Result<bool> {
+    pub fn get_wifi_status(&self) -> crate::Result<NetworkStatus> {
         unsafe {
             let mut ret = 0u32;
```
ian-h-chamberlain, line 99: Make sure you `cargo fmt`, looks like this failed the lint check https://github.com/rust3ds/ctru-rs/actions/runs/8026159593/job/21928158019?pr=140#step:6:83

ian-h-chamberlain (approved):
LGTM other than fixing linter. CI should be up and working now again!

ctru-rs/src/services/ac.rs
```diff
@@ -0,0 +1,386 @@
...
+    ///
+    /// let ac = Ac::new()?;
+    ///
+    /// ac.load_network_slot(NetworkSlot::Second)?;
+    /// #
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[doc(alias = "ACI_LoadNetworkSetting")]
+    pub fn load_network_slot(&self, slot: NetworkSlot) -> crate::Result<()> {
```
Meziu: This must require `&mut self`. That also means it's definitely necessary to use the `ServiceReference` routine to setup the service.

ctru-rs/src/services/ac.rs
```diff
@@ -0,0 +1,386 @@
...
+    ///
+    /// let ac = Ac::new()?;
+    ///
+    /// println!("Wi-Fi status: {:?}", ac.get_wifi_status()?);
+    /// #
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[doc(alias = "ACU_GetWifiStatus")]
+    pub fn get_wifi_status(&self) -> crate::Result<NetworkStatus> {
```
Meziu: Please remove all `get_*` prefixes. It's not standard for Rust APIs to have that.

Meziu (changes requested):
The PR is done, but I really can't merge without these last few changes :pray:

adryzz:
oh, i completely forgot this was still open, will update asap

adryzz:
okay, one small issue: how do we deal with errors related to UTF-8 now? i checked and you can still run into those while being connected (e.g. a proxy username not being valid UTF-8, set from a homebrew app), so the `wifi_status()`-based error approach wouldn't work.

so do we create a custom AC error instead of propagating `FromUtf8Error`?

Meziu:
> so do we create a custom AC error instead of propagating `FromUtf8Error`?

Yeah, that should be a good choice.

ctru-rs/src/services/ac.rs
```diff
@@ -0,0 +1,396 @@
...
+    #[doc(alias = "ACI_LoadNetworkSetting")]
+    pub fn load_network_slot(&mut self, slot: NetworkSlot) -> crate::Result<()> {
+        unsafe {
+            ResultCode(ctru_sys::ACI_LoadNetworkSetting(slot as u32))?;
+            Ok(())
+        }
+    }
+}
+
+impl Drop for Ac {
```
Meziu: You need to remove this `Drop` impl, now that you’re using `ServiceReference`

adryzz: oh completely forgot about this

adryzz:
also, since we know the contents of the buffer are invalid text, we could still return the buffer as a `Vec<u8>`

and problem: the functions can return both `crate::Error` and `ac::Error`. how do we handle that? i see that `crate::Error` does not propagate any module-specific errors

adryzz:
i completely forgot about this PR being still open, will get it done with asap then

PR #141, opened 2023-10-29
Clean up ctru-sys version checking

ian-h-chamberlain:
Relates to #100 but maybe not totally resolved by this PR

- Don't issue a warning for "mismatched version" since the new bindings make this irrelevant
- Minor build script refactoring and code cleanup
- Output a `release` version (-X) for libctru env vars
- Revert version to 0.5.0, which is higher than we were using before the major 22.x version but should be ok since we never released to crates.io
- Unconditionally link `ctru` (not `ctrud`) due to https://github.com/rust3ds/cargo-3ds/issues/14 
	- [x] don't love this, maybe there's a better option

- [x] TODO I should update the README since it was documenting the version number convention but that also doesn't apply anymore

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,17 +75,17 @@ jobs:
       # unless cargo-3ds actually runs them as separate commands. See
       # https://github.com/rust3ds/cargo-3ds/issues/44 for more details
       - name: Build lib and integration tests
-        run: cargo 3ds test --no-run --tests --package ctru-rs
+        run: cargo 3ds test --no-run --tests
 
       - name: Run lib and integration tests
         uses: rust3ds/test-runner/run-tests@v1
         with:
-          args: --tests --package ctru-rs
+          args: --tests
 
       - name: Build and run doc tests
         uses: rust3ds/test-runner/run-tests@v1
         with:
-          args: --doc --package ctru-rs
+          args: --doc
 
       - name: Upload citra logs and capture videos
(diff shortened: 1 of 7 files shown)


Meziu:
Nice PR. As i said regarding the bindings PR, I am convinced that `ctru-sys`'s version number should be separate from the libctru version that it links. I've also seen some changes regarding the general structure of tests and other things. I'll look around your changes once I manage to find some time :sweat_smile:.

ian-h-chamberlain:
> Nice PR. As i said regarding the bindings PR, I am convinced that `ctru-sys`'s version number should be separate from the libctru version that it links. I've also seen some changes regarding the general structure of tests and other things. I'll look around your changes once I manage to find some time 😅.

No rush to review! With the way we're generating the bindings now I agreed in terms of version numbering, which is why I bumped it back down to 0.5.0. Most of the other stuff was just adding build script warnings and refactoring but feel free to review at your own pace 👍

Meziu (approved):
Great, the changes were as I expected them to be. The only "issue" would be finding a way to link `ctrud` correctly, but for the sake of this PR it might not be necessary. Merge if you want.

ian-h-chamberlain:
I was able to figure out something that I think works a little better for linking `ctru`, which in combination with https://github.com/rust3ds/cargo-3ds/pull/47 I think makes for a much better solution. Let me know what you think!

Meziu (approved):

PR #149, opened 2023-12-28
Applets fixup and feature implementation

Meziu:
This is a simple PR containing the long awaited fixes for the general use of the applets and some additions to the `Software Keyboard`.

### DIFF
--- a/ctru-rs/examples/software-keyboard.rs
+++ b/ctru-rs/examples/software-keyboard.rs
@@ -11,6 +11,11 @@ fn main() {
     let gfx = Gfx::new().unwrap();
     let _console = Console::new(gfx.top_screen.borrow_mut());
 
+    // Prepares a software keyboard with two buttons: one to cancel input and one
+    // to accept it. You can also use `SoftwareKeyboard::new()` to launch the keyboard
+    // with different configurations.
+    let mut keyboard = SoftwareKeyboard::default();
+
     println!("Press A to enter some text or press Start to exit.");
 
     while apt.main_loop() {
@@ -22,14 +27,9 @@ fn main() {
 
         // Check if the user request to write some input.
         if hid.keys_down().contains(KeyPad::A) {
-            // Prepares a software keyboard with two buttons: One to cancel input and one
-            // to accept it. You can also use `SoftwareKeyboard::new()` to launch the keyboard in different
-            // configurations.
-            let mut keyboard = SoftwareKeyboard::default();
-
             // Raise the software keyboard. You can perform different actions depending on which
             // software button the user pressed.
-            match keyboard.get_string(2048) {
+            match keyboard.get_string(2048, &apt, &gfx) {
                 Ok((text, Button::Right)) => println!("You entered: {text}"),
                 Ok((_, Button::Left)) => println!("Cancelled"),
                 Ok((_, Button::Middle)) => println!("How did you even press this?"),
--- a/ctru-rs/examples/software-keyboard.rs
+++ b/ctru-rs/examples/software-keyboard.rs
@@ -2,8 +2,11 @@
 //!
 //! This example showcases the use of the Software Keyboard applet to receive text input from the user.
 
-use ctru::applets::swkbd::{Button, SoftwareKeyboard};
+use ctru::applets::swkbd::{Button, CallbackResult, SoftwareKeyboard};
 use ctru::prelude::*;
+
+use std::borrow::Cow;
+use std::ffi::CString;
 
 fn main() {
     let apt = Apt::new().unwrap();
@@ -15,6 +18,21 @@ fn main() {
     // to accept it. You can also use `SoftwareKeyboard::new()` to launch the keyboard
     // with different configurations.
     let mut keyboard = SoftwareKeyboard::default();
+
+    // Custom filter callback to handle the given input.
+    // Using this callback it's possible to integrate the applet
+    // with custom error messages when the input is incorrect.
+    keyboard.set_filter_callback(Some(Box::new(|str| {
+        // The string is guaranteed to contain valid Unicode text, so we can safely unwrap and use it as a normal `&str`.
+        if str.to_str().unwrap().contains("boo") {
+            return (
+                CallbackResult::Retry,
+                Some(Cow::Owned(CString::new("Ah, you scared me!").unwrap())),
+            );
+        }
+
+        (CallbackResult::Ok, None)
+    })));
 
     println!("Press A to enter some text or press Start to exit.");
 
--- a/ctru-rs/examples/software-keyboard.rs
+++ b/ctru-rs/examples/software-keyboard.rs
@@ -14,10 +14,7 @@ fn main() {
     // Prepares a software keyboard with two buttons: one to cancel input and one
     // to accept it. You can also use `SoftwareKeyboard::new()` to launch the keyboard
     // with different configurations.
-    let mut keyboard = SoftwareKeyboard::new(
-        ctru::applets::swkbd::Kind::Normal,
-        ctru::applets::swkbd::ButtonConfig::LeftMiddleRight,
-    );
+    let mut keyboard = SoftwareKeyboard::default();
 
     println!("Press A to enter some text or press Start to exit.");
 
--- a/ctru-rs/examples/software-keyboard.rs
+++ b/ctru-rs/examples/software-keyboard.rs
@@ -14,7 +14,10 @@ fn main() {
     // Prepares a software keyboard with two buttons: one to cancel input and one
     // to accept it. You can also use `SoftwareKeyboard::new()` to launch the keyboard
     // with different configurations.
-    let mut keyboard = SoftwareKeyboard::default();
+    let mut keyboard = SoftwareKeyboard::new(
+        ctru::applets::swkbd::Kind::Normal,
+        ctru::applets::swkbd::ButtonConfig::LeftMiddleRight,
+    );
 
     println!("Press A to enter some text or press Start to exit.");
 
--- a/ctru-rs/examples/software-keyboard.rs
+++ b/ctru-rs/examples/software-keyboard.rs
@@ -5,7 +5,6 @@
 use ctru::applets::swkbd::{Button, CallbackResult, SoftwareKeyboard};
 use ctru::prelude::*;
 
-use std::borrow::Cow;
 use std::ffi::CString;
 
 fn main() {
@@ -27,7 +26,7 @@ fn main() {
         if str.to_str().unwrap().contains("boo") {
             return (
                 CallbackResult::Retry,
-                Some(Cow::Owned(CString::new("Ah, you scared me!").unwrap())),
+                Some(CString::new("Ah, you scared me!").unwrap()),
             );
         }
 
--- a/ctru-rs/src/applets/mii_selector.rs
+++ b/ctru-rs/src/applets/mii_selector.rs
@@ -277,6 +277,7 @@ impl MiiSelector {
     /// # }
     /// ```
     #[doc(alias = "miiSelectorLaunch")]
+    #[allow(unused_variables)]
     pub fn launch(&mut self, apt: &Apt, gfx: &Gfx) -> Result<Selection, Error> {
         let mut return_val = Box::<ctru_sys::MiiSelectorReturn>::default();
         unsafe { ctru_sys::miiSelectorLaunch(self.config.as_mut(), return_val.as_mut()) }
--- a/ctru-rs/src/applets/mii_selector.rs
+++ b/ctru-rs/src/applets/mii_selector.rs
@@ -281,8 +281,7 @@ impl MiiSelector {
     /// # }
     /// ```
     #[doc(alias = "miiSelectorLaunch")]
-    #[allow(unused_variables)]
-    pub fn launch(&mut self, apt: &Apt, gfx: &Gfx) -> Result<Selection, Error> {
+    pub fn launch(&mut self, _apt: &Apt, _gfx: &Gfx) -> Result<Selection, Error> {
         let mut return_val = Box::<ctru_sys::MiiSelectorReturn>::default();
         unsafe { ctru_sys::miiSelectorLaunch(self.config.as_mut(), return_val.as_mut()) }
 
--- a/ctru-rs/src/applets/mii_selector.rs
+++ b/ctru-rs/src/applets/mii_selector.rs
@@ -4,6 +4,8 @@
 //! The selected Mii is readable as a [`Mii`].
 
 use crate::mii::Mii;
+use crate::services::{apt::Apt, gfx::Gfx};
+
 use bitflags::bitflags;
 use std::{ffi::CString, fmt};
 
@@ -252,9 +254,8 @@ impl MiiSelector {
 
     /// Launch the Mii Selector.
     ///
-    /// Depending on the configuration, the Mii Selector window will appear either on the bottom screen (default behaviour) or the top screen (see [`Options::USE_TOP_SCREEN`]).
-    ///
-    /// TODO: UNSAFE OPERATION, LAUNCHING APPLETS REQUIRES GRAPHICS, WITHOUT AN ACTIVE GFX THIS WILL CAUSE A SEGMENTATION FAULT.
+    /// Depending on the configuration, the Mii Selector window will appear either
+    /// on the bottom screen (default behaviour) or the top screen (see [`Options::USE_TOP_SCREEN`]).
     ///
     /// # Example
     ///
@@ -276,7 +277,7 @@ impl MiiSelector {
     /// # }
     /// ```
     #[doc(alias = "miiSelectorLaunch")]
-    pub fn launch(&mut self) -> Result<Selection, Error> {
+    pub fn launch(&mut self, apt: &Apt, gfx: &Gfx) -> Result<Selection, Error> {
         let mut return_val = Box::<ctru_sys::MiiSelectorReturn>::default();
         unsafe { ctru_sys::miiSelectorLaunch(self.config.as_mut(), return_val.as_mut()) }
 
--- a/ctru-rs/src/applets/mii_selector.rs
+++ b/ctru-rs/src/applets/mii_selector.rs
@@ -262,6 +262,10 @@ impl MiiSelector {
     /// ```no_run
     /// # use std::error::Error;
     /// # fn main() -> Result<(), Box<dyn Error>> {
+    /// # use ctru::services::{apt::Apt, gfx::Gfx};
+    /// #
+    /// # let gfx = Gfx::new().unwrap();
+    /// # let apt = Apt::new().unwrap();
     /// #
     /// use ctru::applets::mii_selector::{MiiSelector, Options};
     ///
@@ -271,7 +275,7 @@ impl MiiSelector {
     /// let opts = Options::ENABLE_CANCEL & Options::ENABLE_GUESTS;
     /// mii_selector.set_options(opts);
     ///
-    /// let result = mii_selector.launch()?;
+    /// let result = mii_selector.launch(&apt, &gfx)?;
     /// #
     /// # Ok(())
     /// # }
--- a/ctru-rs/src/applets/swkbd.rs
+++ b/ctru-rs/src/applets/swkbd.rs
@@ -399,7 +399,10 @@ impl SoftwareKeyboard<'_> {
     /// # let _runner = test_runner::GdbRunner::default();
     /// # fn main() {
     /// #
+    /// use std::borrow::Cow;
+    /// use std::ffi::CString;
     /// use ctru::applets::swkbd::{SoftwareKeyboard, CallbackResult};
+    /// 
     /// let mut keyboard = SoftwareKeyboard::default();
     ///
     /// keyboard.set_filter_callback(Some(Box::new(|str| {
--- a/ctru-rs/src/applets/swkbd.rs
+++ b/ctru-rs/src/applets/swkbd.rs
@@ -1,7 +1,6 @@
 //! Software Keyboard applet.
 //!
 //! This applet opens a virtual keyboard on the console's bottom screen which lets the user write UTF-16 valid text.
-// TODO: Implement remaining functionality (filter callbacks, etc.). Also improve "max text length" API.
 #![doc(alias = "keyboard")]
 
 use crate::services::{apt::Apt, gfx::Gfx};
@@ -9,6 +8,7 @@ use ctru_sys::{self, SwkbdState};
 
 use bitflags::bitflags;
 use libc;
+use std::ffi::CStr;
 use std::fmt::Display;
 use std::iter::once;
 use std::str;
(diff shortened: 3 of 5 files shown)


ctru-rs/src/applets/swkbd.rs
```diff
@@ -311,6 +368,71 @@ impl SoftwareKeyboard {
...
+            let closure = Box::from_raw(user as *mut Box<F>);
+
+            let text = CStr::from_ptr(text);
+            let text_slice: &str = text.to_str().unwrap();
+
+            let result = closure(text_slice);
+
+            if let Some(cstr) = result.1 {
+                *pp_message = cstr.as_ptr();
+                Box::leak(Box::new(cstr)); // Definitely SHOULD NOT do this, but as far as design goes, it's clean.
```
Meziu: There is an interesting issue with lifetimes here. Since the `libctru` code only asks back for a pointer to a string (the one that contains the "error message" to display to the user"), there is no check about the memory's validity over time time. This implies that, if the CString containing the message were to be allocated during the filter callback's runtime, the message text would be dropped and deallocated before being shown. Furthermore, being constrained by `extern "C"` FFI, we can't make any form of real lifetime checks ourselves.

As it stands, the `CString` is simply returned by the callback and then leaked, as I've deemed memory leaking a "suitable" drop-in solution for now, but alternatives are possible. I want to discuss about an elegant solution that doesn't disrupt the API too much.

ctru-rs/src/applets/swkbd.rs
```diff
@@ -311,6 +368,71 @@ impl SoftwareKeyboard {
         self.state.filter_flags = filters.bits();
     }
 
+    /// Configure a custom filtering function to validate the input.
+    ///
+    /// The callback function will return a callback result and the error message to display when the input is invalid.
+    ///
+    /// # Notes
+    ///
+    /// The filter callback will work only for the next time the keyboard is used. After using it once, it must be set again.
```
ian-h-chamberlain: Is this a limitation of the `libctru` API, or something you imposed while writing this? It seems a bit unfortunate if we have to re-box the user-provided closure every time the applet is activated, maybe we could store it as a member of the `SoftwareKeyboard` struct or something?

ctru-rs/src/applets/swkbd.rs
```diff
@@ -311,6 +368,71 @@ impl SoftwareKeyboard {
...
+    /// ```
+    /// # let _runner = test_runner::GdbRunner::default();
+    /// # fn main() {
+    /// #
+    /// use ctru::applets::swkbd::{SoftwareKeyboard, CallbackResult};
+    /// let mut keyboard = SoftwareKeyboard::default();
+    ///
+    /// keyboard.set_filter_callback(|text| {
+    ///     if text.contains("boo") {
+    ///         println!("Ah, you scared me!");
```
ian-h-chamberlain: It might be more useful to show an example of returning something other than `(Ok, None)` here, since it seems like the filter callback is more meant for sanitizing/rejecting the input rather than just acting on its value (although of course any application could do sanitization of its own after getting the value too).

ctru-rs/src/applets/swkbd.rs
```diff
@@ -311,6 +368,71 @@ impl SoftwareKeyboard {
...
+    ///     }
+    ///
+    ///     (CallbackResult::Ok, None)
+    /// });
+    /// #
+    /// # }
+    pub fn set_filter_callback<F>(&mut self, callback: F)
+    where
+        F: FnOnce(&str) -> (CallbackResult, Option<CString>),
```
ian-h-chamberlain: I'd think it would make sense for this API to be symmetrical, e.g. `&str -> (CallbackResult, Option<String>)` or `&CStr -> (CallbackResult, Option<CString>)`?

ctru-rs/src/applets/swkbd.rs
```diff
@@ -311,6 +368,71 @@ impl SoftwareKeyboard {
...
+        unsafe extern "C" fn internal_callback<F>(
+            user: *mut libc::c_void,
+            pp_message: *mut *const libc::c_char,
+            text: *const libc::c_char,
+            _text_size: libc::size_t,
+        ) -> ctru_sys::SwkbdCallbackResult
+        where
+            F: FnOnce(&str) -> (CallbackResult, Option<CString>),
+        {
+            let closure = Box::from_raw(user as *mut Box<F>);
```
ian-h-chamberlain: nit: `as` -> `.cast()`, I actually thought there was a clippy lint for that...

ctru-rs/src/applets/swkbd.rs
```diff
@@ -381,7 +503,70 @@ impl SoftwareKeyboard {
...
+    /// // The right numpad key will not be shown.
+    /// keyboard.set_numpad_keys(Some('!'), None);
+    /// #
+    /// # }
+    #[doc(alias = "swkbdSetNumpadKeys")]
+    pub fn set_numpad_keys(&mut self, left_key: Option<char>, right_key: Option<char>) {
+        let mut keys = (0, 0);
+
+        if let Some(k) = left_key {
+            keys.0 = k as i32;
```
ian-h-chamberlain, line 586: Might be good to verify the user passed an ASCII digit (I'm assuming non-ASCII is not supported here).

ctru-rs/src/applets/swkbd.rs
```diff
@@ -497,13 +735,40 @@ impl Display for Error {
...
+            ctru_sys::SWKBD_POWERPRESSED => Error::PowerPressed,
+            ctru_sys::SWKBD_PARENTAL_OK => Error::ParentalOk,
+            ctru_sys::SWKBD_PARENTAL_FAIL => Error::ParentalFail,
+            ctru_sys::SWKBD_BANNED_INPUT => Error::BannedInput,
+            ctru_sys::SWKBD_D0_CLICK => Error::ButtonPressed,
+            ctru_sys::SWKBD_D1_CLICK0 => Error::ButtonPressed,
+            ctru_sys::SWKBD_D1_CLICK1 => Error::ButtonPressed,
+            ctru_sys::SWKBD_D2_CLICK0 => Error::ButtonPressed,
+            ctru_sys::SWKBD_D2_CLICK1 => Error::ButtonPressed,
+            ctru_sys::SWKBD_D2_CLICK2 => Error::ButtonPressed,
```
ian-h-chamberlain, line 791: Do you think it would make sense to have this include which button was pressed, like `Error::ButtonPressed(Left)` or `Error::ButtonPressed(Home)`?

ctru-rs/src/applets/mii_selector.rs
```diff
@@ -270,13 +275,14 @@ impl MiiSelector {
...
     ///
-    /// let result = mii_selector.launch()?;
+    /// let result = mii_selector.launch(&apt, &gfx)?;
     /// #
     /// # Ok(())
     /// # }
     /// ```
     #[doc(alias = "miiSelectorLaunch")]
-    pub fn launch(&mut self) -> Result<Selection, Error> {
+    #[allow(unused_variables)]
```
AzureMarker: nit: Instead of blanket allowing unused variables, rename unused arguments to start with an underscore.

ctru-rs/src/applets/swkbd.rs
```diff
@@ -234,30 +288,33 @@ impl SoftwareKeyboard {
...
     ///
-    /// let button = keyboard.write_exact(&mut buffer)?;
+    /// let button = keyboard.write_exact(&mut buffer, &apt, &gfx)?;
     /// #
     /// # Ok(())
     /// # }
     /// ```
     #[doc(alias = "swkbdInputText")]
-    pub fn write_exact(&mut self, buf: &mut [u8]) -> Result<Button, Error> {
+    #[allow(unused_variables)]
```
AzureMarker: See other comment

ctru-rs/src/applets/swkbd.rs
AzureMarker: According to libctru, who is supposed to deallocate the string and when? We should try to emulate the memory management they expect.

ctru-rs/src/applets/swkbd.rs
```diff
@@ -448,27 +635,84 @@ impl SoftwareKeyboard {
...
+    ///     Err(_) => println!("Something wrong happened during the parental lock prompt.")
+    /// }
+    /// #
+    /// # }
+    #[doc(alias = "swkbdInputText")]
+    #[allow(unused_variables)]
+    pub fn launch(&mut self, apt: &Apt, gfx: &Gfx) -> Result<(), Error> {
+        unsafe {
+            let mut buf = [0; 10];
+            ctru_sys::swkbdInputText(self.state.as_mut(), buf.as_mut_ptr(), 10);
```
AzureMarker: Why is the buffer length 10, and what is the content after this call?

AzureMarker (commented):
Also check the lint warnings in the file-explorer example.

ctru-rs/src/applets/swkbd.rs
FenrirWolf: If you check the [source for swkbdSetFilterCallback](https://github.com/devkitPro/libctru/blob/master/libctru/source/applets/swkbd.c#L138-L143), it looks like libctru just stashes pointers to the CBs and expects the user to keep everything alive as long as it's needed.

And if that's the case, we should probably store that data in the `Swkbd` struct.

ian-h-chamberlain (approved):

ctru-rs/src/applets/swkbd.rs
ian-h-chamberlain: I was trying to think of some clever ways to express that "the `SoftwareKeyboard` must outlive anything the given closureborrows", but couldn't come up with anything obvious after a few tries. One option might be something like this?

```rs
struct SoftwareKeyboard {
    // ...
    filter_result: Option<CString> // defaults to None
}

impl SoftwareKeyboard {
    pub fn set_filter_callback<F>(&mut self, callback: F)
    where
        F: FnOnce(&str, &mut Option<CString>) -> CallbackResult,
    { 
        // probably requires an extra layer of indirection to capture self:
        let closure = |s| closure(self, s);
        // ...
        let result = closure(text_slice);
        // ...
    }
}
```

Then we'd always guarantee the filter result, if set, was alive as long as the keyboard was alive?

FenrirWolf: That's what I would recommend. I believe we also need to save the callback that way too.

ian-h-chamberlain: I think something like this would work? I haven't really tested though:

```rs
// This can't be `Clone` anymore if we have a boxed closure in there...
pub struct SoftwareKeyboard {
    state: Box<SwkbdState>,
    callback: Option<Box<dyn Fn(&CStr, &mut Option<CString>) -> CallbackResult>>,
    filtered_string: Option<CString>,
}

impl SoftwareKeyboard {
    pub fn set_filter_callback<F>(&mut self, callback: F)
    where
        F: Fn(&CStr, &mut Option<CString>) -> CallbackResult + 'static,
    {
        extern "C" fn internal_callback(
            user: *mut libc::c_void,
            pp_message: *mut *const libc::c_char,
            text: *const libc::c_char,
            _text_size: libc::size_t,
        ) -> ctru_sys::SwkbdCallbackResult {
            let this: *mut SoftwareKeyboard = user.cast();

            // maybe reset filtered_string to None here?

            unsafe {
                let text = CStr::from_ptr(text);
                let result = {
                    if let Some(callback) = &mut (*this).callback { // or unwrap?
                        callback(text, &mut (*this).filtered_string)
                    } else {
                        CallbackResult::Ok
                    }
                };

                if let Some(newstr) = &(*this).filtered_string {
                    *pp_message = newstr.as_ptr();
                }
                result.into()
            }
        }

        self.callback = Some(Box::new(callback));

        unsafe {
            ctru_sys::swkbdSetFilterCallback(
                self.state.as_mut(),
                Some(internal_callback),
                (self as *mut Self).cast(),
            );
        }
    }
}
```

ctru-rs/src/applets/swkbd.rs
```diff
@@ -311,6 +368,71 @@ impl SoftwareKeyboard {
...
+    /// keyboard.set_filter_callback(|text| {
+    ///     if text.contains("boo") {
+    ///         println!("Ah, you scared me!");
+    ///     }
+    ///
+    ///     (CallbackResult::Ok, None)
+    /// });
+    /// #
+    /// # }
+    pub fn set_filter_callback<F>(&mut self, callback: F)
```
ian-h-chamberlain: Something else just occurred to me: should we have `Option<F>` here so that users can also clear the callback if it was previously set? They could of course just pass a trivial callback like `|_| Callbackresult::Ok` but not sure if there's any reason to set the callback to `null` (via `None`) explicitly.

ctru-rs/src/applets/swkbd.rs
Meziu: The keyboard supports full UTF-8 (but not all characters are shown via the system font, of course).

ctru-rs/src/applets/swkbd.rs
Meziu: I guess it could make sense to unify `ButtonPressed`  and the `HomePressed` (and such) variants. However, in this particular context the `SWKBD_D{n}_CLICK{n}` are referring only to the "positional" buttons put at the bottom, and the `CLICK` part is still quite misleading, since there isn't actually any form of "end" to how many times the prompt can be re-tried.

ctru-rs/src/applets/swkbd.rs
Meziu: For some reason, even though the applet doesn't seem to write anything into the buffer (as it should? my guess is that for safety reasons the parental lock PIN shouldn't be accessible by the application), at the end of the prompt the applet still tries to read the memory there and crashes in an ARM panic. I used an empty buffer to give it something to read, but I've just now tested that even an empty slice of size 0 is enough.

ctru-rs/src/applets/swkbd.rs
Meziu: The closure gets dropped after use. I think it's a good idea to just store it in the configuration.

ctru-rs/src/applets/swkbd.rs
Meziu: The idea was to return "Ah, you scared me!" as the error message, since it would nicely appear in a text bubble on screen. I will change it once the filter callback API is finalized.

ctru-rs/src/applets/swkbd.rs
Meziu: Hmm, that's true. I'd like to keep the `CString` design as it more closely represents what's underneath (to avoid conversions on the heap), but the `&str -> String` design is obviously much cleaner. Tell me which one you prefer.

ctru-rs/src/applets/swkbd.rs
Meziu: That's pretty much necessary if we decide to store the closure within the `SoftwareKeyboard` configuration.

ctru-rs/src/applets/swkbd.rs
Meziu: That would definitely work. Just note that what you called `filtered_string` isn't, in fact, the original message with "filters applied". It's actually just the error message shown to the user for why the string isn't valid.

ian-h-chamberlain: Ah, I misunderstood how the filter callback worked then – seems like it would be perfect for the example then

ctru-rs/src/applets/swkbd.rs
ian-h-chamberlain: Hmm, maybe for the sake of allocation we could do `&CStr -> Cow<CStr>`? Then if someone wants to do regular string operations they could use `CStr::to_string_lossy()` or other conversion functions, then convert back afterwards? This makes the user ensure their string is valid — if we did the conversion internally, we'd have to figure out what we want to do in the error case, and I think this way probably makes it more flexible for the user to handle errors how they want.

https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=4c20c6b02b66c96b40d5806465ee5e2c

ctru-rs/src/applets/swkbd.rs
ian-h-chamberlain: Ah, so these are just the `u32` representation of those `char`s cast into `i32`s then? cool

ctru-rs/src/applets/swkbd.rs
ian-h-chamberlain: I guess I was thinking if the user exited the keyboard via e.g. either `SWKBD_D1_CLICK0` or `SWKBD_D2_CLICK0` it might make sense to return some kind of `Left` in both cases, since the app might want to know which button was pressed to exit the keyboard. I think this only applies to buttons that are configured to submit the input, not for retries, right?

ctru-rs/src/applets/swkbd.rs
ian-h-chamberlain: If the buf is size 0 now then we should probably also pass 0 as the bufsize just to be safe, right?

ctru-rs/src/applets/swkbd.rs
Meziu: Yeah, retries have a separate prompt with 2 fixed buttons (and which of those the user pressed can't be known). This refers to the button used to actually exit the `SoftwareKeyboard`.

ctru-rs/src/applets/swkbd.rs
adryzz: the Cow<Cstr> approach is what i'd do too

ctru-rs/src/applets/swkbd.rs
Meziu: > If the buf is size 0 now then we should probably also pass 0 as the bufsize just to be safe, right?

Yeah, I'm just tired :face_with_spiral_eyes:

ctru-rs/src/applets/swkbd.rs
Meziu: Tell me if you prefer the current version, and what you think should be changed. I've also moved it to the general struct's `impl` block since it needs to be visible from `write_exact()`.

ctru-rs/src/applets/swkbd.rs
Meziu: Uff, seems like `libctru` actually disposes of the callback after use. I've put the raw setter into the `write_exact` function.

ian-h-chamberlain (approved):

ctru-rs/examples/software-keyboard.rs
```diff
@@ -2,15 +2,38 @@
...
+    // Prepares a software keyboard with two buttons: one to cancel input and one
+    // to accept it. You can also use `SoftwareKeyboard::new()` to launch the keyboard
+    // with different configurations.
+    let mut keyboard = SoftwareKeyboard::default();
+
+    // Custom filter callback to handle the given input.
+    // Using this callback it's possible to integrate the applet
+    // with custom error messages when the input is incorrect.
+    keyboard.set_filter_callback(Some(Box::new(|str| {
+        // The string is guaranteed to contain valid Unicode text, so we can safely unwrap and use it as a normal `&str`.
```
AzureMarker, line 25: nit: If this is so, we could do this in our library code so the user code doesn't need to assume this invariant. It would make it more Rust-friendly, but I get that it abstracts away the CString, so it's up to you.

ctru-rs/src/applets/swkbd.rs
```diff
@@ -448,27 +659,84 @@ impl SoftwareKeyboard {
...
+    /// match parental_lock.launch(&apt, &gfx) {
+    ///     Ok(_) => println!("You can access parental-only features and settings."),
+    ///     Err(Error::ParentalFail) => println!("Is a kid trying to access this?"),
+    ///     Err(_) => println!("Something wrong happened during the parental lock prompt.")
+    /// }
+    /// #
+    /// # }
+    #[doc(alias = "swkbdInputText")]
+    #[allow(unused_variables)]
+    pub fn launch(&mut self, apt: &Apt, gfx: &Gfx) -> Result<(), Error> {
```
AzureMarker: Remove `#[allow(unused_variables)]` and use underscore prefixes on unused parameters

ctru-rs/src/applets/swkbd.rs
```diff
@@ -311,6 +383,80 @@ impl SoftwareKeyboard {
...
+            (*this).error_message = None;
+
+            let text = CStr::from_ptr(text);
+
+            let result = {
+                // Run the callback if still available.
+                if let Some(callback) = &mut (*this).callback {
+                    let (res, cstr) = callback(text);
+
+                    (*this).error_message = cstr;
```
AzureMarker, line 445: nit: Add comments to explain why we have to store the error message and then reference it in this way.

ctru-rs/src/applets/swkbd.rs
```diff
@@ -1,24 +1,36 @@
...
+use std::ffi::CStr;
 use std::fmt::Display;
 use std::iter::once;
 use std::str;
 
+type CallbackFunction = dyn Fn(&CStr) -> (CallbackResult, Option<Cow<CStr>>);
+
 /// Configuration structure to setup the Software Keyboard applet.
 #[doc(alias = "SwkbdState")]
+pub struct SoftwareKeyboard<'a> {
```
AzureMarker: nit: Add documentation to explain why there is a lifetime

ctru-rs/src/applets/swkbd.rs
```diff
@@ -369,67 +385,76 @@ impl SoftwareKeyboard {
...
+            (*this).error_message = None;
 
             let text = CStr::from_ptr(text);
-            let text_slice: &str = text.to_str().unwrap();
 
-            let result = closure(text_slice);
+            let result = {
+                // Run the callback if still available.
+                if let Some(callback) = &mut (*this).callback {
+                    let (res, cstr) = callback(text);
```
ian-h-chamberlain, line 441: I'm realizing now... if the user callback returned a `Cow::Borrowed` which references the original `text`, how can we ensure that lives long enough here? I started thinking about what the `Cow<'a>` lifetime actually represents, and I think it really is the lifetime of the incoming `text: *const libc::c_char`, which is presumably just the lifetime of the input buffer from `write_exact`, right? 

Picture this example:

```rs
keyboard.set_filter_callback(Box::new(|text| (CallbackResult::Retry, Some(Cow::Borrowed(text)))));

let mut buf = vec![0u8; 32];
let _result = keyboard.write_exact(&mut buf, &apt, &gfx);
drop(buf);

// callback gets invoked, what's it borrowing from here?
let _result = keyboard.get_string(&apt, &gfx);
```

This might explain why `libctru` resets the callback every time, since each call to `swkbdInputText` may refer to different backing buffers... all this to say, maybe it is simpler to stick with the `-> CString` API that you started off with to avoid dealing with this particular problem? It would also eliminate the `'a` from `SoftwareKeyboard` as a bonus.

Meziu: I've reverted the type to a simple `CString`.

ctru-rs/src/applets/swkbd.rs
Meziu: The lifetime requirement has been removed.

ctru-rs/examples/software-keyboard.rs
Meziu: For the input string, it makes sense to read it as a normal `&str`, since all of its features are known. However, handling the output as a `String` would require reallocations to add the needed `NUL` character at the end.

As we discussed before, the "design problem" comes to the asymmetrical nature of the callback's parameters and return values. I've decided to keep the `&Cstr -> CString` as it more closely relates to what happens behind the scenes.

Meziu:
The tests on the latest nightly will fail regardless due to #152. I will take the tests run on the "known to work" nightly as good since it also passes them on my local machine using a very recent nightly.

Issue #124, opened 2023-05-17
Misaligned pointers in thread_local static variables upon creation of a HashMap


DeltaF1:
Creating a hashmap causes the panic `misaligned pointer dereference: address must be a multiple of 0x8 but is ...`. Discovered this while trying to run Meizu's realtime-audio example https://github.com/Meziu/realtime-audio/issues/1

[This issue](https://github.com/rust-lang/rust/issues/109997) and [this](https://github.com/ykjit/yk/pull/638#issuecomment-1498697036) seem to be related. They're talking about static variables being linked in un-aligned ways. It's a bit over my head but it might be useful context.

I tracked it to the Symphonia library creating a HashMap. `HashMap::new` initializes `RandomState` with a static variable inside `thread_local!`. This is where the crash happens:

`/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/collections/hash/map.rs:3127`
```rs
        KEYS.with(|keys| {
            let (k0, k1) = keys.get();
            keys.set((k0.wrapping_add(1), k1));
            RandomState { k0, k1 }
        })
```

The minimal test case appears to be "creating a hashmap". Can someone else please run `examples/hashmaps.rs` in debug and confirm that it panics?

Meziu:
Alignment issues on thread local variables... where have I heard this [before](https://github.com/rust3ds/ctru-rs/issues/60#issuecomment-1150529440)?

Still, this looks like a pretty complex situation, most likely relating back to the TLS code in `libctru`. 

@ian-h-chamberlain since you made those changes, can you run similar tests again? It looks like nothing changed in the mean time regarding TLS in `libctru`, so it's very possible we've always used misaligned pointers (and I wonder if that has anything to do with [corrupted](https://github.com/rust3ds/pthread-3ds/pull/28) TLS destructors)...

ian-h-chamberlain:
oh no, hopefully this isn't another case of similar shenanigans, that one took _ages_ to track down...

I am able to reproduce using the `hashmaps` example, once I updated my toolchain to something recent enough. I'll see if I can figure out why the pointer is misaligned and go from there.

Edit: ok, so far this is the smallest reproduction I've found. I'm trying to inline all the calls there and see if I can narrow it down further but haven't been able to get it so far:
```rs
fn main() {
    ctru::use_panic_handler();
    let _s = std::collections::hash_map::RandomState::new();
}
```

[...]

PR #150, opened 2024-01-02
Removal of all FS code and member renaming

Meziu:
Inspired by [this](https://github.com/rust3ds/ctru-rs/issues/124#issuecomment-1738807712) comment.

The `fs` service module is the only piece of this repo that hasn't received any form of update since the evolution from the [old implementation](https://github.com/rust3ds/ctru-rs-old).

Much of the reason why I decided to change the approach on the Rust3DS toolchain was to align as much as possible with the official Rust APIs and guidelines, so that we could embrace as much of the public ecosystem as possible. However, the `fs` module has for long been left behind as it wasn't really a priority (thanks to the already existing compatibility with C files for the SD card and ROMFS) and as such received no real improvements or testing.

What I removed in this PR was the original re-write of the `std::fs` API that was made to closely resemble the standard Rust experience. At this point, all this code (which was still unable to communicate with any system `Archives` other than SDMC) results completely pointless to keep around (and even quite confusing).

If we ever need to implement the general `libctru` file-system functionality into `ctru-rs`, we should do it by either writing a completely new API that fits the needs nicely, or work to make the already existing `std::fs` API interoperable with more `Archive`s, just how it already works with SDMC and ROMFS (an idea that was already [thought of](https://github.com/rust3ds/ctru-rs-old/issues/43#issuecomment-354535841) in the old days).


ctru-rs/src/services/fs.rs
```diff
@@ -6,34 +6,23 @@
 #![doc(alias = "filesystem")]
```
AzureMarker, line 4: Update module documentation (e.g. there are no more methods exported here).

ctru-rs/src/services/fs.rs
```diff
@@ -6,34 +6,23 @@
...
-use std::path::{Path, PathBuf};
-use std::ptr;
-use std::slice;
-use std::sync::Arc;
-use widestring::{WideCStr, WideCString};
 
 bitflags! {
     #[derive(Default, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Clone, Copy)]
-    struct FsOpen: u32 {
+    struct Open: u32 {
```
AzureMarker, line 10: Are all of the remaining enums/bitflags used somewhere?

Meziu: Not all, but it makes sense to keep them around just for some nice compatibility with `ctru_sys` (in case some end-user needs it).

AzureMarker (approved):

ctru-rs/src/services/fs.rs
```diff
@@ -124,953 +111,6 @@ pub enum ArchiveID {
...
-    /// as many times as desired and the service will not exit until all
-    /// instances of Fs drop out of scope.
-    pub fn new() -> crate::Result<Fs> {
-        unsafe {
-            let r = ctru_sys::fsInit();
-            if r < 0 {
-                Err(r.into())
-            } else {
-                Ok(Fs(()))
-            }
```
ian-h-chamberlain, line 354: Do we still need `fsInit` for `std::fs` stuff to work, or does `libctru` initialize it for us? I guess `romfs:/` already has its own service, but do `sdmc:/` files still work after this change?

ian-h-chamberlain (approved):

ian-h-chamberlain:
Btw, it looks like there might be some renewed interest in updated `std::fs::path` APIs for prefixes too (https://github.com/rust-lang/rust/issues/52331), which would be nice if we could develop some kinds of `PathExt`, `FileExt` in the upstream stdlibs.

ctru-rs/src/services/fs.rs
Meziu: Yes, that work is done automatically by `libctru`.

PR #156, opened 2024-02-08
Implement UDS service with example

Jhynjhiruu:
This PR adds basic local networking functionality with the `UDS` service.
It's very much not feature-complete, but enough is implemented to be able to interface with the devkitPRO local networking example - i.e. joining a network, creating a network, sending and pulling packets, and basic configuration. This is probably enough functionality for most homebrew applications, and I plan to try and implement a Download Play sniffer with it.

Todo: finish implementing the rest of the functions; test properly; cleanup code?

An example is included (`local-networking.rs`; this is a mostly-but-not-entirely-complete port of the devkitPRO local networking example).


ctru-rs/src/services/ir_user.rs
```diff
@@ -150,7 +150,7 @@ impl IrUser {
                         shared_mem.shared_memory_layout,
                     );
 
-                    Ok(())
+                    Ok::<_, Error>(())
```
Meziu, line 153: @AzureMarker Ehm, do you remember why a closure is used here? It doesn't seem to be necessary.

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,960 @@
...
+
+/// Information about a network node.
+/// Ported to Rust so that Debug can be derived on it.
+#[allow(missing_docs)]
+#[doc(alias = "udsNodeInfo")]
+#[derive(Debug)]
+pub struct NodeInfo {
+    pub uds_friendcodeseed: u64,
+    pub username: String,
+    pub unk_x1c: u16,
```
Meziu: Unknown members, padding bytes and other "generic" data for which use isn't clear shouldn't be `pub`.

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,960 @@
...
+    /// No context bound.
+    NoContext,
+    /// Cannot send data on a network as a spectator.
+    Spectator,
+    /// No network created.
+    NoNetwork,
+    /// The provided app data buffer was too large.
+    TooMuchAppData,
+    /// ctru-rs error
+    Lib(crate::Error),
```
Meziu, line 54: Nit: This currently doesn't serve a purpose. If possible, I'd like to keep the Error types distinct (though, it might become useful if this module were to be expanded upon).

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,960 @@
...
+                node_id: value.NetworkNodeID,
+                pad_x22: value.pad_x22,
+                word_x24: value.word_x24,
+            }
+        }
+    }
+}
+
+/// Information returned from scanning for networks.
+/// Ported to Rust so that Debug can be derived on it.
```
Meziu: This should probably be a developer comment, rather than documentation.

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,960 @@
...
+        match value as u32 {
+            ctru_sys::UDSCONTYPE_Client => Ok(Self::Client),
+            ctru_sys::UDSCONTYPE_Spectator => Ok(Self::Spectator),
+            _ => Err(()),
+        }
+    }
+}
+
+/// Information about a network node.
+/// Ported to Rust so that Debug can be derived on it.
```
Meziu: Same as `NetworkScanInfo`.

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,960 @@
...
+                pad_x22: value.pad_x22,
+                word_x24: value.word_x24,
+            }
+        }
+    }
+}
+
+/// Information returned from scanning for networks.
+/// Ported to Rust so that Debug can be derived on it.
+#[allow(missing_docs)]
```
Meziu: It would be nice to add descriptions of the members.

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,960 @@
...
+    /// Size of receive buffer; max frame size * 8.
+    const RECV_BUF_SIZE: u32 = ctru_sys::UDS_DEFAULT_RECVBUFSIZE;
+
+    /// Shared memory size; must be slightly larger
+    /// than `RECV_BUF_SIZE`.
+    const SHAREDMEM_SIZE: usize = 0x3000;
+
+    /// Buffer used while scanning for networks.
+    /// This value is taken from the devkitPRO example.
+    const SCAN_BUF_SIZE: usize = 0x4000;
```
Meziu, line 417: This is quite a large number. It's more than 16kB of allocated memory for *each* call of the `scan()` function. Do we need this much memory?

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,960 @@
...
+    /// # }
+    /// ```
+    #[doc(alias = "udsInit")]
+    pub fn new(username: Option<&str>) -> Result<Self, Error> {
+        if let Some(n) = username {
+            if n.len() > 10 {
+                return Err(Error::UsernameTooLong);
+            }
+        }
+        let cstr = username.map(CString::new);
```
Meziu: Using `CString::new(username)` is preferred.

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,960 @@
...
+                Ok(c) => Some(c),
+                Err(_) => return Err(Error::UsernameContainsNull),
+            }
+        } else {
+            None
+        };
+        let handler = ServiceReference::new(
+            &UDS_ACTIVE,
+            || {
+                let ptr = cstr.map(|c| c.as_ptr()).unwrap_or(null());
```
Meziu, line 478: What happens if the pointer passed is `NUL`?

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,960 @@
...
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[doc(alias = "udsGetNetworkStructApplicationData")]
+    pub fn get_network_appdata(
+        &self,
+        network: &NetworkScanInfo,
+        max_size: Option<usize>,
+    ) -> crate::Result<Vec<u8>> {
+        let mut appdata_buffer = vec![
```
Meziu, line 583: Nit: here you could use `MaybeUninit`, to tie in better with the rest of the code.

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,960 @@
...
+        if self.service_status() == ServiceStatus::Disconnected {
+            return Err(Error::NotConnected);
+        }
+
+        Ok(unsafe { ctru_sys::udsWaitConnectionStatusEvent(next, wait) })
+    }
+
+    /// Returns the current [`ctru_sys::udsConnectionStatus`] struct.
+    ///
+    /// TODO: should this return an error if not connected?
```
Meziu: I believe so, the internal `ctru_sys::udsConnectionStatus` contains informations specific to an existing network.

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,960 @@
...
+    ///     println!("Connection status event signalled");
+    ///     let status = uds.get_connection_status()?;
+    ///     println!("Status: {status:#X?}");
+    /// }
+    /// #
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[doc(alias = "udsGetConnectionStatus")]
+    pub fn get_connection_status(&self) -> crate::Result<ctru_sys::udsConnectionStatus> {
```
Meziu: `ctru_sys::udsConnectionStatus` should get wrapped within a more standard struct that holds a safe method API.

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,960 @@
...
+    /// # let _runner = test_runner::GdbRunner::default();
+    /// # use std::error::Error;
+    /// # fn main() -> Result<(), Box<dyn Error>> {
+    /// #
+    /// use ctru::services::uds::{ConnectionType, SendFlags, Uds};
+    /// let mut uds = Uds::new(None)?;
+    ///
+    /// let networks = uds.scan(b"HBW\x10", None, None)?;
+    /// uds.connect_network(&networks[0], b"udsdemo passphrase c186093cd2652741\0", ConnectionType::Client, 1)?;
+    /// uds.send_packet(b"Hello, World!", ctru_sys::UDS_BROADCAST_NETWORKNODEID as u16, 1, SendFlags::Default)?;
```
Meziu: `ctru_sys::UDS_BROADCAST_NETWORKNODEID` and `ctru_sys::UDS_HOST_NETWORKNODEID` should be handled as bitflags. Programmatically adding nodes to the flag would be quite useful as well.

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,960 @@
...
+        })?;
+
+        let frame = unsafe { frame.assume_init() };
+        let actual_size = unsafe { actual_size.assume_init() };
+        let src_node_id = unsafe { src_node_id.assume_init() };
+
+        Ok(if actual_size == 0 {
+            None
+        } else {
+            Some((frame[..actual_size].to_vec(), src_node_id))
```
Meziu: It might be wiser to `truncate()` and possibly `shrink_to_fit()` the original vector, rather than reallocating.

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,960 @@
...
+    pub fn set_appdata(&self, data: &[u8]) -> Result<(), Error> {
+        if self.service_status() != ServiceStatus::Server {
+            return Err(Error::NoNetwork);
+        }
+
+        if data.len() > Self::MAX_APPDATA_SIZE {
+            return Err(Error::TooMuchAppData);
+        }
+
+        ResultCode(unsafe { ctru_sys::udsSetApplicationData(data.as_ptr().cast(), data.len()) })?;
```
Meziu, line 1173: Is the `appdata` saved within the struct? If it just holds the pointer it might cause segmentation faults when the original buffer is dropped.

Meziu (commented):
Great PR! A bit long for my liking, but I saw you followed our standards for the public API. Sadly, I have no way to thoroughly test the changes since I can't make a local network (I only have one 3DS). Is there any way I can test things out, maybe with an emulator?

ctru-rs/src/services/ir_user.rs
AzureMarker: This was done so I could use `?`, like the unstable `try` block.

Jhynjhiruu:
> Is there any way I can test things out, maybe with an emulator?

I've looked a little bit into it and I think it's maybe possible to connect to a 3DS's network with another device, but it'd require more knowledge about how Wi-Fi works than I have.
I think Citra supports relaying local networking data for multiplayer, but I don't know how to set it up.

ctru-rs/src/services/ir_user.rs
Jhynjhiruu: This change was the easiest way to fix a compile error introduced by my implementation of `FromResidual`. If I handle errors a different way, this change can be reverted.

ctru-rs/src/services/uds.rs
Jhynjhiruu: Right, sorry. I basically blindly copied that without really thinking about it.

ctru-rs/src/services/uds.rs
Jhynjhiruu: There are a bunch of functions which return this error type that also need to be able to return ctru-rs errors. I'm not massively well-versed in how ctru-rs's error system works, and this was the solution I came up with for this. If there's a better way to handle it, I'm very happy to change it, but I don't know how to do that myself.

ctru-rs/src/services/uds.rs
Jhynjhiruu: This buffer has to hold all of the info for beacons found during the scanning process, which can essentially be arbitrarily large depending on how many beacons there are; the devkitPRO example uses 16KB.
However, looking at the example code, I realise now I was a bit silly, because the buffer can absolutely just be allocated on the heap once when the service is initialised and reused for each scan. Is this acceptable?

ctru-rs/src/services/uds.rs
Jhynjhiruu: How so? `CString::new` can't take an `Option`, and also we want to raise an error if the provided username contains NULL bytes, so that the user doesn't unexpectedly set a different username to what they thought.

ctru-rs/src/services/uds.rs
Jhynjhiruu: The 3DS's system username is used.

ctru-rs/src/services/uds.rs
Meziu: I guess so, it would surely help mitigate memory fragmentation. Now I imagine a room filled with 200 Nintendo 3DS trying to download play and the chaos that would ensue 🥸

ctru-rs/src/services/uds.rs
Jhynjhiruu: Does `MaybeUninit` work with non-constant sizes? I _could_ always allocate a max-sized array, but that seems a bit wasteful.

ctru-rs/src/services/uds.rs
Jhynjhiruu: I haven't yet checked for sure, but I'm pretty sure that network node IDs aren't actually bitflags, even though they probably should be - the devkitPRO example implies that you access nodes numerically. I'll check this with my consoles.

ctru-rs/src/services/uds.rs
Jhynjhiruu: I forgot to check the standard library documentation to see if there's a better way to do this. Whoops.

ctru-rs/src/services/uds.rs
Jhynjhiruu: It's not entirely clear from the `libctru` documentation, but reading [the code](https://github.com/devkitPro/libctru/blob/master/libctru/source/services/uds.c#L744), it seems like it's copied to a TLS buffer. Oh god, not TLS again.

ctru-rs/src/services/uds.rs
Meziu: You should specify that in the docs 👍

ctru-rs/src/services/uds.rs
Jhynjhiruu: Yep, confirmed - ID 0x0001 is the host, then 0x0002, 0x0003, 0x0004 etc. are the nodes on the network. Sending to node 0x0000 does nothing, sending to node 0xFFFF broadcasts to all nodes, sending to nodes not on the network does nothing, sending to nodes > 16 also does nothing, and sending to the device's own node returns an error (0xC90113FA, uds, status, wrong arg, not found). I suppose I could add an enum for this?

ctru-rs/src/services/uds.rs
Jhynjhiruu: It seems like calling it when not connected actually returns success. On New 3DS, I get
```rs
ConnectionStatus {
    status: 3,
    unk_x4: 0,
    cur_node_id: 2,
    unk_xa: 0,
    unk_xc: [0, 0, 0, 0, 0, 0, 0, 0],
    total_nodes: 0,
    max_nodes: 0,
    node_bitmask: 0,
}
```
and on Old 3DS I get
```rs
ConnectionStatus {
    status: 3,
    unk_x4: 0,
    cur_node_id: 0,
    unk_xa: 0,
    unk_xc: [0, 0, 0, 0, 0, 0, 0, 0],
    total_nodes: 0,
    max_nodes: 0,
    node_bitmask: 0,
}
```
consistently (they're identical besides `cur_node_id`).

Meziu: Interesting. At this point I wonder whether the `status` member itself specifies whether or not the client is connected to a network. We should probably try to understand its meaning.

Jhynjhiruu: `libctru` doesn't describe it at all, so it would very much be a reverse-engineering effort. From my 20 seconds of experimentation, it looks like a value of 6 represents connected. Presumably it's a bitfield.

ctru-rs/src/services/uds.rs
Meziu: I think I wanted to remove a degree of complexity (`map` -> `new` with separate `match`) and convert the final error directly, but after revisiting the changes with the other parts of the code it's definitely fine as is.

ctru-rs/src/services/uds.rs
Meziu: I got confused with uninitialized slices, sorry :sweat_smile:

ctru-rs/src/services/uds.rs
Meziu: So the packet needs to be sent out each time for every node that needs to receive it? Other than for broadcasts I guess. Wack,  I just believed it worked like/together with the nodes bitmask.

ctru-rs/src/services/uds.rs
Meziu: I mean, it's not like TLS storage doesn't work. The linker just hates us in particular. :laughing:

Jhynjhiruu:
Functions left to implement:

- [ ] udsGenerateNodeInfo (not really intended for use standalone, I think, and doesn't really need to be in the safe API)
- [ ] udsGetNodeInfoUsername (the safe API replaces the `NodeInfo` type and that decodes the string, so there's no need for this)
- [ ] udsCheckNodeInfoInitialized (I have no idea why this is supposed to be useful, since you always know this already)
- [ ] udsGenerateDefaultNetworkStruct (also not really intended for use standalone, used as part of `Uds::new()`)
- [x] udsWaitDataAvailable (would be good to have, but not really needed since you can just check whether `Uds::pull_packet()` returns `Some` or `None`)
- [x] udsEjectClient
- [x] udsEjectSpectator
- [ ] udsUpdateNetworkAttribute (definitely not really intended for use standalone, other functions call this internally)
- [x] udsSetNewConnectionsBlocked
- [x] udsAllowSpectators
- [x] udsGetNodeInformation

So, only about 5½ that would be actually useful, and they're mostly trivial.

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,1273 @@
...
+pub struct NodeInfo {
+    pub uds_friendcodeseed: u64,
+    pub username: String,
+    unk_x1c: u16,
+    pub flag: u8,
+    pad_x1f: u8,
+    pub node_id: NodeID,
+    pad_x22: u16,
+    word_x24: u32,
+}
```
ian-h-chamberlain, line 184: Personally, I'd probably implement this by just wrapping the underlying type, especially since a lot of the debug output will just end up as padding this way. Maybe something like this?

```rust
#[derive(Copy, Clone)]
pub struct NodeInfo(ctru_sys::udsNodeInfo);

impl Debug for NodeInfo {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        write!(f, "NodeInfo(")?;

        let user_and_flag = unsafe { self.0.__bindgen_anon_1.__bindgen_anon_1 };

        f.debug_struct("udsNodeInfo")
            .field("uds_friendcodeseed", &self.0.uds_friendcodeseed)
            .field(
                "username",
                &String::from_utf16_lossy(&user_and_flag.username),
            )
            .field("flag", &user_and_flag.flag)
            .field("NetworkNodeID", &self.0.NetworkNodeID)
            .finish()?;

        write!(f, ")")
    }
}
```

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,1273 @@
...
+                }
+            }),
+        }
+    }
+}
+
+/// Status of the connection.
+#[doc(alias = "udsConnectionStatus")]
+#[derive(Debug)]
+pub struct ConnectionStatus {
```
ian-h-chamberlain: Why does this need to be reimplemented in Rust? Again could we just wrap the upstream type?

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,1273 @@
...
+    /// Broadcast to all nodes
+    Broadcast,
+}
+
+impl From<NodeID> for u16 {
+    fn from(value: NodeID) -> Self {
+        match value {
+            NodeID::None => 0,
+            NodeID::Node(node) => node as u16,
+            NodeID::Broadcast => ctru_sys::UDS_BROADCAST_NETWORKNODEID as u16,
```
ian-h-chamberlain, line 146: Might make sense for us to just generate this constant as a `u16` via bindgen's [`ParseCallbacks::int_macro`](https://docs.rs/bindgen/latest/bindgen/callbacks/trait.ParseCallbacks.html#method.int_macro), to help prevent conversions.

In fact I wonder if we can't do this for a bunch of other constants and stop using `as`, maybe I'll file a separate issue for this more broadly

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,1273 @@
...
+        }
+        let cstr = username.map(CString::new);
+        let cstr = if let Some(conv) = cstr {
+            match conv {
+                Ok(c) => Some(c),
+                Err(_) => return Err(Error::UsernameContainsNull),
+            }
+        } else {
+            None
+        };
```
ian-h-chamberlain: I think you could simply this a bit if we had an `impl From<std::ffi::NulError> for Error`:

```suggestion
        let cstr = username.map(CString::new).transpose()?;
```

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,1273 @@
...
+        let handler = ServiceReference::new(
+            &UDS_ACTIVE,
+            || {
+                let ptr = cstr.map(|c| c.as_ptr()).unwrap_or(null());
+
+                ResultCode(unsafe { ctru_sys::udsInit(Self::SHAREDMEM_SIZE, ptr) })?;
```
ian-h-chamberlain, line 480: Does this need to be `Pin` or something, to make sure the string isn't moved while the service is initialized? I would imagine that if you moved the resulting `UDS` struct the pointer copied into the `udsInit` could become invalid.

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,1273 @@
...
+}
+
+static UDS_ACTIVE: Mutex<()> = Mutex::new(());
+
+impl Uds {
+    /// Size of one frame.
+    const RECV_FRAME_SIZE: usize = ctru_sys::UDS_DATAFRAME_MAXSIZE as usize;
+
+    /// Size of receive buffer; max frame size * 8.
+    const RECV_BUF_SIZE: u32 = ctru_sys::UDS_DEFAULT_RECVBUFSIZE;
```
ian-h-chamberlain, line 409: Is there any reason to make this an associated const if it's not `pub`? The extra comment might be helpful but only if these are exported

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,1273 @@
...
+    /// This value is taken from the devkitPRO example.
+    const SCAN_BUF_SIZE: usize = 0x4000;
+
+    /// The maximum number of nodes that can ever be connected
+    /// to a network (16). Can be further limited.
+    const MAX_NODES: u8 = ctru_sys::UDS_MAXNODES as u8;
+
+    /// The maximum amount of app data any server can provide.
+    /// Limited by the size of a struct in libctru.
+    const MAX_APPDATA_SIZE: usize = 200;
```
ian-h-chamberlain: This could probably avoid a hardcode with a little trickery (not sure if `SCAN_BUF_SIZE` and `SHAREDMEM_SIZE` would benefit from something similar): https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9e32fe75295dc99916039224351f337c

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,1273 @@
...
+        })?;
+
+        let networks = unsafe { networks.assume_init() };
+        let total_networks = unsafe { total_networks.assume_init() };
+
+        let networks = if total_networks > 0 {
+            // Safety: `networks` is malloced in application memory with size = `total_networks`
+            unsafe { Vec::from_raw_parts(networks, total_networks, total_networks) }
+                .into_iter()
+                .map(<_ as Into<NetworkScanInfo>>::into)
```
ian-h-chamberlain: ?

```suggestion
                .map(NetworkScanInfo::from)
```

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,1273 @@
...
+    /// uds.connect_network(&networks[0], b"udsdemo passphrase c186093cd2652741\0", ConnectionType::Client, 1)?;
+    /// #
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[doc(alias = "udsConnectNetwork")]
+    pub fn connect_network(
+        &mut self,
+        network: &NetworkScanInfo,
+        passphrase: &[u8],
```
ian-h-chamberlain, line 686: Does this always need to be nul-terminated? I Wonder if `&CStr` makes sense here instead of `&[u8]` but not positive what kinds of values might be passed in

ctru-rs/src/services/uds.rs
ian-h-chamberlain: I think I found some details, seems like it's more of a weird enum with sparse value choices? https://www.3dbrew.org/wiki/NWMUDS:GetConnectionStatus#Status_values

ctru-rs/src/services/uds.rs
```diff
@@ -0,0 +1,1273 @@
...
+            )
+        });
+
+        if code.0
+            != ctru_sys::MAKERESULT(
+                ctru_sys::RL_STATUS as _,
+                ctru_sys::RS_OUTOFRESOURCE as _,
+                ctru_sys::RM_UDS as _,
+                ctru_sys::RD_BUSY as _,
+            )
```
ian-h-chamberlain, line 964: What does this status represent? it still seems like it would be an error of some kind but we
return `Ok` for this?

ctru-rs/src/services/uds.rs
Jhynjhiruu: No, it can be whatever data you like. Games conventionally use a null-terminated string, but it's just raw bytes.

ctru-rs/src/services/uds.rs
Jhynjhiruu: See [libctru](https://github.com/devkitPro/libctru/blob/master/libctru/include/3ds/services/uds.h#L23) and [the UDS example](https://github.com/devkitPro/3ds-examples/blob/master/network/uds/source/uds.c#L286C7-L286C34). There's probably a nicer way to handle this, but this does work.

ctru-rs/src/services/uds.rs
Jhynjhiruu: Uuh, yeah, that'll work probably. Didn't know you could do that!

ctru-rs/src/services/uds.rs
Jhynjhiruu: There's no good reason (as far as I'm aware) to have the user able to change them, but they're properties of the data structure, rather than something you'd want to limit within a single function.

ctru-rs/src/services/uds.rs
Jhynjhiruu: `udsInit` copies the string (to be precise, it converts it to UTF-16), so I think it's safe.

ctru-rs/src/services/uds.rs
Jhynjhiruu: I'm not happy with most of the error handling in this module, so spending some time reworking it all would be useful I think.

ctru-rs/src/services/uds.rs
Jhynjhiruu: Is that likely to get done any time soon? If so, I'll hold off from fixing this until it's merged.

ctru-rs/src/services/uds.rs
Jhynjhiruu: Could do, but then all the accesses to the fields would have to be through methods. I've forgotten why I didn't do it that way originally at this point.

ctru-rs/src/services/uds.rs
Jhynjhiruu: Yeah, that's probably better than what I've done.

ctru-rs/examples/local-networking.rs
ian-h-chamberlain, line 1: Wow, this is a pretty intense example but it seems necessary given how many features there are in the `uds` service. I was wondering if some of this functionality could go onto `Uds` itself instead of just in this example, but honestly on first read I don't see any obvious things that make sense, so 👍 good with me

ctru-rs/src/services/uds.rs
ian-h-chamberlain: ah gotcha, perfect

ctru-rs/src/services/uds.rs
ian-h-chamberlain: Probably not, I'll capture this in a separate issue and you can leave this as-is (edit: #166)

ctru-rs/src/services/uds.rs
ian-h-chamberlain: Ok, so this error code basically means a packet was still sent but some other error occurred, I guess? Seems fine

Jhynjhiruu: Yeah, something like that. Maybe it'd be a good idea to hardcode it as a constant somewhere rather than calling it in the function, though - `MAKERESULT` presumably isn't `const` so it'd have to be a magic number with some explanation.

ian-h-chamberlain: If it isn't `const`, we probably can/should make it `const`, same for any of those other helper functions that just do bit operations on `Result`

ctru-rs/examples/local-networking.rs
```diff
@@ -0,0 +1,301 @@
...
+    let gfx = Gfx::new().unwrap();
+    let console = Console::new(gfx.top_screen.borrow_mut());
+
+    println!("Local networking demo");
+
+    let mut uds = Uds::new(None).unwrap();
+
+    println!("UDS initialised");
+
+    enum State {
```
adryzz, line 43: This looks a lot like a manually implemented future state machine. could this have worked better with async?

Jhynjhiruu: Very possibly! I'm scared of `async`.

ctru-rs/src/services/uds.rs
Jhynjhiruu: Presumably, this would need to be changed in ctru_sys. Once somebody does that, I'll change this to be a little more elegant.

ctru-rs/examples/local-networking.rs
Jhynjhiruu: I've done some basic reading about futures, and I don't understand how this could be done with async. Could you explain how that would work? It would definitely be nice to clean up the code if it's possible.

ian-h-chamberlain (approved):

ctru-rs/src/services/uds.rs
Meziu: The `TODO` doc comment should be removed now that it's fixed.

ctru-rs/src/services/uds.rs
```diff
@@ -945,6 +1176,199 @@ impl Uds {
...
+    /// let mut uds = Uds::new(None)?;
+    ///
+    /// uds.create_network(b"HBW\x10", None, None, b"udsdemo passphrase c186093cd2652741\0", 1)?;
+    /// let node_info = uds.get_node_info(NodeID::Node(2))?;
+    /// #
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[doc(alias = "udsGetNodeInformation")]
+    pub fn get_node_info(&self, address: NodeID) -> Result<NodeInfo, Error> {
```
Meziu: Nit: remove the `get_*` prefix.

Meziu (approved):

PR #157, opened 2024-02-12
Get software keyboard input without max byte size parameter

FenrirWolf:
Ever since I first started messing with the software keyboard from Rust, I've always *hated* that libctru's `swkbd` API forces you to use a fixed-size buffer in order to receive string data from the keyboard. We've worked around that restriction by just forcing the user to specify a maximum size for the intermediate text buffer, but I've always hated that solution too. And so I randomly decided it's finally time to do something about it.

However, it turns out that "doing something about it" involves reimplementing both [swkbdInputText](https://github.com/devkitPro/libctru/blob/master/libctru/source/applets/swkbd.c#L170-L272) and [swkbdMessageCallback](https://github.com/devkitPro/libctru/blob/master/libctru/source/applets/swkbd.c#L144-L168) in Rust and handling a whole bunch of gnarly unsafe code ourselves instead of just leaving all of that nonsense to libctru.

Is it worth introducing this much unsafe code just to solve a small API nitpick? Frankly I'm not entirely sure, but I thought I'd throw this out there and see what you guys think.

### DIFF
--- a/ctru-rs/src/applets/swkbd.rs
+++ b/ctru-rs/src/applets/swkbd.rs
@@ -5,10 +5,10 @@
 
 use crate::services::{apt::Apt, gfx::Gfx};
 use ctru_sys::{
-    self, aptLaunchLibraryApplet, aptSetMessageCallback, envGetAptAppId, svcBreak, svcCloseHandle,
+    self, aptLaunchLibraryApplet, aptSetMessageCallback, envGetAptAppId, svcCloseHandle,
     svcCreateMemoryBlock, APT_SendParameter, SwkbdButton, SwkbdDictWord, SwkbdExtra,
     SwkbdLearningData, SwkbdState, SwkbdStatusData, APPID_SOFTWARE_KEYBOARD, APTCMD_MESSAGE,
-    NS_APPID, SWKBD_CALLBACK_OK, USERBREAK_PANIC,
+    NS_APPID, SWKBD_CALLBACK_OK,
 };
 
 use bitflags::bitflags;
@@ -281,7 +281,7 @@ impl SoftwareKeyboard {
                 (self as *mut Self).cast(),
             );
 
-            match Self::swkbd_input_text(self.state.as_mut(), output.as_mut_vec()) {
+            match Self::swkbd_input_text(self.state.as_mut(), &mut output) {
                 ctru_sys::SWKBD_BUTTON_NONE => Err(self.state.result.into()),
                 ctru_sys::SWKBD_BUTTON_LEFT => Ok((output, Button::Left)),
                 ctru_sys::SWKBD_BUTTON_MIDDLE => Ok((output, Button::Middle)),
@@ -631,7 +631,7 @@ impl SoftwareKeyboard {
     // A reimplementation of `swkbdInputText` from `libctru/source/applets/swkbd.c`. Allows us to
     // get text from the software keyboard and put it directly into a `String` without requiring
     // an intermediate fixed-size buffer
-    fn swkbd_input_text(swkbd: &mut SwkbdState, buf: &mut Vec<u8>) -> SwkbdButton {
+    fn swkbd_input_text(swkbd: &mut SwkbdState, output: &mut String) -> SwkbdButton {
         use ctru_sys::{
             MEMPERM_READ, MEMPERM_WRITE, R_FAILED, SWKBD_BUTTON_LEFT, SWKBD_BUTTON_MIDDLE,
             SWKBD_BUTTON_NONE, SWKBD_BUTTON_RIGHT, SWKBD_D0_CLICK, SWKBD_D1_CLICK0,
@@ -715,11 +715,11 @@ impl SoftwareKeyboard {
             swkbd.initial_text_offset = 0;
 
             unsafe {
-                let utf16_iter = CStr::from_ptr(extra.initial_text)
-                    .to_str()
-                    .unwrap()
-                    .encode_utf16()
-                    .chain(once(0));
+                let utf16_iter =
+                    str::from_utf8_unchecked(CStr::from_ptr(extra.initial_text).to_bytes())
+                        .encode_utf16()
+                        .take(swkbd.max_text_len as _)
+                        .chain(once(0));
 
                 let mut initial_text_cursor = SWKBD_SHARED_MEM.cast::<u16>();
 
@@ -803,17 +803,16 @@ impl SoftwareKeyboard {
 
         let text16 = if swkbd.text_length > 0 {
             unsafe {
-                widestring::Utf16Str::from_slice(std::slice::from_raw_parts(
+                widestring::Utf16Str::from_slice_unchecked(std::slice::from_raw_parts(
                     SWKBD_SHARED_MEM.add(swkbd.text_offset as _).cast::<u16>(),
                     swkbd.text_length as usize,
                 ))
-                .unwrap()
             }
         } else {
             widestring::utf16str!("")
         };
 
-        buf.extend(text16.encode_utf8());
+        *output = text16.to_string();
 
         if swkbd.save_state_flags & (1 << 0) != 0 {
             unsafe {
@@ -860,20 +859,13 @@ impl SoftwareKeyboard {
         }
 
         let text16 = unsafe {
-            widestring::Utf16Str::from_slice(std::slice::from_raw_parts(
+            widestring::Utf16Str::from_slice_unchecked(std::slice::from_raw_parts(
                 SWKBD_SHARED_MEM.add(swkbd.text_offset as _).cast::<u16>(),
                 swkbd.text_length as usize + 1,
             ))
-            .unwrap()
         };
 
-        let text8 = match String::from_utf8(text16.encode_utf8().collect()) {
-            Ok(s) => s,
-            Err(_) => {
-                svcBreak(USERBREAK_PANIC);
-                unreachable!();
-            }
-        };
+        let text8 = text16.to_string();
 
         let mut retmsg = std::ptr::null();
 
--- a/ctru-rs/src/applets/swkbd.rs
+++ b/ctru-rs/src/applets/swkbd.rs
@@ -5,14 +5,13 @@
 
 use crate::services::{apt::Apt, gfx::Gfx};
 use ctru_sys::{
-    self, aptLaunchLibraryApplet, aptSetMessageCallback, envGetAptAppId, svcCloseHandle,
+    aptLaunchLibraryApplet, aptSetMessageCallback, envGetAptAppId, svcCloseHandle,
     svcCreateMemoryBlock, APT_SendParameter, SwkbdButton, SwkbdDictWord, SwkbdExtra,
     SwkbdLearningData, SwkbdState, SwkbdStatusData, APPID_SOFTWARE_KEYBOARD, APTCMD_MESSAGE,
     NS_APPID,
 };
 
 use bitflags::bitflags;
-use libc;
 
 use std::ffi::{CStr, CString};
 use std::fmt::Display;
--- a/ctru-rs/src/applets/swkbd.rs
+++ b/ctru-rs/src/applets/swkbd.rs
@@ -809,12 +809,12 @@ impl SoftwareKeyboard {
                     SWKBD_SHARED_MEM
                         .add((*swkbd).text_offset as _)
                         .cast::<u16>(),
-                    (*swkbd).text_length as usize + 1,
+                    (*swkbd).text_length as usize,
                 ))
                 .unwrap()
             }
         } else {
-            widestring::utf16str!("\0")
+            widestring::utf16str!("")
         };
 
         buf.extend(text16.encode_utf8());
--- a/ctru-rs/src/applets/swkbd.rs
+++ b/ctru-rs/src/applets/swkbd.rs
@@ -21,15 +21,12 @@ use std::str;
 
 type CallbackFunction = dyn Fn(&CStr) -> (CallbackResult, Option<CString>);
 
-// I hate that we have to use this, but sometimes you gotta smuggle pointers into C callbacks
-// and that's just how things are
-static mut SWKBD_SHARED_MEM: *mut libc::c_void = std::ptr::null_mut();
-
 /// Configuration structure to setup the Software Keyboard applet.
 #[doc(alias = "SwkbdState")]
 pub struct SoftwareKeyboard {
     state: Box<SwkbdState>,
     callback: Option<Box<CallbackFunction>>,
+    callback_data: MessageCallbackData,
     error_message: Option<CString>,
     initial_text: Option<CString>,
 }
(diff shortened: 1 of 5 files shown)


Meziu:
Wow! This re-implementation looks quite nice. I must say I'm not against the "amounts" of unsafe code present here, since it manages to strengthen the API quite a lot and give us more control over the internal implementation (as I said in the past, especially regarding `ndsp`, there is a lot of things we could improve by getting our hands a bit dirty).

I'll try a deep dive review once I manage to get some time.

FenrirWolf:
Okay, I think this PR is in a pretty good state now and I probably won't need to push any more changes. I'm not sure what the nightly test failures are about though. They don't seem related to anything actually going on in the code here.

ctru-rs/src/applets/swkbd.rs
```diff
@@ -207,6 +213,20 @@ bitflags! {
     }
 }
 
+struct MessageCallbackData {
+    extra: *mut SwkbdExtra,
+    swkbd_shared_mem_ptr: *mut libc::c_void,
+}
```
ian-h-chamberlain, line 219: Is this something that we might eventually be able to generalize to a `MessageCallbackData<T>` where `T = SwkbdExtra` in this case? I know some people have been interested in general-purpose `svc` helpers to work with system calls that don't have "nice" APIs in `libctru`, so just wondering how generic this code is.

Definitely not asking for something like that in this PR, but just wondering if this PR might provide a good starting point from which that could be built. I believe some of the `ir_user` module by @AzureMarker also did a bit of `svc` memory allocation and stuff but I am not familiar with the specifics so not sure if there is any overlap there.

ctru-rs/src/applets/swkbd.rs
```diff
@@ -675,6 +639,292 @@ impl SoftwareKeyboard {
...
+        };
+
+        let swkbd = self.state.as_mut();
+        let mut extra = unsafe { swkbd.__bindgen_anon_1.extra };
+
+        // Calculate shared mem size
+        let mut shared_mem_size = 0;
+
+        shared_mem_size +=
+            (std::mem::size_of::<u16>() * (swkbd.max_text_len as usize + 1) + 3) & !3;
```
ian-h-chamberlain: I assume the `+ 3) & !3` is an alignment thing (align to `usize`, maybe)? does it make sense to have something like this just to make the math more obvious?

```rs
let max_len = swkbd.max_text_len as usize + 1;

shared_mem_size += (size_of::<u16>() * max_len).next_multiple_of(size_of::<usize>());

// etc.
```

ctru-rs/src/applets/swkbd.rs
```diff
@@ -675,6 +640,292 @@ impl SoftwareKeyboard {
...
+
+        if swkbd.save_state_flags & (1 << 0) != 0 {
+            swkbd.status_offset = shared_mem_size as _;
+            shared_mem_size += std::mem::size_of::<SwkbdStatusData>();
+        }
+
+        if swkbd.save_state_flags & (1 << 1) != 0 {
+            swkbd.learning_offset = shared_mem_size as _;
+            shared_mem_size += std::mem::size_of::<SwkbdLearningData>();
+        }
```
ian-h-chamberlain, line 682: It would be nice if we could use `bitflags` for this but I think this struct is defined by `ctru_sys` right? so maybe there's no great option but maybe defining consts for the bitshifts would be useful

ctru-rs/src/applets/swkbd.rs
```diff
@@ -675,6 +640,292 @@ impl SoftwareKeyboard {
...
+
+        shared_mem_size = (shared_mem_size + 0xFFF) & !0xFFF;
```
ian-h-chamberlain: similar comment, could this be

```rs
swkbd.shared_memory_size = shared_mem_size.next_multiple_of(0x1000);
```

ctru-rs/src/applets/swkbd.rs
```diff
@@ -675,6 +640,292 @@ impl SoftwareKeyboard {
...
+
+        swkbd.shared_memory_size = shared_mem_size;
+
+        // Allocate shared mem
+        let swkbd_shared_mem_ptr = unsafe { libc::memalign(0x1000, shared_mem_size) };
```
ian-h-chamberlain, line 689: huh, I was thinking it would be nice to use Rust `alloc` for this but I guess it's passed to C immediately anyway so there's not really a compelling reason to do that that I can think of.

ctru-rs/src/applets/swkbd.rs
```diff
@@ -675,6 +640,292 @@ impl SoftwareKeyboard {
...
+                shared_mem_size as _,
+                MEMPERM_READ | MEMPERM_WRITE,
+                MEMPERM_READ | MEMPERM_WRITE,
+            )
+        };
+
+        if R_FAILED(res) {
+            unsafe {
+                libc::free(swkbd_shared_mem_ptr);
+                swkbd.result = SWKBD_OUTOFMEM;
```
ian-h-chamberlain, line 711: Does this really indicate `OUTOFMEM` ? Or is there any way we could pass the info from `res` back to the `swkbd.result`... I'm not sure what possible return values there are here but I assume if it's already allocated an error would indicate something else like bad permissions or whatever (e.g. the example in https://www.3dbrew.org/wiki/SVC#Kernel_error-codes)

ctru-rs/src/applets/swkbd.rs
```diff
@@ -675,6 +640,292 @@ impl SoftwareKeyboard {
...
+            }
+        }
+
+        // Copy stuff to shared mem
+        if !extra.initial_text.is_null() {
+            swkbd.initial_text_offset = 0;
+
+            unsafe {
+                let utf16_iter =
+                    str::from_utf8_unchecked(CStr::from_ptr(extra.initial_text).to_bytes())
```
ian-h-chamberlain, line 722: just for clarity, this is safe to do because we check UTF-8 validity when the initial text is first set (from #158) ? Might be good to include a comment indicating that

ctru-rs/src/applets/swkbd.rs
```diff
@@ -675,6 +640,292 @@ impl SoftwareKeyboard {
...
+                        .encode_utf16()
+                        .take(swkbd.max_text_len as _)
+                        .chain(once(0));
```
ian-h-chamberlain, line 725: I was going to suggest
```suggestion
                        .chain(once(b'\0'));
```

but I guess this is actually a `u16` so that would need `.into()` and I'm not sure it's really better. Could go either way

ctru-rs/src/applets/swkbd.rs
```diff
@@ -675,6 +640,292 @@ impl SoftwareKeyboard {
...
+
+                let mut initial_text_cursor = swkbd_shared_mem_ptr.cast();
+
+                for code_point in utf16_iter {
+                    *initial_text_cursor = code_point;
+                    initial_text_cursor = initial_text_cursor.add(1);
```
ian-h-chamberlain, line 731: Just for my understanding, is this equivalent to e.g.
```rs
let utf16: Vec<u16> = utf16_iter.collect();
std::ptr::copy(utf16.as_ptr(), initial_text_cursor, utf16.len());
```
other than the extra `Vec` allocation? I guess there's no equivalent `ptr::write(impl Iterator)` or something like that

ctru-rs/src/applets/swkbd.rs
```diff
@@ -675,6 +640,292 @@ impl SoftwareKeyboard {
...
+
+                aptSetMessageCallback(
+                    Some(Self::swkbd_message_callback),
+                    std::ptr::addr_of_mut!(self.callback_data).cast(),
+                );
+            }
+
+            aptLaunchLibraryApplet(
+                APPID_SOFTWARE_KEYBOARD,
+                swkbd as *mut _ as *mut _,
```
ian-h-chamberlain: can at least one of these be a `cast()`/`cast_mut()`? Presumable you can `as *mut T` and then cast to `c_void`?

ctru-rs/src/applets/swkbd.rs
```diff
@@ -675,6 +640,292 @@ impl SoftwareKeyboard {
...
+        let button = match swkbd.result {
+            SWKBD_D1_CLICK0 | SWKBD_D2_CLICK0 => SWKBD_BUTTON_LEFT,
+            SWKBD_D2_CLICK1 => SWKBD_BUTTON_MIDDLE,
+            SWKBD_D0_CLICK | SWKBD_D1_CLICK1 | SWKBD_D2_CLICK2 => SWKBD_BUTTON_RIGHT,
+            _ => SWKBD_BUTTON_NONE,
+        };
+
+        if swkbd.text_length > 0 {
+            let text16 = unsafe {
+                widestring::Utf16Str::from_slice_unchecked(std::slice::from_raw_parts(
```
ian-h-chamberlain, line 814: I assume the `_unchecked` here means that we are trusting either the OS or `libctru` to give us a valid string, right? Just wondering if an extra check here might be worth a little bit of safety

ctru-rs/src/applets/swkbd.rs
```diff
@@ -675,6 +640,292 @@ impl SoftwareKeyboard {
...
+            };
+        }
+
+        if swkbd.save_state_flags & (1 << 1) != 0 {
+            unsafe {
+                std::ptr::copy_nonoverlapping(
+                    swkbd_shared_mem_ptr.add(swkbd.learning_offset as _).cast(),
+                    extra.learning_data,
+                    1,
+                )
```
ian-h-chamberlain, line 839: nit: might increase readibility here to use the method form:
```rs
extra.learning_data.copy_from_nonoverlapping(
	swkbd_shared_mem_ptr.add(swkbd.learning_offset as _).cast(),
	1,
);
```

ctru-rs/src/applets/swkbd.rs
```diff
@@ -675,6 +640,292 @@ impl SoftwareKeyboard {
...
+            };
+        }
+
+        unsafe { libc::free(swkbd_shared_mem_ptr) };
```
ian-h-chamberlain, line 843: Allocating and freeing all the options every time is probably fine, and I don't think we need to optimize right now — but I wonder if we could also have more of a `Builder` API pattern that sets all the sizing options up front and allocates just once, which would probably make this call cheaper.
Don't want to prematurely optimize though since swkbd isn't likely to be a hot path for most apps (I hope, lol!)

ctru-rs/src/applets/swkbd.rs
```diff
@@ -675,6 +640,292 @@ impl SoftwareKeyboard {
...
+
+        button
+    }
+
+    // A reimplementation of `swkbdMessageCallback` from `libctru/source/applets/swkbd.c`.
+    // This is only needed because the original function is private to libctru, so we can't
+    // simply reuse their version
+    #[deny(unsafe_op_in_unsafe_fn)]
```
ian-h-chamberlain, line 850: ❤️❤️❤️

I have sometimes considered if we should consider denying as a library lint but I think it would be a nontrivial amount of work so never bothered to try

ian-h-chamberlain (approved):
Mostly LGTM, had some small comments / questions about some of the implementation but a lot of it is just musing about future possibilities and I don't think anything blocking.

IMO probably would be fine to merge as-is with one more approval

ctru-rs/src/applets/swkbd.rs
FenrirWolf: I only made this struct because `libctru` doesn't include a pointer to the shared memory block in `SwkbdExtra` like it ought to, and instead uses a `static` to pass the pointer into the APT message callback. I decided there's no reason we have to do the same in our implementation.

I've never messed with any other APT message calls, so I'm not sure if this specific struct would be helpful outside of this particular context.

ctru-rs/src/applets/swkbd.rs
FenrirWolf: Sure, that might make more sense. I'm honestly not very fluent in bitwise math, so I just kinda translated the C code directly without getting too fancy about interpreting its meaning.

ctru-rs/src/applets/swkbd.rs
FenrirWolf: I did think about using `std::alloc`, but as you said the memory block is only being manipulated by C code anyway and so it's more straightforward to just use `memalign` and `free`

ctru-rs/src/applets/swkbd.rs
FenrirWolf: Yep, we know it's valid UTF-8 since the user is only able to supply a `&str` to create the input from

ctru-rs/src/applets/swkbd.rs
FenrirWolf: Yep, this code is just copying each UTF-16 code unit into the shared memory block.

ctru-rs/src/applets/swkbd.rs
FenrirWolf: Yeah, we're trusting the Software Keyboard to give us proper UTF-16 text here and I don't think there's any reason to suspect it would fail at that job

ctru-rs/src/applets/swkbd.rs
FenrirWolf: Yeah, calling the software keyboard involves a lot of setup no matter how you slice it. I'm not sure if it would be worth trying to reduce allocations here, especially since any changes to the software keyboard config would require recomputing and reallocating the memory block anyway.

ctru-rs/src/applets/swkbd.rs
FenrirWolf: I return `OUTOFMEM` there because [libctru does the same](https://github.com/devkitPro/libctru/blob/master/libctru/source/applets/swkbd.c#L204-L211). And since `swkbd.result` can only contain [these values](https://libctru.devkitpro.org/swkbd_8h.html#aa02a4a80747eb05ddb85235a0a41ee93), there's not really any other choice to go with unless we feel like stashing extra values in unused enum variants, which imo would be both overkill and a bad idea in general.

I think this is a case where we just need to trust that the 3DS system calls are going to do the Right Thing™ given the right inputs.

FenrirWolf:
Also while we're messing around in here, what do you guys think about renaming `SoftwareKeyboard::get_string` to `SoftwareKeyboard::launch`? I only chose the former name back when there were 2 functions for getting text data from the keyboard and I wanted to make sure there was a strong distinction between the two, but since this PR does away with the `write_exact` method, I feel like that would be a sensible change

ian-h-chamberlain:
> Also while we're messing around in here, what do you guys think about renaming `SoftwareKeyboard::get_string` to `SoftwareKeyboard::launch`? 

Works for me

ian-h-chamberlain (approved):

ctru-rs/src/applets/swkbd.rs
Meziu: I am in favour of this as well! Unsafe blocks should be marked only when necessary and not all unsafe functions use unsafe code (especially around here...).

Meziu (approved):
LGTM. The only things I'd like to see improved are the alignment calculations that @ian-h-chamberlain pointed out. Other than that, it's fine.

FenrirWolf:
Apparently `next_multiple_of` is only stable as of rust 1.73 and `nightly-2023-06-01` is older than that. Everything works on my end though.

PR #159, opened 2024-02-15
Fix console API issues

FenrirWolf:
This PR consists of 2 parts. The first change is to use `addr_of_mut!` when getting pointers to the `EMPTY_CONSOLE`, which fixes some warnings about mutable references to `static mut` objects on recent nightlies.

The second change is one I'm making based more on "vibes" than anything, but I've long felt suspicious about the soundness of our `Console` API when it comes to taking references and pointers to the internal `PrintConsole`. For example, we have a number of console APIs that take `&self` even though there are underlying mutations in the `PrintConsole` struct every time a character is written to the screen.

I've never actually run into any problems or strange behavior when dealing with the console, but I suspect that the "proper" way to handle things is to wrap the `PrintConsole` in an `UnsafeCell` so that the compiler won't make assumptions about having unique access to the console state unless we specifically tell it to do so, and so I've made that change in this PR too.

It's possible that this change might be overly paranoid, but I'm not sure there's any harm in being paranoid in this particular way either. Either way, feel free to let me know what you guys think about the issue too.

### DIFF
--- a/ctru-rs/src/console.rs
+++ b/ctru-rs/src/console.rs
@@ -143,7 +143,7 @@ impl<'screen> Console<'screen> {
     /// ```
     pub fn exists() -> bool {
         unsafe {
-            let current_console = ctru_sys::consoleSelect(&mut EMPTY_CONSOLE);
+            let current_console = ctru_sys::consoleSelect(std::ptr::addr_of_mut!(EMPTY_CONSOLE));
 
             let res = (*current_console).consoleInitialised;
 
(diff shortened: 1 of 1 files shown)


Meziu (approved):
It's not overly-pessimistic to assume the compiler doesn't really understand what is going on with `PrintConsole`. Due to its weird nature as a globally shared struct, we probably got lucky to have never experienced errors before.

This can be merged right away.

PR #161, opened 2024-02-17
Implement low hanging apt sleep/homemenu functions and chainloader

adryzz:
Saw that these weren't added yet, although they're very simple.

## TLDR, Implemented
- [x] `aptSetSleepAllowed`
- [x] `aptIsSleepAllowed`
- [x] `aptSetHomeAllowed`
- [x] `aptIsHomeAllowed`
- [x] `aptJumpToHomeMenu`
- [x] chainloader stuff
- [ ] applist stuff

### DIFF
--- a/ctru-rs/src/services/apt.rs
+++ b/ctru-rs/src/services/apt.rs
@@ -98,9 +98,7 @@ impl Apt {
     /// You can set whether the console is allowed to sleep with [Apt::set_sleep_allowed].
     #[doc(alias = "aptIsSleepAllowed")]
     pub fn is_sleep_allowed(&mut self) -> bool {
-        unsafe {
-            ctru_sys::aptIsSleepAllowed()
-        }
+        unsafe { ctru_sys::aptIsSleepAllowed() }
     }
 
     /// Set if the console is allowed to enter the home menu.
@@ -114,21 +112,17 @@ impl Apt {
     }
 
     /// Check if the console is allowed to enter the home menu.
-    /// 
+    ///
     /// You can set whether the console is allowed to enter the home menu with [Apt::set_home_allowed].
     #[doc(alias = "aptIsHomeAllowed")]
     pub fn is_home_allowed(&mut self) -> bool {
-        unsafe {
-            ctru_sys::aptIsHomeAllowed()
-        }
+        unsafe { ctru_sys::aptIsHomeAllowed() }
     }
 
     /// Immediately jumps to the home menu.
     #[doc(alias = "aptIsHomeAllowed")]
     pub fn jump_to_home_menu(&mut self) {
-        unsafe {
-            ctru_sys::aptJumpToHomeMenu()
-        }
+        unsafe { ctru_sys::aptJumpToHomeMenu() }
     }
 }
 
--- a/ctru-rs/src/services/apt.rs
+++ b/ctru-rs/src/services/apt.rs
@@ -120,7 +120,7 @@ impl Apt {
     }
 
     /// Immediately jumps to the home menu.
-    #[doc(alias = "aptIsHomeAllowed")]
+    #[doc(alias = "aptJumpToHomeMenu")]
     pub fn jump_to_home_menu(&mut self) {
         unsafe { ctru_sys::aptJumpToHomeMenu() }
     }
@@ -159,19 +159,19 @@ impl<'a> Chainloader<'a> {
 
     /// Configures the chainloader to launch a specific application.
     #[doc(alias = "aptSetChainloader")]
-    pub fn set_chainloader(&mut self, title: &super::am::Title<'_>) {
+    pub fn set(&mut self, title: &super::am::Title<'_>) {
         unsafe { ctru_sys::aptSetChainloader(title.id(), title.media_type() as u8) }
     }
 
     /// Configures the chainloader to launch the previous application.
     #[doc(alias = "aptSetChainloaderToCaller")]
-    pub fn set_chainloader_to_caller(&mut self) {
+    pub fn set_to_caller(&mut self) {
         unsafe { ctru_sys::aptSetChainloaderToCaller() }
     }
 
     /// Configures the chainloader to relaunch the current application (i.e. soft-reset)
     #[doc(alias = "aptSetChainloaderToSelf")]
-    pub fn set_chainloader_to_self(&mut self) {
+    pub fn set_to_self(&mut self) {
         unsafe { ctru_sys::aptSetChainloaderToSelf() }
     }
 }
--- a/ctru-rs/src/services/apt.rs
+++ b/ctru-rs/src/services/apt.rs
@@ -82,6 +82,54 @@ impl Apt {
             Ok(())
         }
     }
+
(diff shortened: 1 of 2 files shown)


adryzz:
oh, looks like i made my PR on the wrong branch

edit: fixed

FenrirWolf (approved):

Meziu:
> Saw that these weren't added yet, although they're very simple.

It’s a pretty straightforward addition. Just one question, what does “sleep allowed” actually do? Can the app stop the system to sleep even if the shell is closed?

> All the chainloader stuff is still not implemented, as that probably requires a big abstraction layer to list apps and other things.

Chainloader and applist stuff could be implemented with the already existing `AM` API (perhaps?).

adryzz:
>  It’s a pretty straightforward addition. Just one question, what does “sleep allowed” actually do? Can the app stop the system to sleep even if the shell is closed?

It's used in online multiplayer games and stuff like that. i've seen the no-sleep/no-home combo for example in Steel Diver: Sub Wars when playing online.

essentially, if you close the shell (or use the toggle switch on the O2DS), the displays turn off, but the console does not go to sleep. you can see this because the blue LED does not oscillate in brightness when it's not sleeping.

I should probably write better documentation, since it wasn't clear at first glance what it did.

>  Chainloader and applist stuff could be implemented with the already existing AM API (perhaps?).

edit: turns out i didn't know the existing AM API provided that, i have made a chainloader prototype, although i don't really like it as-is

adryzz:
Alright, implemented a chainloader API, the problem is that the bindings are missing a function.

https://github.com/devkitPro/libctru/blob/d0936b879bd2088cae61b69707fe70cb66ede651/libctru/source/services/apt.c#L111

FenrirWolf:
`static` functions like that one aren't exported outside of their compilation unit. In other words, they're private to libctru and you have to reimplement them if you want to use them. Fortunately the impl is pretty simple for that function and the other ones like it, so hopefully that won't be too much of a problem.

ctru-rs/src/services/apt.rs
```diff
@@ -82,6 +82,48 @@ impl Apt {
...
+    /// Check if the console is allowed to enter the home menu.
+    ///
+    /// You can set whether the console is allowed to enter the home menu with [Apt::set_home_allowed].
+    #[doc(alias = "aptIsHomeAllowed")]
+    pub fn is_home_allowed(&self) -> bool {
+        unsafe { ctru_sys::aptIsHomeAllowed() }
+    }
+
+    /// Immediately jumps to the home menu.
+    #[doc(alias = "aptIsHomeAllowed")]
```
ian-h-chamberlain: This alias looks like maybe a copy-paste mistake?

ctru-rs/src/services/apt.rs
```diff
@@ -90,3 +132,46 @@ impl Drop for Apt {
...
+
+    /// Clears the chainloader state.
+    #[doc(alias = "aptClearChainloader")]
+    pub fn clear(&mut self) {
+        unsafe { ctru_sys::aptClearChainloader() }
+    }
+
+    /// Configures the chainloader to launch a specific application.
+    #[doc(alias = "aptSetChainloader")]
+    pub fn set_chainloader(&mut self, title: &super::am::Title<'_>) {
```
ian-h-chamberlain: nit: Since these are methods on `Chainloader`, I'd probably drop `chainloader` from the method names, maybe something like:
- `Chainloader::set(title)`
- `Chainloader::set_caller()`
- `Chainloader::set_self()`

ian-h-chamberlain (commented):
LGTM just a couple small tweaks I'd recommend

ctru-rs/src/services/apt.rs
adryzz: oops

ctru-rs/src/services/apt.rs
```diff
@@ -90,3 +132,48 @@ impl Drop for Apt {
...
+
+    /// Clears the chainloader state.
+    #[doc(alias = "aptClearChainloader")]
+    pub fn clear(&mut self) {
+        unsafe { ctru_sys::aptClearChainloader() }
+    }
+
+    /// Configures the chainloader to launch a specific application.
+    ///
+    /// See also [super::am::Title]
```
Meziu: Nit: we usually use code blocks in doc links.
```suggestion
    /// See also [`Title`](crate::services::am::Title]
```

ctru-rs/src/services/apt.rs
```diff
@@ -90,3 +132,48 @@ impl Drop for Apt {
         unsafe { ctru_sys::aptExit() };
     }
 }
+
+/// Can launch other applications when the current one exits.
+pub struct Chainloader<'a> {
```
Meziu, line 137: `Chainloader` should clear on `Drop`, so that leftovers from previous usages won't influence in un-mutable environments.

ctru-rs/src/services/apt.rs
```diff
@@ -90,3 +132,48 @@ impl Drop for Apt {
...
+    _apt: &'a Apt,
+}
+
+impl<'a> Chainloader<'a> {
+    /// Gets a handle to the chainloader
+    pub fn new(apt: &'a Apt) -> Self {
```
Meziu, line 143: Maybe `Chainloader` should require a mutable reference to `Apt`, but that may not be necessary if *no* other aspect of `Apt` reads or uses data handled by `Chainloader`.

Tell me what you think.

Meziu (approved):
The changes are small and contained. LGTM

ctru-rs/src/services/apt.rs
adryzz: i did think about this, but then forgot to propose it.

the problem is that the destructor will run when you exit the app, as the chainloader goes out of scope.

ctru-rs/src/services/apt.rs
adryzz: i tried this, but then using it inside the main `apt` loop is a mess because it takes an immutable reference to loop

ctru-rs/src/services/apt.rs
Meziu: Ah right, didn’t think about it.

Wait, does this mean the setup only works because there is a reference to `Apt` that doesn’t get uninitialised? (The one made in libctru, i mean). If this were to not work without that (because the apt service were to be unused) then we *must* find a way around this.

ctru-rs/src/services/apt.rs
Meziu: Yeah I know, that’s exactly what I was thinking would happen. It’s not an issue as long as all of the chainloader code is only accessible from here.

ctru-rs/src/services/apt.rs
adryzz: at a first look it definitely looks that way.

ctru-rs/src/services/apt.rs
adryzz: these are the only chainloader functions, so i think it's fine to have it be this way

ian-h-chamberlain (approved):

PR #162, opened 2024-02-19
Add error applet support + error applet panic hook

FenrirWolf:
Adds support for the [error applet](https://libctru.devkitpro.org/error_8h.html), which is a simple applet that lets you display error text in a pop-up window on the bottom screen.

The applet is also capable of launching the EULA registration dialogue, as well as "infrastructure communications-related error messages" via error codes of a completely unspecified nature, but I've opted not to support either of these functions. The applet also supports changing the language of the "An error has occurred" header text at the top of the applet, but I don't think it's worth including that functionality either since it defaults to your system's preferred language anyway.

The real fun part about this PR was discovering that Bindgen was generating the `errorConf` struct with the wrong size and field offsets because the struct contains enum values that were themselves generated with the wrong sizes. Hopefully that's a problem we won't run into very often again, but it's one that's worth watching out for in the future.

### DIFF
--- a/ctru-rs/src/applets/error.rs
+++ b/ctru-rs/src/applets/error.rs
@@ -88,12 +88,16 @@ impl PopUp {
     }
 }
 
-/// Sets a custom panic hook that uses the error applet to display panic messages. You can also choose to have the
-/// previously registered panic hook called along with the error applet message, which can be useful if you want
-/// to use input redirection to display panic messages over `3dslink` or `GDB`.
+/// Sets a custom [panic hook](https://doc.rust-lang.org/std/panic/fn.set_hook.html) that uses the error applet to display panic messages.
 ///
-/// If the `Gfx` service is not initialized during a panic, the error applet will not be displayed and the old
+/// You can also choose to have the previously registered panic hook called along with the error applet message, which can be useful
+/// if you want to use output redirection to display panic messages over `3dslink` or `GDB`.
+///
+/// If the [`Gfx`] service is not initialized during a panic, the error applet will not be displayed and the old
 /// panic hook will be called.
+///
+/// You can use [`std::panic::take_hook`](https://doc.rust-lang.org/std/panic/fn.take_hook.html) to unregister the panic hook
+/// set by this function.
 pub fn set_panic_hook(call_old_hook: bool) {
     use crate::services::gfx::GFX_ACTIVE;
     use std::sync::TryLockError;
--- a/ctru-rs/src/applets/error.rs
+++ b/ctru-rs/src/applets/error.rs
@@ -1,6 +1,7 @@
 //! Error applet
 //!
 //! This applet displays error text as a pop-up message on the lower screen.
+
 use crate::services::{apt::Apt, gfx::Gfx};
 
 use ctru_sys::errorConf;
@@ -66,7 +67,16 @@ impl PopUp {
     /// Launches the error applet.
     #[doc(alias = "errorDisp")]
     pub fn launch(&mut self, _apt: &Apt, _gfx: &Gfx) -> Result<(), Error> {
-        unsafe { ctru_sys::errorDisp(self.state.as_mut()) }
+        unsafe { self.launch_unchecked() }
+    }
+
+    /// Launches the error applet without requiring an [`Apt`] or [`Gfx`] handle.
+    ///
+    /// # Safety
+    ///
+    /// Causes undefined behavior if the aforementioned services are not actually active when the applet launches.
+    unsafe fn launch_unchecked(&mut self) -> Result<(), Error> {
+        unsafe { ctru_sys::errorDisp(self.state.as_mut()) };
 
         match self.state.returnCode {
             ctru_sys::ERROR_NONE | ctru_sys::ERROR_SUCCESS => Ok(()),
@@ -76,15 +86,6 @@ impl PopUp {
             ctru_sys::ERROR_SOFTWARE_RESET => Err(Error::ResetPressed),
             _ => Err(Error::Unknown),
         }
-    }
-
-    /// Launches the error applet without requiring an [`Apt`] or [`Gfx`] handle.
-    ///
-    /// # Safety
-    ///
-    /// Causes undefined behavior if the aforementioned services are not actually active when the applet launches.
-    unsafe fn launch_unchecked(&mut self) {
-        unsafe { ctru_sys::errorDisp(self.state.as_mut()) };
     }
 }
 
@@ -123,7 +124,7 @@ pub fn set_panic_hook(call_old_hook: bool) {
             popup.set_text(&payload);
 
             unsafe {
-                popup.launch_unchecked();
+                let _ = popup.launch_unchecked();
             }
         } else {
             old_hook(panic_info);
--- a/ctru-rs/src/applets/error.rs
+++ b/ctru-rs/src/applets/error.rs
@@ -89,19 +89,16 @@ impl PopUp {
 }
 
 /// Sets a custom panic hook that uses the error applet to display panic messages. You can also choose to have the
-/// default panic hook called to print the message over stderr.
+/// previously registered panic hook called along with the error applet message, which can be useful if you want
+/// to use input redirection to display panic messages over `3dslink` or `GDB`.
 ///
-/// If the `Gfx` service is not initialized during a panic, the error applet will not be displayed and the default
-/// hook will be called.
-pub fn set_panic_hook(call_default_hook: bool) {
+/// If the `Gfx` service is not initialized during a panic, the error applet will not be displayed and the old
+/// panic hook will be called.
+pub fn set_panic_hook(call_old_hook: bool) {
     use crate::services::gfx::GFX_ACTIVE;
     use std::sync::TryLockError;
 
-    // Ensure we get the default hook instead of a previously registered user hook.
-    let default_hook = {
-        let _ = std::panic::take_hook();
-        std::panic::take_hook()
-    };
+    let old_hook = std::panic::take_hook();
 
     std::panic::set_hook(Box::new(move |panic_info| {
         let thread = std::thread::current();
@@ -111,8 +108,8 @@ pub fn set_panic_hook(call_default_hook: bool) {
         // If we get a `WouldBlock` error, we know that the `Gfx` service has been initialized.
         // Otherwise fallback to printing over stderr.
         if let (Err(TryLockError::WouldBlock), Ok(_apt)) = (GFX_ACTIVE.try_lock(), Apt::new()) {
-            if call_default_hook {
-                default_hook(panic_info);
+            if call_old_hook {
+                old_hook(panic_info);
             }
 
             let payload = format!("thread '{name}' {panic_info}");
@@ -125,7 +122,7 @@ pub fn set_panic_hook(call_default_hook: bool) {
                 popup.launch_unchecked();
             }
         } else {
-            default_hook(panic_info);
+            old_hook(panic_info);
         }
     }));
 }
--- a/ctru-rs/src/applets/error.rs
+++ b/ctru-rs/src/applets/error.rs
@@ -108,25 +108,22 @@ pub fn set_panic_hook(use_stderr: bool) {
 
         // If we get a `WouldBlock` error, we know that the `Gfx` service has been initialized.
         // Otherwise fallback to printing over stderr.
-        match GFX_ACTIVE.try_lock() {
-            Err(TryLockError::WouldBlock) => {
-                if use_stderr {
-                    print_to_stderr(name, panic_info);
-                }
-
-                let payload = format!("thread '{name}' {panic_info}");
-
-                let mut popup = PopUp::new(Kind::Top);
-
-                popup.set_text(&payload);
-
-                unsafe {
-                    popup.launch_unchecked();
-                }
-            }
-            _ => {
+        if let Err(TryLockError::WouldBlock) = GFX_ACTIVE.try_lock() {
+            if use_stderr {
                 print_to_stderr(name, panic_info);
             }
+
+            let payload = format!("thread '{name}' {panic_info}");
+
+            let mut popup = PopUp::new(Kind::Top);
+
+            popup.set_text(&payload);
+
+            unsafe {
+                popup.launch_unchecked();
+            }
+        } else {
+            print_to_stderr(name, panic_info);
         }
     }));
 
--- a/ctru-rs/src/applets/error.rs
+++ b/ctru-rs/src/applets/error.rs
@@ -100,15 +100,13 @@ pub fn set_panic_hook(use_stderr: bool) {
     use std::sync::TryLockError;
 
     std::panic::set_hook(Box::new(move |panic_info| {
-        let _apt = Apt::new();
-
         let thread = std::thread::current();
 
         let name = thread.name().unwrap_or("<unnamed>");
 
         // If we get a `WouldBlock` error, we know that the `Gfx` service has been initialized.
         // Otherwise fallback to printing over stderr.
-        if let Err(TryLockError::WouldBlock) = GFX_ACTIVE.try_lock() {
+        if let (Err(TryLockError::WouldBlock), Ok(_apt)) = (GFX_ACTIVE.try_lock(), Apt::new()) {
             if use_stderr {
                 print_to_stderr(name, panic_info);
             }
--- a/ctru-rs/src/applets/error.rs
+++ b/ctru-rs/src/applets/error.rs
@@ -22,7 +22,24 @@ pub enum Kind {
     Top = ctru_sys::ERROR_TEXT_WORD_WRAP,
 }
 
-impl ErrorApplet {
+/// Error returned by an unsuccessful [`PopUp::launch()`].
+#[doc(alias = "errorReturnCode")]
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[repr(i8)]
+pub enum Error {
+    /// Unknown error occurred.
+    Unknown = ctru_sys::ERROR_UNKNOWN,
+    /// Operation not supported.
+    NotSupported = ctru_sys::ERROR_NOT_SUPPORTED,
+    /// Home button pressed while [`PopUp`] was running.
+    HomePressed = ctru_sys::ERROR_HOME_BUTTON,
+    /// Power button pressed while [`PopUp`] was running.
+    PowerPressed = ctru_sys::ERROR_POWER_BUTTON,
+    /// Reset button pressed while [`PopUp`] was running.
+    ResetPressed = ctru_sys::ERROR_SOFTWARE_RESET,
+}
+
+impl PopUp {
     /// Initialize the error applet with the provided text kind.
     #[doc(alias = "errorInit")]
     pub fn new(kind: Kind) -> Self {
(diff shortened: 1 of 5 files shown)


adryzz:
this is super cool

ctru-rs/src/applets/error.rs
```diff
@@ -0,0 +1,56 @@
+//! Error applet
+//!
+//! This applet displays error text as a pop-up message on the lower screen.
+#![doc(alias = "Error")]
```
Meziu: Nit: I don't know if it makes any difference, but we usually write doc aliases in lowercase.
```suggestion
#![doc(alias = "error")]
```

ctru-rs/src/applets/error.rs
```diff
@@ -0,0 +1,56 @@
...
+
+use crate::services::{apt::Apt, gfx::Gfx};
+
+use ctru_sys::errorConf;
+
+/// Configuration struct to set up the Error applet.
+#[doc(alias = "errorConf")]
+pub struct ErrorApplet {
```
Meziu: We need a better name. How about `Display` or `View`? Using it with the module prefix (`error::View`) makes more sense than a redundant `ErrorApplet` :sweat_smile:.

Meziu (approved):
Great, simple and powerful. As with all `applets`, it depends on `Apt` and `Gfx`, but I wonder how cool a panic hook using this would be!

ctru-rs/src/applets/error.rs
FenrirWolf: I've also thought about `Applet`, but then that also sounds a bit too generic in its own right

FenrirWolf: Other options that come to mind are `Window` or `PopUp` since the applet looks like a pop-up window when it's displayed

adryzz:
> Great, simple and powerful. As with all applets, it depends on Apt and Gfx, but I wonder how cool a panic hook using this would be!

tried it a few days ago, it's super super cool, but yeah, without memory crimes you can't easily do such a thing unfortunately.

ctru-rs/src/applets/error.rs
Meziu: PopUp sounds better, and it’d help differentiate it from the other “error” modules in this crates

ctru-rs/src/applets/error.rs
FenrirWolf: Turns out there's no need for an alias since the module is already named `error` anyway

FenrirWolf:
> tried it a few days ago, it's super super cool, but yeah, without memory crimes you can't easily do such a thing unfortunately.

Setting up a custom panic hook isn't too bad actually. For example, I could easily add a method like this to the PR if we want:
```rust
/// Sets a custom panic hook that uses the error applet to display panic messages.
///
/// SAFETY: The error applet requires that both the [`Apt`] and [`Gfx`] services are active when it launches.
/// By calling this function, you promise that you will keep those services alive until either the program ends
/// or you manually unregister this hook with [`std::panic::take_hook`].
pub unsafe fn set_panic_hook() {
    std::panic::set_hook(Box::new(|panic_info| {
        let mut popup = PopUp::new(Kind::Top);

        let thread = std::thread::current();

        let name = thread.name().unwrap_or("<unnamed>");

        let payload = format!("thread '{name}' {panic_info}");

        popup.set_text(&payload);

        unsafe {
            ctru_sys::errorDisp(popup.state.as_mut());
        }
    }))
}
```

FenrirWolf:
Okay, we've got ourselves a panic hook now. It works pretty well!

2024-02-23_17-35-39 701_to 1

Meziu:
@AzureMarker I’d like to hear your opinion on this addition since it directly continues this [discussion](https://github.com/orgs/rust3ds/discussions/153).

With the use of the error applet, we can cut down the necessity of the `Console` functionality and just stand on `Gfx` and `Apt`. Still, the only way I can see this working is as an `unsafe` function, but it might still be a worthy alternative to `stderr` redirection with `3dslink`.

FenrirWolf:
The way I see it, providing this hook is an overall win because:

1) since `launch_unchecked` is private, the user cannot write their own equivalent to this code at all and they would have to start from scratch with raw `ctru-sys` calls. The error applet isn't very complicated to call, but it would still be a bit annoying for the user to create and audit multiple unsafe blocks to accomplish what we can provide here with just one, I guess we could instead expose `launch_unchecked`, but the only practical use for that fn being public would be to implement a hook like this one anyway.

2) the preconditions for `set_panic_hook` are pretty "easy" for most users to fulfill: just init `Apt` and `Gfx` at the top of your program and everything will Just Work. And coincidentally enough, that's what the vast majority of programs using `ctru-rs` are already doing anyway.

3) There might be room to further improve the safety situation. For example, the majority of `libctru` services are atomically reference-counted on each `init` and `exit`, meaning it should be safe to create and distribute multiple handles to those services, and some services like `Apt` are [pre-initialized in every libctru program](https://github.com/devkitPro/libctru/blob/master/libctru/source/system/appInit.c) even if the user never does it anyway. So technically it should be perfectly fine for us to just `Apt::new` right inside the panic hook to guarantee access to an `Apt` handle. I believe that `Gfx` is one of the few handles that doesn't internally reference count in this way, but I did notice that we have a `static GFX_ACTIVE` value that the panic hook could check against before doing its thing.

FenrirWolf:
I pushed a change that makes `set_panic_hook` a safe function. Are there any other sources of unsafety when calling the applet though, or is this change enough to cover all our bases?

I'm also not sure if this method of checking for `Gfx` access is good enough or if it's technically vulnerable to time-of-check-to-time-of-use problems. I might have to see if there's some way to smuggle a proper `Gfx` handle into the closure instead.

ctru-rs/src/applets/error.rs
```diff
@@ -0,0 +1,149 @@
...
+        } else {
+            print_to_stderr(name, panic_info);
+        }
+    }));
+
+    fn print_to_stderr(name: &str, panic_info: &std::panic::PanicInfo) {
+        let mut stderr = std::io::stderr().lock();
+
+        let _ = writeln!(stderr, "thread '{name}' {panic_info}");
+    }
```
ian-h-chamberlain: Instead of `use_stderr` could we provide an option to use the default hook or not? something like
```rs
pub fn set_panic_hook(call_existing_hook: bool) {
	let old_hook = std::panic::take_hook();

	std::panic::set_hook(Box::new(move |panic_info| {
		if let Err(TryLockError::WouldBlock) = GFX_ACTIVE.try_lock() {
			// ...
		}

		if call_existing {
			old_hook(panic_info);
		}
	})
}
```

ctru-rs/src/applets/error.rs
```diff
@@ -0,0 +1,149 @@
...
+            }
+
+            let payload = format!("thread '{name}' {panic_info}");
+
+            let mut popup = PopUp::new(Kind::Top);
+
+            popup.set_text(&payload);
+
+            unsafe {
+                popup.launch_unchecked();
```
ian-h-chamberlain: Just to check my understanding, this will block the app until the popup is closed, and then the app will quit (since there's no `loop` or anything)?

ctru-rs/src/applets/error.rs
```diff
@@ -0,0 +1,149 @@
...
+/// If the `Gfx` service is not initialized during a panic, the error applet will not be displayed and the panic
+/// message will be printed over stderr.
+pub fn set_panic_hook(use_stderr: bool) {
+    use crate::services::gfx::GFX_ACTIVE;
+
+    use std::io::Write;
+    use std::sync::TryLockError;
+
+    std::panic::set_hook(Box::new(move |panic_info| {
+        let _apt = Apt::new();
```
ian-h-chamberlain: We should check the error result here right? Like along with the `try_lock()` lower down this needs to be `Ok` for the popup to work, right?

ctru-sys/src/lib.rs
```diff
@@ -16,6 +16,24 @@
 pub mod result;
 pub use result::*;
 
+// Fun fact: C compilers are allowed to represent enums as the smallest integer type that can hold all of its variants,
+// meaning that enums are allowed to be the size of a `c_short` or a `c_char` rather than the size of a `c_int`.
+// Libctru's `errorConf` struct contains two enums that depend on this narrowing property for size and alignment purposes,
+// and since `bindgen` generates all enums with `c_int` sizing, we have to blocklist those types and manually define them
+// here with the proper size.
```
ian-h-chamberlain, line 23: This is really interesting! Similar to what I mentioned in #166 , maybe we can just force these constants to generate using `errorReturnCode` as the type instead of `c_int`?

To catch the struct size/alignment issues, I wonder if we could have caught that using bindgen's layout tests. I think they were omitted before because we didn't have any good way to run them, but with `test-runner` now we might be able to actually do it in CI. I'll file an issue (edit: #167)

Edit: `bindgen` also has a `.fit_macro_constants(true)` we could try to use for this purpose, but I'm not sure if it would help with the actual struct size.

ian-h-chamberlain (approved):
LGTM

ctru-sys/src/lib.rs
```diff
@@ -16,6 +16,24 @@
...
+pub const ERROR_UNKNOWN: errorReturnCode = -1;
+pub const ERROR_NONE: errorReturnCode = 0;
+pub const ERROR_SUCCESS: errorReturnCode = 1;
+pub const ERROR_NOT_SUPPORTED: errorReturnCode = 2;
+pub const ERROR_HOME_BUTTON: errorReturnCode = 10;
+pub const ERROR_SOFTWARE_RESET: errorReturnCode = 11;
+pub const ERROR_POWER_BUTTON: errorReturnCode = 12;
+pub type errorReturnCode = libc::c_schar;
```
ian-h-chamberlain: nit: I would normally put the type definition (or alias in this case) before any values that use that type, even though this order is allowed by the compiler.

ctru-rs/src/applets/error.rs
```diff
@@ -88,17 +88,21 @@ impl PopUp {
...
-pub fn set_panic_hook(use_stderr: bool) {
+/// If the `Gfx` service is not initialized during a panic, the error applet will not be displayed and the default
+/// hook will be called.
+pub fn set_panic_hook(call_default_hook: bool) {
     use crate::services::gfx::GFX_ACTIVE;
-
-    use std::io::Write;
     use std::sync::TryLockError;
 
+    // Ensure we get the default hook instead of a previously registered user hook.
```
ian-h-chamberlain: Hmm, this is actually the opposite behavior I was thinking, since users might want to use this _then_ their own hook. Since it's just a single call to `take_hook` for the user to restore the default I'd personally prefer this to just say something like `let prev_hook = std::panic::take_hook();`

FenrirWolf: The reason I chose to fall back to the default hook is because there are multiple mechanisms to redirect stdout/stderr on the 3DS (see [Soc::redirect_to_3dslink](https://rust3ds.github.io/ctru-rs/crates/ctru/services/soc/struct.Soc.html#method.redirect_to_3dslink) and [consoleDebugInit](https://libctru.devkitpro.org/console_8h.html#ac4fe073b8a8251a5216b66eef8788da2)), and this implementation makes it easy to still get those printed panic messages over `3dslink` or `GDB` if you want them.

I guess we could allow for calling arbitrary hooks alongside our own, but I'm not sure how useful that would be compared to ensuring a stable panic hook environment that smoothly hooks into existing APIs that we already support. But I'm also down to support that case as long as we can document everything clearly enough.

ctru-rs/src/applets/error.rs
FenrirWolf: Yep, you've got that right. Though the app can technically continue if the panic is stopped by a thread boundary or `catch_unwind` or whatever.

ctru-rs/src/applets/error.rs
FenrirWolf: Changed it to use the previous hook instead of forcing the default

ctru-rs/src/applets/error.rs
```diff
@@ -0,0 +1,146 @@
+//! Error applet
+//!
+//! This applet displays error text as a pop-up message on the lower screen.
```
AzureMarker, line 3: Does the message scroll if it's long?

ctru-rs/src/applets/error.rs
```diff
@@ -0,0 +1,146 @@
...
+use crate::services::{apt::Apt, gfx::Gfx};
+
+use ctru_sys::errorConf;
+
+/// Configuration struct to set up the Error applet.
+#[doc(alias = "errorConf")]
+pub struct PopUp {
```
AzureMarker, line 11: nit: I kind of liked the ErrorApplet name from before, but this works I guess. If not the original name, maybe it could be clearer that it's a pop up message. `PopUpMessage`?

ctru-rs/src/applets/error.rs
```diff
@@ -0,0 +1,146 @@
...
+
+/// The kind of error applet to display.
+#[doc(alias = "errorType")]
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[repr(u32)]
+pub enum Kind {
+    /// Error text is centered in the error applet window.
+    Center = ctru_sys::ERROR_TEXT,
+    /// Error text starts at the top of the error applet window.
+    Top = ctru_sys::ERROR_TEXT_WORD_WRAP,
```
AzureMarker: Does Center not have word wrap? Does it really put the text in the center of the screen? The difference between enum names in Rust and C is confusing, is one of them wrong?

ctru-rs/src/applets/error.rs
```diff
@@ -0,0 +1,146 @@
...
+    ResetPressed = ctru_sys::ERROR_SOFTWARE_RESET,
+}
+
+impl PopUp {
+    /// Initialize the error applet with the provided text kind.
+    #[doc(alias = "errorInit")]
+    pub fn new(kind: Kind) -> Self {
+        let mut state = Box::<errorConf>::default();
+
+        unsafe { ctru_sys::errorInit(state.as_mut(), kind as _, 0) };
```
AzureMarker: Does 0 indicate the current system language?

ctru-rs/src/applets/error.rs
```diff
@@ -0,0 +1,146 @@
...
+    PowerPressed = ctru_sys::ERROR_POWER_BUTTON,
+    /// Reset button pressed while [`PopUp`] was running.
+    ResetPressed = ctru_sys::ERROR_SOFTWARE_RESET,
+}
+
+impl PopUp {
+    /// Initialize the error applet with the provided text kind.
+    #[doc(alias = "errorInit")]
+    pub fn new(kind: Kind) -> Self {
+        let mut state = Box::<errorConf>::default();
```
AzureMarker, line 47: Why does this need to be boxed?

ctru-rs/src/applets/error.rs
```diff
@@ -0,0 +1,146 @@
...
+
+        Self { state }
+    }
+
+    /// Sets the error text to display.
+    #[doc(alias = "errorText")]
+    pub fn set_text(&mut self, text: &str) {
+        for (idx, code_unit) in text
+            .encode_utf16()
+            .chain(std::iter::once(0))
```
AzureMarker: This chain should be after the `take`, so we guarantee it is null terminated. Right now if a long enough string is passed, the resulting string is not null terminated.

ctru-rs/src/applets/error.rs
```diff
@@ -0,0 +1,146 @@
...
+        unsafe { ctru_sys::errorInit(state.as_mut(), kind as _, 0) };
+
+        Self { state }
+    }
+
+    /// Sets the error text to display.
+    #[doc(alias = "errorText")]
+    pub fn set_text(&mut self, text: &str) {
+        for (idx, code_unit) in text
+            .encode_utf16()
```
AzureMarker, line 63: Why not use CString? Does it have to be UTF16?

ctru-rs/src/applets/error.rs
```diff
@@ -0,0 +1,146 @@
...
+            ctru_sys::ERROR_SOFTWARE_RESET => Err(Error::ResetPressed),
+            _ => Err(Error::Unknown),
+        }
+    }
+
+    /// Launches the error applet without requiring an [`Apt`] or [`Gfx`] handle.
+    ///
+    /// # Safety
+    ///
+    /// Causes undefined behavior if the aforementioned services are not actually active when the applet launches.
```
AzureMarker: Has the behavior been checked? For example if GFX isn't initialized, what actually happens? I'm wondering if the system would handle it as an error, vs full undefined behavior.

ctru-rs/src/applets/error.rs
```diff
@@ -0,0 +1,146 @@
...
+    unsafe fn launch_unchecked(&mut self) {
```
AzureMarker: This could have the return code parsing and return the error. Then `launch` can just directly call this.

ctru-rs/src/applets/error.rs
```diff
@@ -0,0 +1,146 @@
...
+        unsafe { ctru_sys::errorDisp(self.state.as_mut()) };
+    }
+}
+
+/// Sets a custom [panic hook](https://doc.rust-lang.org/std/panic/fn.set_hook.html) that uses the error applet to display panic messages.
```
AzureMarker, line 97: nit: Are you able to link to set_hook by module path instead of by HTTP link?

ctru-rs/src/applets/error.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    let old_hook = std::panic::take_hook();
+
+    std::panic::set_hook(Box::new(move |panic_info| {
+        let thread = std::thread::current();
+
+        let name = thread.name().unwrap_or("<unnamed>");
+
+        // If we get a `WouldBlock` error, we know that the `Gfx` service has been initialized.
+        // Otherwise fallback to printing over stderr.
```
AzureMarker: This comment isn't entirely accurate, because we really just run the old hook, and we don't know what it will do.

AzureMarker:
Thanks for adding this!

ctru-rs/src/applets/error.rs
Meziu: Yep, I did checks back in the day to the applets to see whether they were actually working without the required services. `Apt` is obviously needed to handle the service code (like launching the libapplet) while I remember clearly that without `Gfx` the application either panicked or quit immediately on applet launching.

ctru-rs/src/applets/error.rs
FenrirWolf: We don't seem to get local docs for `std` when doing `cargo 3ds doc`, so module path links to `std` don't work.

ctru-rs/src/applets/error.rs
FenrirWolf: I might rename these after playing around with them some more. When I first experimented with them, it seemed like the word wrapping behavior was identical and only the text position was different. But I've since run into other examples that disprove that idea.

ctru-rs/src/applets/error.rs
FenrirWolf: The argument is unused since we never call the function with the `ERROR_LANGUAGE_FLAG` set.

ctru-rs/src/applets/error.rs
FenrirWolf: [`errorConf` is a big struct](https://libctru.devkitpro.org/structerrorConf.html)

ctru-rs/src/applets/error.rs
FenrirWolf: The error applet displays UTF-16 text rather than UTF-8

ctru-rs/src/applets/error.rs
FenrirWolf: Nope, and that's an unfortunate limitation of the error applet that I should probably document somewhere, especially if it leads to truncated panic messages. That's part of why I wanted to make sure the hook could still print to stderr along with the popup message.

ctru-rs/src/applets/error.rs
FenrirWolf: I guess the question is if that behavior counts as a controlled panic or shutdown, or if things start dereferencing null pointers or whatever and that's where the trouble comes from.

ctru-rs/src/applets/error.rs
FenrirWolf: But now the question is, what names are actually good here? `WordWrap` and `NoWordWrap`?

FenrirWolf: I went with `WordWrap::Disabled` and `WordWrap::Enabled`

ctru-rs/src/applets/error.rs
FenrirWolf: I added some documentation about the subject. I'll have to experiment some more to figure out exactly how many characters one can expect to get displayed though.

ctru-rs/src/applets/error.rs
Meziu: Ok, I just did a dive in my old tests and can confirm that improper (or missing) setup of `Gfx` when using applets causes an ARM panic with a segfault.

Most likely (though I have never checked) it has to do with the lack of properly allocated video buffers, but I digress.

AzureMarker (approved):

ctru-rs/src/applets/error.rs
FenrirWolf: False alarm everybody, the applet actually does give a scrollbar for long messages.

I'm just a dummy and had tested with really long strings of single characters. You know, like "AAAAAAAAA" but much longer. If you do that, the text will go outside of view and you won't get a scrollbar. But long messages made up of normal words? Those will scroll just fine.

edit: didn't mean to unresolve the convo there, my bad

FenrirWolf:
By the way, the error applet actually does do scrolling for long panic messages. I'm just a dummy and had tested things wrong earlier:

2024-02-27_00-13-16 346_to 1

So let's get this thing merged.

PR #163, opened 2024-02-20
Improve filter callback API redux

FenrirWolf:
Now that our code is in control of calling the software keyboard, we can switch away from using `CString`s in the filter callback without incurring extra allocations. Plus there's a lot of room to simplify the implementation overall, which I've been able to do here too.

### DIFF
--- a/ctru-rs/examples/software-keyboard.rs
+++ b/ctru-rs/examples/software-keyboard.rs
@@ -5,8 +5,6 @@
 use ctru::applets::swkbd::{Button, CallbackResult, SoftwareKeyboard};
 use ctru::prelude::*;
 
-use std::ffi::CString;
-
 fn main() {
     let apt = Apt::new().unwrap();
     let mut hid = Hid::new().unwrap();
@@ -22,11 +20,10 @@ fn main() {
     // Using this callback it's possible to integrate the applet
     // with custom error messages when the input is incorrect.
     keyboard.set_filter_callback(Some(Box::new(|str| {
-        // The string is guaranteed to contain valid Unicode text, so we can safely unwrap and use it as a normal `&str`.
-        if str.to_str().unwrap().contains("boo") {
+        if str.contains("boo") {
             return (
                 CallbackResult::Retry,
-                Some(CString::new("Ah, you scared me!").unwrap()),
+                Some(String::from("Ah, you scared me!")),
             );
         }
 
--- a/ctru-rs/examples/software-keyboard.rs
+++ b/ctru-rs/examples/software-keyboard.rs
@@ -19,7 +19,7 @@ fn main() {
     // Custom filter callback to handle the given input.
     // Using this callback it's possible to integrate the applet
     // with custom error messages when the input is incorrect.
-    keyboard.set_filter_callback(Some(Box::new(|str| {
+    keyboard.set_filter_callback(Some(Box::new(move |str| {
         if str.contains("boo") {
             return (CallbackResult::Retry, Some("Ah, you scared me!".into()));
         }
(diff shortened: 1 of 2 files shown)


Meziu (approved):
LGTM! A great shaving of wrapper code around `libctru` thanks to an up-close implementation. The API feels much better now.

ctru-rs/examples/software-keyboard.rs
```diff
@@ -21,13 +19,9 @@ fn main() {
...
     // with custom error messages when the input is incorrect.
-    keyboard.set_filter_callback(Some(Box::new(|str| {
-        // The string is guaranteed to contain valid Unicode text, so we can safely unwrap and use it as a normal `&str`.
-        if str.to_str().unwrap().contains("boo") {
-            return (
-                CallbackResult::Retry,
-                Some(CString::new("Ah, you scared me!").unwrap()),
-            );
+    keyboard.set_filter_callback(Some(Box::new(move |str| {
+        if str.contains("boo") {
```
ian-h-chamberlain, line 23: Huh, just realized I'm surprised there's no lint for shadowing builtin `str`, but I guess shadowing in general is not linted for so idk

ctru-rs/src/applets/swkbd.rs
```diff
@@ -6,26 +6,25 @@
...
 use bitflags::bitflags;
 
+use std::borrow::Cow;
 use std::ffi::{CStr, CString};
 use std::fmt::Display;
 use std::iter::once;
 use std::str;
 
-type CallbackFunction = dyn Fn(&CStr) -> (CallbackResult, Option<CString>);
+type CallbackFunction = dyn Fn(&str) -> (CallbackResult, Option<Cow<'static, str>>);
```
ian-h-chamberlain, line 21: Shouldn't this be `Cow<'a>` where `'a` is the lifetime of the input string? i.e.
```suggestion
type CallbackFunction<'a> = dyn Fn(&'a str) -> (CallbackResult, Option<Cow<'a, str>>);
```

ian-h-chamberlain (approved):
LGTM too!

ctru-rs/src/applets/swkbd.rs
FenrirWolf: The thing we actually work with in the API is `Option<Box<CallbackFunction>>`, and that has an implicit `'static` bound on the output lifetime. So I figured we might as well have the typedef reflect that too.

ian-h-chamberlain: Hm, is there really any reason to make it a `Cow` then? I thought the idea with that was we could reuse the incoming string's lifetime, but it seems like with the `'static` bound might as well just make it a `String`

FenrirWolf: Take a look at [SoftwareKeyboard::set_filter_callback](https://rust3ds.github.io/ctru-rs/crates/ctru/applets/swkbd/struct.SoftwareKeyboard.html#method.set_filter_callback) to see how this function is actually used. The input string contains text that was just typed into the Software Keyboard. The output is a response message that the keyboard can reply with based on whatever logic you have written. The input and output lifetimes are unrelated by design.

Notice how in the current API, the user is forced to heap-allocate their response message. The purpose of using `Cow` is so that the user doesn't have to allocate unless they want to reply with a formatted message.

ian-h-chamberlain: ahhh, ok, so the idea would be so you could return `Cow::from("foo")` or something, but not necessarily `Cow::from(input_str)` (would need `Cow::from(input_str.to_string()`?). Is that understanding accurate?

FenrirWolf: Yep, you got it.

PR #165, opened 2024-02-23
Require unsafe blocks in unsafe fns

FenrirWolf:
Applies `#![deny(unsafe_op_in_unsafe_fn)]` to the whole crate, because marking unsafe code is a good idea even in unsafe functions.


FenrirWolf:
Think I'll just merge this since, [as discussed in this review thread](https://github.com/rust3ds/ctru-rs/pull/157#discussion_r1493107166), we already know we want it in

ian-h-chamberlain (commented):
LGTM

PR #168, opened 2024-02-26
Enable layout tests for generated bindings

### LINKED ISSUE #167: See if we can use bindgen layout tests
> To catch the struct size/alignment issues, I wonder if we could have caught that using bindgen's layout tests. I think they were omitted before because we didn't have any good way to run them, but with `test-runner` now we might be able to actually do it in CI. I'll file an issue

_Originally posted by @ian-h-chamberlain in https://github.com/rust3ds/ctru-rs/pull/162#discussion_r1501939671_

ian-h-chamberlain:
Closes #167 

Bindgen can help us catch ABI issues sometimes! Let's use it, and see if it catches any issues.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,11 +43,16 @@ jobs:
       - name: Check formatting
         run: cargo fmt --all --verbose -- --check
 
+      # We always run the next two steps, so that even if formatting fails we
+      # still get compilation errors for the same run (mainly for matchers).
+
       - name: Cargo check ctru-sys (without tests)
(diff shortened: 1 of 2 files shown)


ian-h-chamberlain:
Well, I guess this didn't fix the issue mentioned in #162, and didn't flag any other issue, so I'm not sure how helpful it will be, but we may as well enable it since it works with very little effort.

FenrirWolf:
Yeah, bindgen can't detect enum-related layout issues if it legit thinks every enum will be 4 bytes wide. But we might as well have these tests too.

FenrirWolf (approved):

PR #169, opened 2024-02-26
Support console stderr redirection

FenrirWolf:
Allows you send stderr output to `GDB` or other debuggers. It's a nice alternative to 3dslink redirection if you're at the point where you have to jump into an interactive debugging session.


ctru-rs/src/console.rs
```diff
@@ -38,6 +39,18 @@ pub enum Dimension {
     Height,
 }
 
+/// Destination for stderr redirection with [`redirect_stderr`].
+#[doc(alias = "debugDevice")]
+#[repr(u32)]
+pub enum Destination {
```
FenrirWolf, line 43: I'm not sure if we should call this `Destination` or `Redirect` or what. I'm open to suggestions about that.

ctru-rs/src/console.rs
```diff
@@ -66,6 +79,12 @@ pub struct Console<'screen> {
     screen: RefMut<'screen, dyn ConsoleScreen>,
 }
 
+/// Send output from stderr to the specified [`Destination`]. This function can be used to capture error and panic messages with `GDB` or other debuggers.
+#[doc(alias = "consoleDebugInit")]
+pub fn redirect_stderr(dest: Destination) {
+    unsafe { consoleDebugInit(dest as _) }
+}
```
ian-h-chamberlain, line 90: This looks fine to me offhand, I wonder how it behaves if you also use `redirect_to_3dslink` with `stderr = true` ? I think they use the same mechanism under the hood, so maybe whichever you call first gets overridden by the second one?

ctru-rs/src/console.rs
ian-h-chamberlain: `Destination` or `RedirectTarget` maybe?

Also I doubt more would ever be added but maybe `#[non_exhaustive]` would be good, in case we wanted to make e.g. a `Socket` target that would use `redirect_to_3dslink` under the hood or something like that

ian-h-chamberlain (approved):

ctru-rs/src/console.rs
FenrirWolf: I took a look inside libctru and the mechanisms are dfferent. `consoleDebugInit` [changes the underlying devoptab for stderr](https://github.com/devkitPro/libctru/blob/master/libctru/source/console.c#L758-L761) and `3dslinkConnectToHost` [replaces stderr's file descriptor via dup2](https://github.com/devkitPro/libctru/blob/master/libctru/source/3dslink.c#L41-L45). So unless I'm mistaken, 3dslink redirection will always take priority over console redirection.

Meziu: That behaviour should be properly documented.

Meziu (approved):
Forgot about this one. LGTM

PR #172, opened 2024-03-07
More swkbd API improvements

FenrirWolf:
A few small changes for `set_initial_text` and `set_hint_text` now that we don't have to use `CString`s anymore.

### DIFF
--- a/ctru-rs/src/applets/swkbd.rs
+++ b/ctru-rs/src/applets/swkbd.rs
@@ -545,7 +545,7 @@ impl SoftwareKeyboard {
 
         for (idx, code_unit) in text
             .encode_utf16()
-            .take(ctru_sys::SWKBD_MAX_BUTTON_TEXT_LEN as _)
+            .take(button_text.len() - 1)
             .chain(once(0))
             .enumerate()
         {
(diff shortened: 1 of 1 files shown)


ctru-rs/src/applets/swkbd.rs
```diff
@@ -577,10 +585,13 @@ impl SoftwareKeyboard {
...
-    // A reimplementation of `swkbdInputText` from `libctru/source/applets/swkbd.c`. Allows us to
-    // get text from the software keyboard and put it directly into a `String` without requiring
-    // an intermediate fixed-size buffer
-    fn swkbd_input_text(&mut self, output: &mut String) -> SwkbdButton {
+    /// A reimplementation of `swkbdInputText` from `libctru/source/applets/swkbd.c`. Allows us to fix various
+    /// API nits and get rid of awkward type conversions when interacting with the Software Keyboard.
+    ///
+    /// # Safety
+    ///
+    /// The [`Apt`] and [`Gfx`] services must be active when this function is called.
```
AzureMarker: Is there a way to avoid making this unsafe? Can it take them as params, or maybe in the struct constructor?

FenrirWolf: Sure, we could pass in the Apt and Gfx handles. I just didn't see that as necessary since this is an internal API and the function's single caller already ensures that the handles are alive.

Meziu (approved):

ian-h-chamberlain (approved):

PR #177, opened 2024-04-12
Make ndsp::wave::Wave Send + Sync

0x00002a:
Currently the `Wave` wrapper struct is not `Send + Sync`, this is an issue for actually using it as data really needs to be fed to the ndsp service using threads or else it will cause audio clipping (libctru example uses threads and my own experience with a cpal backend supports this).

I need this to properly implement a cpal backend without resorting to [a wrapper](https://github.com/Team-Yarg/cpal/blob/9650bcaba8646d66fcf97ebc0665df9e89443612/src/host/ctru3ds/mod.rs#L129). 

In terms of soundness, as far as I can tell from reading the code and having the cpal backend seem to work fine, this should be okay. The libctru example also backs this up since it uses it in a multithreaded manner but I don't wanna rely on that being 100% correct :p.

Note I've added a dependency on `static_assertions` so that this part of the API is tested but I'm not married to it if its an issue

### DIFF
--- a/ctru-rs/Cargo.toml
+++ b/ctru-rs/Cargo.toml
@@ -26,6 +26,7 @@ libc = "0.2.121"
 bitflags = "2.3.3"
 macaddr = "1.0.1"
 widestring = "1.0.2"
+static_assertions = "1.1.0"
 
 [build-dependencies]
 toml = "0.5"
--- a/ctru-rs/src/services/ndsp/wave.rs
+++ b/ctru-rs/src/services/ndsp/wave.rs
@@ -2,6 +2,8 @@
 //!
 //! This modules has all methods and structs required to work with audio waves meant to be played via the [`ndsp`](crate::services::ndsp) service.
 
+use static_assertions::assert_impl_all;
+
 use super::{AudioFormat, Error};
 use crate::linear::LinearAllocator;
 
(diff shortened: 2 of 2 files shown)


Meziu:
I perfectly get where this PR is coming from. As standard (and as myself even stated in my many experiences using `ndsp`), using threads to implement audio is pretty much a necessity if your objective is to have anything that sounds better than hot garbage.

However, and I don’t like saying this, I don’t trust `libctru`’s implementation of `ndsp` to make the required checks to implement the `Send + Sync` traits. In the past I’ve already had to tip-toe around many issues inherent in `libctru`’s implementation (which mainly stem by the dubious way a linked list is embedded in every `ndspWave` object). One of these cases is the weird behaviour around the `Drop` implementation of `Wave`, which obligated me to clear the queue in case the wave was being used [here](https://github.com/rust3ds/ctru-rs/blob/cd80a338fc343ced91dd431101d3883278a8d3fa/ctru-rs/src/services/ndsp/wave.rs#L211).

In the context of single-thread applications (or apps that implement the whole audio logic on a single thread), the drop order issue is something that can be reasonably worked with (and in general, it’s not much of an issue to drop data since any piece of code will be able to check whether data is valid or not before acting).

In multi-threaded apps, these are assumptions that simply cannot be made (not at our crate’s level at least, your app may put more restrictions in place). Even if somehow I did good enough of a job in my implementation to avoid use-after-free, I’m sure a lot of “randomly clearing the queue and breaking the audio code” shenanigans would still happen, and I’m not ready to work around fixing those.

This is the reason why I suggested the idea of making a complete reimplementation in Rust for `ndsp` [here](https://github.com/rust3ds/ctru-rs/blob/cd80a338fc343ced91dd431101d3883278a8d3fa/ctru-rs/src/services/ndsp/mod.rs#L537).

Regardless of my doubts about this, I want this PR to start more discussion about handling this problem, since it’s the first time anyone has showed interest in improving the `ndsp` wrapper 😛.

PR #178, opened 2024-04-20
Add test-runner back to ctru-rs workspace

ian-h-chamberlain:
Related: #106, https://github.com/rust3ds/test-runner/issues/5

Simple drop-in from https://github.com/rust3ds/test-runner without any changes except for Cargo.toml

TODO:
- [x] update README / docs for usage of the test runner crate (maybe also publish to github pages?)
- [x] wait for CI to be fixed probably

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-members = ["ctru-rs", "ctru-sys"]
+members = ["ctru-rs", "ctru-sys", "test-runner"]
 default-members = ["ctru-rs", "ctru-sys"]
 resolver = "2"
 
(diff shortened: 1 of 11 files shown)


Meziu (approved):
Great, `test-runner` is coming back home it seems :smile:.

Also, just noticed, but the changes made to `errorConf` (which rightfully set the final size to 3820) aren't being properly checked by the automated `bindgen` layout tests. Is there a way to override the expected result or avoid testing that type?

ian-h-chamberlain:
> Great, `test-runner` is coming back home it seems 😄.
> 
> Also, just noticed, but the changes made to `errorConf` (which rightfully set the final size to 3820) aren't being properly checked by the automated `bindgen` layout tests. Is there a way to override the expected result or avoid testing that type?

Actually, this is interesting — yes, the `bindgen` test seems to be incorrect (should be 3820), but our generated bindings _also_ seem to be wrong! `errorType` should have size 2 but it actually has size 4, from what I can tell. I'm not sure if there is a way to forcefully adjust the generated test or make sure it's the right size in practice, but I'll look into it.

My big fear is that this means some of our other binding tests are passing when really they should be failing 😱 so hopefully we can adjust something to make the tests more reliable

Meziu:
Great, all checks are passing now 😃

Let's just fix https://github.com/rust3ds/test-runner/pull/20 and we can merge this.

PR #180, opened 2024-04-21
Generate smaller enum types

### LINKED ISSUE #166: Generate constants to match the types they're used as
For a number of "macro constants", the bindings generate them using the `c_int` type, but they're actually intended for use as another type like `u16` or something.

> Might make sense for us to just generate this constant as a `u16` via bindgen's [`ParseCallbacks::int_macro`](https://docs.rs/bindgen/latest/bindgen/callbacks/trait.ParseCallbacks.html#method.int_macro), to help prevent conversions.
> 
> In fact I wonder if we can't do this for a bunch of other constants and stop using `as`, maybe I'll file a separate issue for this more broadly

_Originally posted by @ian-h-chamberlain in https://github.com/rust3ds/ctru-rs/pull/156#discussion_r1501898908_

ian-h-chamberlain:
Closes #166 I guess?

Related to https://github.com/rust3ds/ctru-rs/pull/178#pullrequestreview-2013120422 and some of the changes in #162 

By passing -fshort-enums to clang, we can get the same smaller types that gcc generates for some of these enums.

This also makes sure the layout tests have the same smaller sizes, so the expected sizes now match the actual sizes of the same C structs.

----

One thing this doesn't do is truly check that GCC is generating the same everything... ideally, I think we'd want to link against a C lib that had a bunch of APIs like this or something:
```c
size_t sizeof_errorType() { return sizeof(errorType); }
size_t alignof_errorType() { return alignof(errorType); }
size_t offsetof_errorType() { return offsetof(errorConf, type); }
```
and then a Rust test like
```rs
extern "C" {
	fn sizeof_errorType() -> libc::size_t;
	fn aligneof_errorType() -> libc::size_t;
	fn offsetof_errorType() -> libc::size_t;
}

#[test]
fn layouts() {
	assert_eq!(mem::size_of<errorType>(), sizeof_errorType());
	assert_eq!(mem::align_of<errorType>(), alignof_errorType());
	assert_eq!(mem::offset_of!(errorConf, type_), offsetof_errorType());
}
```

But without `bindgen` doing something like this for us, this seems like a tricky thing to write unless we just do it manually. So for now I think I'll kinda ignore that and leave it as a future enhancement if we really need it.

### DIFF
--- a/ctru-rs/src/applets/error.rs
+++ b/ctru-rs/src/applets/error.rs
@@ -15,7 +15,7 @@ pub struct PopUp {
 /// Determines whether the Error applet will use word wrapping when displaying a message.
 #[doc(alias = "errorType")]
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
-#[repr(u32)]
+#[repr(u16)]
 pub enum WordWrap {
     /// Error text is centered in the error applet window and does not use word wrapping.
     Disabled = ctru_sys::ERROR_TEXT,
--- a/ctru-rs/src/applets/mii_selector.rs
+++ b/ctru-rs/src/applets/mii_selector.rs
@@ -43,7 +43,7 @@ bitflags! {
     ///
     /// See [`MiiSelector::set_options()`] to learn how to use them.
     #[derive(PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Clone, Copy)]
-    pub struct Options: u32 {
+    pub struct Options: u8 {
         /// Show the cancel button.
         const ENABLE_CANCEL = ctru_sys::MIISELECTOR_CANCEL;
         /// Make guest Miis available to select.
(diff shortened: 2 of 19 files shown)


Meziu (approved):
Ouch, quite a lot of size changes. I must say it scares me quite a bit to have all of these representations changed on a whim, but it makes sense seeing how there is some inconsistency between the Rust and C repr.

Praying that this won't break some hidden API :pray: (at least we spare some space compared to before :man_shrugging: ).

FenrirWolf:
If anything the scary/weird part is that gcc seems to use short enums by default for `arm-none-eabi` while clang doesn't. You'd think they would agree on those settings, but at least we can do it manually this way. We'll probably want a similar PR for `citro3d-rs` too

FenrirWolf (approved):

PR #181, opened 2024-04-23
Generate layout tests using real `sizeof`/`alignof`

ian-h-chamberlain:
Similar to the tests generated by `bindgen`, but using our actual toolchain to spit out the expected values instead of what libclang generates (bindgen's default behavior).

The full generated test can be downloaded from Actions: https://github.com/rust3ds/ctru-rs/actions/runs/9072691412/artifacts/1499437252

Here's a partial example:
```rs
#[test]
fn layout_test_tag_aptHookCookie() {
    assert_eq!(
        size_of!(tag_aptHookCookie),
        cpp ! (unsafe [] -> usize as "size_t" { return sizeof (tag_aptHookCookie) ; }),
        "{} != {}",
        stringify!(size_of!(tag_aptHookCookie)),
        stringify!(sizeof(tag_aptHookCookie)),
    );
    assert_eq!(
        align_of!(tag_aptHookCookie),
        cpp ! (unsafe [] -> usize as "size_t" { return alignof (tag_aptHookCookie) ; }),
        "{} != {}",
        stringify!(align_of!(tag_aptHookCookie)),
        stringify!(alignof(tag_aptHookCookie)),
    );
    assert_eq!(
        size_of!(tag_aptHookCookie::callback),
        cpp ! (unsafe [] -> usize as "size_t" { return sizeof (tag_aptHookCookie :: callback) ; }),
        "{} != {}",
        stringify!(size_of!(tag_aptHookCookie::callback)),
        stringify!(sizeof(tag_aptHookCookie::callback)),
    );
    assert_eq!(
        align_of!(tag_aptHookCookie::callback),
        cpp ! (unsafe [] -> usize as "size_t" { return alignof (tag_aptHookCookie :: callback) ; }),
        "{} != {}",
        stringify!(align_of!(tag_aptHookCookie::callback)),
        stringify!(alignof(tag_aptHookCookie::callback)),
    );
    assert_eq!(
        offset_of!(tag_aptHookCookie, callback),
        cpp ! (unsafe [] -> usize as "size_t" { return offsetof (tag_aptHookCookie , callback) ; }),
        "{} != {}",
        stringify!(offset_of!(tag_aptHookCookie, callback)),
        stringify!(offsetof(tag_aptHookCookie, callback)),
    );
}
```


Cargo.toml
```diff
@@ -3,13 +3,19 @@ members = ["ctru-rs", "ctru-sys", "test-runner"]
 default-members = ["ctru-rs", "ctru-sys"]
 resolver = "2"
 
-[patch.'https://github.com/rust3ds/ctru-rs']
+[workspace.dependencies]
+libc = { version = "0.2.121", default-features = false }
```
Meziu: We are behind more than 2 years worth of `libc` releases. Maybe we don't really need them, but there have been some small structural changes in the crate that we might want to catch up to.

Since this PR is heavily dependent on the `libc`/`bindgen`/`cpp` interface, I am suggesting it here, but it isn't a blocking issue.

ctru-sys/Cargo.toml
```diff
@@ -12,19 +12,42 @@ license = "Zlib"
...
+
 [dependencies]
-libc = { version = "0.2.121", default-features = false }
+libc = { workspace = true }
 
 [build-dependencies]
-bindgen = { version = "0.65.1", features = ["experimental"] }
+bindgen = { version = "0.69", features = ["experimental"] }
 cc = "1.0"
+cpp_build = { optional = true, git = "https://github.com/mystor/rust-cpp.git" }
```
Meziu, line 39: Why not use the version on [crates.io](https://crates.io/crates/cpp_build)?

ctru-sys/Cargo.toml
```diff
@@ -12,19 +12,42 @@ license = "Zlib"
...
+regex = { version = "1.10.4", optional = true }
+rust-format = { version = "0.3.4", optional = true, features = ["token_stream"] }
 which = "4.4.0"
 
 [dev-dependencies]
-shim-3ds = { git = "https://github.com/rust3ds/shim-3ds.git" }
-test-runner = { git = "https://github.com/rust3ds/ctru-rs.git" }
+shim-3ds = { workspace = true }
+test-runner = { workspace = true }
+cpp = { git = "https://github.com/mystor/rust-cpp.git" }
```
Meziu: Same thing here about the availability of a version on [crates.io](https://crates.io/crates/cpp). Are they simply not up-to-date?

Meziu (approved):
A very interesting solution for our custom-compiler setup! This might just solve all size-related uncertainties in the testing suite.

Other than for a couple of missing features, which aren't really needed, I feel like this addition has all that it needs to be an useful replacement for bindgen tests. Great job! 👏

Cargo.toml
ian-h-chamberlain: Since we are still using `^0.2.X`, I think we are probably pulling in the latest version, for example: https://github.com/rust3ds/ctru-rs/actions/runs/9072691412/job/24928574082?pr=181#step:6:28

We can certainly update our minimum version, but since there haven't been any breaking changes I don't think it's strictly necessary — but happy to update if you think we should.

ctru-sys/Cargo.toml
ian-h-chamberlain: Mostly so I could use https://github.com/mystor/rust-cpp/pull/111 which hasn't been released yet. We could do without it and just re-define all the `cc` configuration, or pin the commit hash, but I wasn't too worried about it since this is a test-only dependency

ctru-sys/Cargo.toml
ian-h-chamberlain: Ah, I guess I was thinking this one needed to be the same version as `cpp_build`, but I guess it probably doesn't — I can switch this one to be the crates.io version

ian-h-chamberlain:
> Other than for a couple of missing features, which aren't really needed, I feel like this addition has all that it needs to be an useful replacement for bindgen tests. Great job! 👏

This is an interesting point — for now, I've left the bindgen tests in place because I think they might still be useful to assert properties that clang thinks about the generated types... Especially given that only adds around 1 minute to CI time, and takes almost no effort on our part, do you think we should leave those alone and keep running them alongside these new tests? 

Otherwise, I think this PR is pretty much good to go so I will probably merge once everything is passing 🚀

Meziu:
> This is an interesting point — for now, I've left the bindgen tests in place because I think they might still be useful to assert properties that clang thinks about the generated types... Especially given that only adds around 1 minute to CI time, and takes almost no effort on our part, do you think we should leave those alone and keep running them alongside these new tests?

Very much so, `bindgen` tests are still a pretty integral part in the binding generation routine, and they are still the "best" comparative case for the suite added here. I'd keep them around for as long as we feel fit for testing the new suite and tweaking the `bindgen` setup (which after all these past issues I don't trust to be precise enough :sweat_smile:).

> Otherwise, I think this PR is pretty much good to go so I will probably merge once everything is passing 🚀

:+1:

Issue #174, opened 2024-03-26
rust-lang/libc seems to have incorrect definition for `sockaddr_in`


adryzz:
Was trying to get [`mio`](https://github.com/tokio-rs/mio) to build and run on the 3ds, using its [`poll` backend](https://github.com/tokio-rs/mio/issues/1702), such that crates like [`reqwest`](https://github.com/seanmonstar/reqwest) and the literal thousands of libraries that depend on [`tokio`](https://github.com/tokio-rs/tokio)'s [`net`](https://github.com/tokio-rs/tokio/blob/master/examples/connect.rs) feature would run, but i ran into this issue, just asking if it's to report to upstream, and if so, we should check whether other definitions are wrong, such that they can be fixed all at once:

this is `sockaddr_in` in rust:
https://github.com/rust-lang/libc/blob/a0f5b4b21391252fe38b2df9310dc65e37b07d9f/src/unix/newlib/horizon/mod.rs#L34-L38

and this is the devkitpro definition:
https://libctru.devkitpro.org/structsockaddr__in.html

```diff
    pub struct sockaddr_in {
        pub sin_family: ::sa_family_t,
        pub sin_port: ::in_port_t,
        pub sin_addr: ::in_addr,
+       pub sin_zero: [::c_char; 8],
    }
```

Meziu:
Uhm, I thought we had accounted for those definitions already… If we can find more issues in the bindings we could push everything through with one PR, otherwise it’s fine to just merge this one change directly to `libc`.

FenrirWolf:
There's at least one more problem that I was planning on making an issue for: libctru [changed the definition for hostent](https://github.com/devkitPro/libctru/commit/6714c04806c1164eef0718191060ee7a38e64d77) some time ago, and both `libc` and `std` are using the old one still

adryzz:
yeah, we should create some patches to fix libc/std all at once. i'll try checking if other definitions are wrong

adryzz:
uhh one problem: i found this: https://github.com/rust-lang/libc/pull/2725, which is the PR that created this problem. how do we proceed?

edit: for more info, [this](https://github.com/adryzz/mio/tree/3ds-support) branch of [mio](https://github.com/tokio-rs/mio) adds Horizon OS support, and doesn't build without the extra bytes

Meziu:
[...]

PR #186, opened 2024-05-03
Catch conflicts with libc and stop generating them

ian-h-chamberlain:
Basically, we can use a [builtin lint](https://doc.rust-lang.org/beta/nightly-rustc/rustc_lint/builtin/static.AMBIGUOUS_GLOB_REEXPORTS.html) to catch the scenario where we are accidentally re-generating a different type from `libc`. 

When I tried it out using a little bit of `mod` and re-exporting, I found a number of cases where we were generating the same types, which might have led to some incompatibilities if we had e.g. layout mismatches from the `libc` types. This in combination with #181 should help us catch any weird differences between the `libc` definitions and `libctru` as well.


Meziu:
Great find! This is going to help with the issues reported in #174, and give a bit of stability to our building suite.

Gotta love those regressions in the latest nightly, I will push a fix to the Rust tree.

Meziu:
Opened https://github.com/rust-lang/rust/pull/124649 to fix the latest nightly issues.

PR #187, opened 2024-05-08
Check in Cargo.lock and pin older `pthread-3ds` hash to fix CI

ian-h-chamberlain:
See https://github.com/rust3ds/pthread-3ds/issues/30 — still trying to track down the root cause, but we can do this in the meantime to keep CI passing. If we want to unpin `pthread-3ds` after it's fixed we still can.


Meziu:
Closed as of the changes in https://github.com/rust3ds/pthread-3ds/pull/31

PR #188, opened 2024-05-17
Add trait for linear allocation to enforce use of LinearAllocator at compile-time

ian-h-chamberlain:
Also convert `ndsp::Wave` to be generic over any linearly-allocated type so that callers could use Vec etc. if they wanted.

This will also be useful for `citro3d-rs`, e.g. https://github.com/rust3ds/citro3d-rs/pull/42/files#r1477434491

### DIFF
--- a/ctru-rs/examples/audio-filters.rs
+++ b/ctru-rs/examples/audio-filters.rs
@@ -5,7 +5,6 @@
 #![feature(allocator_api)]
 
 use std::f32::consts::PI;
-use std::ops::Deref;
 
 use ctru::linear::LinearAllocator;
 use ctru::prelude::*;
--- a/ctru-rs/examples/audio-filters.rs
+++ b/ctru-rs/examples/audio-filters.rs
@@ -5,6 +5,7 @@
 #![feature(allocator_api)]
 
 use std::f32::consts::PI;
+use std::ops::Deref;
 
 use ctru::linear::LinearAllocator;
 use ctru::prelude::*;
@@ -64,10 +65,10 @@ fn main() {
 
     // We create a buffer on the LINEAR memory that will hold our audio data.
     // It's necessary for the buffer to live on the LINEAR memory sector since it needs to be accessed by the DSP processor.
-    let mut audio_data1 = Box::new_in([0u8; AUDIO_WAVE_LENGTH], LinearAllocator);
+    let mut audio_data1: Box<[_], _> = Box::new_in([0u8; AUDIO_WAVE_LENGTH], LinearAllocator);
 
     // Fill the buffer with the first set of data. This simply writes a sine wave into the buffer.
-    fill_buffer(audio_data1.as_mut_slice(), NOTEFREQ[4]);
+    fill_buffer(&mut audio_data1, NOTEFREQ[4]);
 
     // Clone the original buffer to obtain an equal buffer on the LINEAR memory used for double buffering.
     let audio_data2 = audio_data1.clone();
@@ -154,7 +155,7 @@ fn main() {
         }
 
         // Double buffer alternation depending on the one used.
-        let current: &mut Wave = if altern {
+        let current: &mut Wave<_> = if altern {
             &mut wave_info1
         } else {
             &mut wave_info2
--- a/ctru-rs/src/lib.rs
+++ b/ctru-rs/src/lib.rs
@@ -23,6 +23,7 @@
 #![feature(try_trait_v2)]
 #![feature(allocator_api)]
 #![feature(new_uninit)]
+#![feature(diagnostic_namespace)]
 #![test_runner(test_runner::run_gdb)] // TODO: does this make sense to have configurable?
 #![doc(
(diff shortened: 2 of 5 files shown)


ctru-rs/src/linear.rs
```diff
@@ -47,3 +51,32 @@ unsafe impl Allocator for LinearAllocator {
         }
     }
 }
+
+/// Trait indicating a type has been allocated using [`LinearAllocator`].
+/// This can be used to enforce that a given slice was allocated in LINEAR memory.
+#[diagnostic::on_unimplemented(
+    message = "{Self} is not allocated with `ctru::linear::LinearAllocator`"
+)]
+pub trait LinearAllocation: Sealed {}
```
ian-h-chamberlain: As an alternative to sealing this trait, we could make it `unsafe` to implement, which would allow users to define it on their own types (while being forced to acknowledge the risk if their type is not _actually_ allocated with LINEAR). The alternative is to just force them to convert to `Vec` etc.

ctru-rs/examples/audio-filters.rs
```diff
@@ -5,6 +5,7 @@
 #![feature(allocator_api)]
 
 use std::f32::consts::PI;
+use std::ops::Deref;
```
Meziu: Every once in a while we have to thank GitHub for these automated lint messages.

ctru-rs/examples/audio-filters.rs
```diff
@@ -64,10 +65,10 @@ fn main() {
 
     // We create a buffer on the LINEAR memory that will hold our audio data.
     // It's necessary for the buffer to live on the LINEAR memory sector since it needs to be accessed by the DSP processor.
-    let mut audio_data1 = Box::new_in([0u8; AUDIO_WAVE_LENGTH], LinearAllocator);
+    let mut audio_data1: Box<[_], _> = Box::new_in([0u8; AUDIO_WAVE_LENGTH], LinearAllocator);
```
Meziu, line 67: I haven’t tried running the example locally, but why is it needed to specify the type notation? It looks pretty clear from the call.

ctru-rs/src/linear.rs
```diff
@@ -47,3 +51,32 @@ unsafe impl Allocator for LinearAllocator {
         }
     }
 }
+
+/// Trait indicating a type has been allocated using [`LinearAllocator`].
+/// This can be used to enforce that a given slice was allocated in LINEAR memory.
+#[diagnostic::on_unimplemented(
```
Meziu, line 61: Great use of the new compiler diagnostics messages! :+1:

ctru-rs/src/linear.rs
Meziu: I heavily suggest unsealing the implementation of the `LinearAllocation`, since it would massively simplify interoperability with external types and libraries. Less boilerplate conversion code is a great thing and the responsibility for the `unsafe` implementation would fall on the user anyways.

Meziu (approved):
Great addition. This looks like it might help easing some of the friction that was found when implementing memory requirements in `citro3d` (and might be a long term helping hand in `ctru-rs` as well).

ctru-rs/src/lib.rs
```diff
@@ -23,6 +23,7 @@
 #![feature(try_trait_v2)]
 #![feature(allocator_api)]
 #![feature(new_uninit)]
+#![feature(diagnostic_namespace)]
```
Meziu, line 26: This feature is stable, there is no need to add it to the list.

ctru-rs/examples/audio-filters.rs
Meziu: It's to read it as a slice, gotcha.

ian-h-chamberlain: Yeah, and unfortunately because of the way default type parameters work we can't just write `Box<[_]>` here (since that turns into `Box<[_], GlobalAllocator>` which is the wrong type for the right-hand side). 

We _could_ write the API slightly more generically to allow for `Box<[u8; N]>`, maybe something like this? https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=cf676ded2ec8a48a4ca7be66bec97c70

It's a bit messy though so I think I'd probably prefer to leave it, even if it means callers have to do this kind of coercion to use arrays (and also, `Vec` should always work without needing the extra type annotation).

ctru-rs/src/lib.rs
ian-h-chamberlain: Unfortunately it's not stable yet in the pinned compiler version we have... I could raise the `rust-version` in our crate to `1.78` and bump the pinned nightly, if you think that's preferable?

Edit: actually, tried this but it won't work unless we update to a _very_ recent nightly that contains https://github.com/rust-lang/rust/pull/124649
My preference would probably be to leave it with `#![feature]` for now, but could still go either way.

Meziu: Ah right. I'm for the idea of "if we use an unstable feature, we should update the MSRV as soon as it is stabilised" since stability is the biggest reason I see why we should bump the MSRV. 

Still, for such a small thing it isn't necessary. We can bump it later down the line whenever we see fit.

PR #192, opened 2024-08-15
Slight MSRV bump and lint fixing

Meziu:
An absolutely tiny PR to fix some annoying lints and to push the crate on a higher MSRV.

### DIFF
--- a/ctru-sys/src/lib.rs
+++ b/ctru-sys/src/lib.rs
@@ -3,6 +3,7 @@
 #![allow(non_camel_case_types)]
 #![allow(non_snake_case)]
 #![allow(clippy::all)]
+#![allow(unexpected_cfgs)] // Read below why we necessate a check for rust_analyzer.
 #![deny(ambiguous_glob_reexports)]
 #![cfg_attr(test, feature(custom_test_frameworks))]
 #![cfg_attr(test, test_runner(test_runner::run_gdb))]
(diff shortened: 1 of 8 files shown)


Meziu:
The nightly version (2024-03-10) was chosen because newer 1.78 nightlies suffer from the bug fixed [here](https://github.com/rust-lang/rust/pull/123057).

AzureMarker (approved):

FenrirWolf (approved):

ctru-sys/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 #![allow(non_camel_case_types)]
 #![allow(non_snake_case)]
 #![allow(clippy::all)]
+#![allow(unexpected_cfgs)] // Read below why we necessate a check for rust_analyzer.
```
ian-h-chamberlain, line 6: I probably would have advocated for adding `rust_analyzer` specifically as a [known cfg](https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html#check-cfg-in-lintsrust-table) something like 
```toml
[lints.rust.unexpected_cfgs]
check-cfg = ['cfg(rust_analyzer)']
```
but not a big difference really.

I guess I'm surprised this isn't a [well-known](https://doc.rust-lang.org/nightly/rustc/check-cfg.html#well-known-names-and-values) value but I guess it's rarer to set than many of those other ones

PR #214, opened 2025-03-31
Update ctru-rs to the 2024 edition.

FenrirWolf:
Also bump CI to a more recent nightly and make use of `&raw const/mut` instead of the `addr_of` macros.

I don't remember if edition updates also mean we should update the crate version number. I could do that too if so.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
       matrix:
         toolchain:
           # Run against a "known good" nightly. Rustc version is 1 day behind the toolchain date
-          - nightly-2024-03-10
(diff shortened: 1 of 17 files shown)


Meziu (approved):
Nice, lgtm 👍

adryzz (approved):
so much better without those macros, well done

ian-h-chamberlain (approved):
LGTM, this might also be a good opportunity to clean up clippy warnings in other files since this upgrade likely introduces some new ones.

At one point long ago I was considering adding `-D warnings` to `ci.yml` (for the pinned toolchain job) but guess I never got around to it

PR #216, opened 2025-05-06
Revise swkbd filter callback API

FenrirWolf:
This PR simplifies the definition of `swkbd::CallbackResult` to remove invalid states/force the user to provide error messages when required, and also changes the callback closure from `dyn Fn` to `dyn FnMut` since that lets you do more interactive things with the callback itself.


Meziu (approved):
I enjoy the simplification of the return type (admittedly, that tuple wasn't great syntax), and the `FnMut` spec might help adding state to the callbacks. LGTM :+1:

PR #217, opened 2025-05-06
Move custom panic hook function to the crate root

FenrirWolf:
Setting the ctru panic hook can be pretty useful even if you don't intend to do anything else with the error applet, and I suspect that most users don't realize that this function exists at all. So I figured we might as well move the function to a more conspicuous location.


Meziu (approved):
I'm ok with the transfer, since it should be clear that the hook uses the error applet in the documentation.

PR #222, opened 2025-08-04
Edition bump and update to dependencies

Meziu:
# Changes
Edition bump to 2024, MSRV upped to 1.88 and update to most dependencies.
Especially `bindgen`, which can now generate `MiiData`, so that it won't need to be `opaque` anymore. I also implemented the `From<ctru_sys::MiiData>` trait using the basic `bindgen` functions rather than our original manual implementations.

There have been some minor clippy lints and formatting changes as well.

# Problems ~~(as always)~~
@ian-h-chamberlain The new version of `ferris-says` uses the `regex` crate (great dependency bloat :partying_face:), and this causes a not-so peculiar stack overflow in the `hello-world` example *only* when running in `debug` mode, as should-have-been-fixed in #59. 
This is because the `__stacksize__` symbol is not properly set to 2MB, even though the `big-stack` feature is enabled. This is very much due to the fact that `ctru-rs` is linked before `libctru`, causing the same old link priority issue we have had for the entirety of this toolchain's life.

It's easy to check the symbol's value by doing:
```bash
arm-none-eabi-objdump -t "target/armv6k-nintendo-3ds/debug/examples/hello-world.elf" | grep 'stacksize'
```
and then searching in the binary the symbol's value at the specified section/address. Its value is usually either:
- 0x00800000, 32KB in little endian
- 0x00002000, 2MB in little endian 

The issue can be resolved by inserting the definition directly in the example (since the binary is linked last):
```Rust
#[unsafe(no_mangle)]
static __stacksize__: usize = 2 * 1024 * 1024;
```

How do you think this should be fixed? I'm perfectly fine with yanking `ferris-says` to make the example work, but I didn't know this issue affected the examples, and am worried maybe our patch didn't actually work that well. Maybe we should resort to the "macro implementation" as originally proposed in #59?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,7 +93,7 @@ jobs:
           args: --doc --workspace
 
       - name: Upload citra logs and capture videos
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         if: success() || failure() # always run unless the workflow was cancelled
         with:
           name: citra-logs-${{ matrix.toolchain }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
       matrix:
         toolchain:
           # Run against a "known good" nightly. Rustc version is 1 day behind the toolchain date
-          - nightly-2025-03-30
+          - nightly-2025-07-25
           # Check for breakage on latest nightly
           - nightly
 
@@ -54,7 +54,7 @@ jobs:
     strategy:
       matrix:
         toolchain:
-          - nightly-2025-03-30
+          - nightly-2025-07-25
           - nightly
     continue-on-error: ${{ matrix.toolchain == 'nightly' }}
     runs-on: ubuntu-latest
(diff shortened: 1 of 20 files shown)


FenrirWolf:
Could we just keep using the old version of `ferris_says` to bypass the stack size issue entirely?

But then again, maybe it's good to have an example that illustrates this problem along with the solution since that's something that users might run into in their own programs.

Meziu:
> Could we just keep using the old version of `ferris_says` to bypass the stack size issue entirely?

Absolutely, regardless of the stack issue I'm not a fan of pulling and compiling `regex` in a simple hello world program.

> But then again, maybe it's good to have an example that illustrates this problem along with the solution since that's something that users might run into in their own programs.

True, though I wonder whether a library example is right the place for that. The docs already mention it, but it's true that maybe a little more guidance should be provided.


I think I'll downgrade the version of `ferris_says` back to 0.2 and call it a day about this, but I still wanted to at least report it somewhere.

FenrirWolf:
There are a few suggestions from Github Actions that should be applied, but other than that LGTM

FenrirWolf (approved):

PR #231, opened 2025-09-05
Remove allocations from the ctru panic hook

FenrirWolf:
This PR changes the panic hook so that it creates the error app config ahead of time and then performs no further allocations upon invocation. Or at least no extra allocations on the `ctru-rs` side of things; no idea if there are extra allocations that happen when calling the applet or unwinding the stack, but we don't have control over those things anyway.

### DIFF
--- a/ctru-rs/src/applets/error.rs
+++ b/ctru-rs/src/applets/error.rs
@@ -96,11 +96,13 @@ impl PopUp {
 
 pub(crate) fn set_panic_hook(call_old_hook: bool) {
     use crate::services::gfx::GFX_ACTIVE;
-    use std::sync::TryLockError;
+    use std::sync::{Mutex, TryLockError};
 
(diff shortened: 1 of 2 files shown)


ctru-rs/src/applets/error.rs
```diff
@@ -94,9 +94,41 @@ impl PopUp {
...
+    }
+}
+
 pub(crate) fn set_panic_hook(call_old_hook: bool) {
     use crate::services::gfx::GFX_ACTIVE;
-    use std::sync::TryLockError;
+    use std::fmt::Write;
+    use std::sync::{Mutex, TryLockError};
+
+    static ERROR_CONF: Mutex<errorConf> = unsafe { Mutex::new(std::mem::zeroed()) };
```
melody-rs, line 127: imo this static should be exposed somewhere so other users of ctru-rs can use it?

FenrirWolf:
Think imma just go and merge this.

PR #232, opened 2025-09-07
Add helper type for writing to UTF-16 buffers

FenrirWolf:
This PR takes the new UTF-16 handling code from https://github.com/rust3ds/ctru-rs/pull/231 and applies it everywhere in `ctru-rs` as well as exposing a new public API. It even handles nul-termination properly this time, so that's neat too!

https://github.com/rust3ds/ctru-rs/pull/231 should probably be merged before this one, after which I can rebase and remove the redundant `ErrorConfWriter` code.


FenrirWolf:
Okay, I think I'm done making changes to this branch now and it should be good to review

